PR #1, opened 2023-02-03
Add more helper functions and move InterOp out of cling namespace

sudo-panda:


lib/Interpreter/InterOp.cpp
```diff
@@ -1829,6 +1830,50 @@ namespace InterOp {
...
+
+  TInterp_t CreateInterpreter() {
+    std::string MainExecutableName =
+      llvm::sys::fs::getMainExecutable(nullptr, nullptr);
+    std::string ResourceDir = MakeResourcesPath();
+    std::vector<const char *> ClingArgv = {"-resource-dir", ResourceDir.c_str(),
+                                           "-std=c++14"};
+    ClingArgv.insert(ClingArgv.begin(), MainExecutableName.c_str());
+    return (TInterp_t) new cling::Interpreter(ClingArgv.size(), &ClingArgv[0]);
+  }
```
vgvassilev, line 1869: This part seems available here https://github.com/compiler-research/InterOp/blob/41a32cce4af78c3b1225b90c84515f144055337f/unittests/InterOp/Utils.cpp#L30

sudo-panda: Yeah, but I think this wont be required in the InterOp library as the binary path that we get is the path to python so the resource directory cannot be set properly.

For example, in my system, when running `cppyy` the `GetExecutablePath` function returns:
```
/usr/bin/python3.10
```

The resources path cannot be found by using this and thus is currently using the wrong directory in the `-resource-dir` argument.

lib/Interpreter/InterOp.cpp
```diff
@@ -1829,6 +1830,21 @@ namespace InterOp {
     return 0;
   }
 
+  TInterp_t CreateInterpreter(const char *resource_dir) {
+    std::string MainExecutableName =
+      llvm::sys::fs::getMainExecutable(nullptr, nullptr);
```
vgvassilev, line 1835: We should lift some of the infrastructure I mentioned here https://github.com/compiler-research/InterOp/blob/41a32cce4af78c3b1225b90c84515f144055337f/unittests/InterOp/Utils.cpp#L21-L62

and check if no `resource_dir` is passed we fall back to it.

lib/Interpreter/InterOp.cpp
```diff
@@ -1829,6 +1830,21 @@ namespace InterOp {
     return 0;
   }
 
+  TInterp_t CreateInterpreter(const char *resource_dir) {
```
vgvassilev, line 1833:
```suggestion
  TInterp_t CreateInterpreter(const char *resource_dir = nullptr) {
```

vgvassilev:
@sudo-panda, can you rebase this PR?

Issue #13, opened 2023-02-28
Implement IsConstMethod


vgvassilev:
In the unittests we have a test checking if a method is const. We should provide the relevant implementation for it.

https://github.com/compiler-research/InterOp/blob/a16dedef4d8fa7d5833b179954c073726016d4bf/unittests/InterOp/FunctionReflectionTest.cpp#L607-L621

utkarshraj779:
std::vector<Decl*> Decls, SubDecls;
  std::string code = R"(
    class C {
      void f1() const {}
      const void f1() {}
      void f2() {}
    };
    )";

utkarshraj779:
@vgvassilev  now i merge it in fork i think it gonna work

vgvassilev:
You will need to open a proper pull request and follow the general workflow that we have.

Smit1603:
- @vgvassilev and @sudo-panda can I work on this issue???

vgvassilev:
Sure, go for it.

PR #27, opened 2023-03-04
Add a default argument for InterOp::GetNamed

utkarshraj779:
- [ ] #13


vgvassilev:
Let's close this PR, it proposes to add a commit which is already present from a branch that was already merged in main.

Issue #15, opened 2023-02-28
Implement GetFunctionDefaultArg


vgvassilev:
In the unittests we have a test returning information about function's default argument given a position. We should provide the relevant implementation for it.

https://github.com/compiler-research/InterOp/blob/a16dedef4d8fa7d5833b179954c073726016d4bf/unittests/InterOp/FunctionReflectionTest.cpp#L641-L652

vgvassilev:
Completed in #31.

PR #31, opened 2023-03-06
Implemented GetFunctionArgDefault Module 

Smit1603:
- This pull request will resolve issue #15 
- I have implemented GetFunctionArgDefault module and implemented successfully
- Kindly review and let me know any further changes are required or not
- @vgvassilev and @sudo-panda

### DIFF
--- a/include/clang/Interpreter/InterOp.h
+++ b/include/clang/Interpreter/InterOp.h
@@ -166,6 +166,9 @@ namespace InterOp {
   CallFuncWrapper_t GetFunctionCallWrapper(TInterp_t interp,
                                            TCppFunction_t func);
 
+  /// Returns the default argument value as string.
+  std::string GetFunctionArgDefault(TCppFunction_t func, TCppIndex_t param_index);
+
   TInterp_t CreateInterpreter(const char *resource_dir = nullptr);
 
   TCppSema_t GetSema(TInterp_t interp);
--- a/lib/Interpreter/InterOp.cpp
+++ b/lib/Interpreter/InterOp.cpp
@@ -2221,5 +2221,35 @@ namespace InterOp {
   
     return S->IsDerivedFrom(loc,derivedType,baseType);
   }
+
+  std::string GetFunctionArgDefault(TCppFunction_t func, TCppIndex_t param_index)
+  {
+    auto *D = (clang::Decl *)func;
+    auto *FD = llvm::dyn_cast_or_null<clang::FunctionDecl>(D);
+    auto PI = FD->getParamDecl(param_index);
+
+    if (PI->hasDefaultArg())
+    {
+      std::string Result;
+      llvm::raw_string_ostream OS(Result);
+      Expr *DefaultArgExpr = const_cast<Expr *>(PI->getDefaultArg());
+      DefaultArgExpr->printPretty(OS, nullptr, PrintingPolicy(LangOptions()));
+
+      // FIXME: Floats are printed in clang with the precision of their underlying representation
+      // and not as written. This is a deficiency in the printing mechanism of clang which we require
+      // extra work to mitigate. For example float PI = 3.14 is printed as 3.1400000000000001
+      if (PI->getType()->isFloatingType())
+      {
+        if (!Result.empty() && Result.back() == '.')
+          return Result;
+        auto DefaultArgValue = std::stod(Result);
+        std::ostringstream oss;
+        oss << DefaultArgValue;
+        Result = oss.str();
+      }
+      return Result;
+    }
+    return "";
+  }
 } // end namespace InterOp
 
--- a/unittests/InterOp/FunctionReflectionTest.cpp
+++ b/unittests/InterOp/FunctionReflectionTest.cpp
@@ -634,17 +634,21 @@ TEST(FunctionReflectionTest, DISABLED_GetFunctionArgName) {
   // EXPECT_EQ(InterOp::GetFunctionArgName(Decls[1], 3), "ch");
 }
 
-TEST(FunctionReflectionTest, DISABLED_GetFunctionArgDefault) {
+TEST(FunctionReflectionTest, GetFunctionArgDefault) {
   std::vector<Decl*> Decls, SubDecls;
   std::string code = R"(
     void f1(int i, double d = 4.0, const char *s = "default", char ch = 'c') {}
+    void f2(float i = 0.0, double d = 3.123, long m = 34126) {}
     )";
 
   GetAllTopLevelDecls(code, Decls);
-  // EXPECT_EQ(InterOp::GetFunctionArgDefault(Decls[0], 0), "");
-  // EXPECT_EQ(InterOp::GetFunctionArgDefault(Decls[0], 1), "4.0");
-  // EXPECT_EQ(InterOp::GetFunctionArgDefault(Decls[0], 2), "\"default\"");
-  // EXPECT_EQ(InterOp::GetFunctionArgDefault(Decls[0], 3), "\'c\'");
+  EXPECT_EQ(InterOp::GetFunctionArgDefault(Decls[0], 0), "");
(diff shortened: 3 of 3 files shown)


sudo-panda (changes requested):

lib/Interpreter/InterOp.cpp
```diff
@@ -2165,5 +2165,28 @@ namespace InterOp {
...
+    {
+      std::string Result;
+      llvm::raw_string_ostream OS(Result);
+      Expr *DefaultArgExpr = const_cast<Expr *>(PI->getDefaultArg());
+      DefaultArgExpr->printPretty(OS, nullptr, PrintingPolicy(LangOptions()));
+
+      if (!Result.empty() && Result.back() == '.')
+      {
+        Result += '0';
+      }
```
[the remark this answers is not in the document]
Smit1603: - Actually the main reason to add this branch was that my modified code was giving output for float type default argument as **4.** (rather than **4.0**) , so in order to get proper output I added this branch to check that if default argument is float type and if it stores value like "3.,4.,5." then the code makes sure to add "0" at end of returned string 
- I hope I have justified the question asked by you correctly
- If there are still any questions I am happy to explain you further about it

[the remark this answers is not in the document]
vgvassilev: We need to use a proper pretty printing. This fixes the test case and not the underlying issue.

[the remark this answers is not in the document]
sudo-panda: @vgvassilev I think we can just remove the `0` from the test because I don't see any reason that we would need it in the output. Do you have a case where it could be problematic?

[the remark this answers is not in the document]
vgvassilev: > @vgvassilev I think we can just remove the `0` from the test because I don't see any reason that we would need it in the output. Do you have a case where it could be problematic?

I'd be fine with that if we add a test case with a default argument `3.123` and it passes.

Smit1603:
- As per above discussion I have made few changes
- As mentioned , I tried with default value for float as **3.123** , and I was quite stunned that test case failed as module returned default value as **3.1230000001** 
- Hence in order to remove above error I implemented another conditional statement for float and double type arguments and tried on some new test cases
- Finally recieved correct output
- @vgvassilev and @sudo-panda Kindly review changes made and pls let me know whether any other changes/improvements are required

lib/Interpreter/InterOp.cpp
```diff
@@ -2165,5 +2165,32 @@ namespace InterOp {
...
+    auto *D = (clang::Decl *)func;
+    auto *FD = llvm::dyn_cast_or_null<clang::FunctionDecl>(D);
+    auto PI = FD->getParamDecl(param_index);
+
+    if (PI->hasDefaultArg())
+    {
+      std::string Result;
+      llvm::raw_string_ostream OS(Result);
+      Expr *DefaultArgExpr = const_cast<Expr *>(PI->getDefaultArg());
+      DefaultArgExpr->printPretty(OS, nullptr, PrintingPolicy(LangOptions()));
```
vgvassilev, line 2180: What's in `OS` after this line? Why we need extra work for the floating point case?

Smit1603: - Here OS will store default value of arguments 
- Only reason to add extr work for floating point case is same as I mentioned above that when default value was given **3.123** above function returned value with very high precision and not exact default value (like **3.123000001**)
- Hence to remove this error I added extra work case for floating point case to return same default value as it is provided without giving large precision or wrong answer
- I hope I have answered the question you have asked
- If you do feel some improvement required pls let me know so I could try for better solution
- @vgvassilev

vgvassilev: Clang’s pretty printer does not print that, so it must be an api that already exists. Maybe that’s controlled by the printing policy.

Smit1603: Yes I also think that printing policy is the one that is causing this inappropriate behaviour and searching for a solution to it

Smit1603: - @vgvassilev 
- After searching a lot, I feel that even if I change printing policy I will need to check that if argument is float type or not and if so then its precision must be preserved (using modified policy)
- Hence I feel at the end we might need to check for floating type anyways (as we are doing in current code)
- I am just little confused to what are your expectation from code (as current code also works correctly)
- According to me I don't think that we can remove condition of checking float type variable
- So, could you pls explain me your requirements so I could work on it more properly

vgvassilev: I think I understand now what you are saying. Here is an example which forces clang to print its default arguments: https://godbolt.org/z/vzWxKdsnh 

This seems a deficiency in clang. Perhaps it's worth discussing that at the mailing lists. Clang should print the default argument as written and not as represented.

For the sake of this PR, I think it should be fine to use some rounding mode and add a fixme explaining this implementation detail of clang.

Smit1603: - Sir regarding rounding mode I feel that current code gives default value same as written, so won't it be better than using round mode as it will narrow down precision level of arguments to a fixed number
- And even for round mode I will have to change printing police to setPrecision() (for which floating point variable identification will be necessary)
- So should I add round mode or continue with current code + FIXME 
- @vgvassilev

Smit1603:
- @vgvassilev I have added a FIXME which is explaining the problem, could you pls review it and pls tell me whether I should still use round mode (as clarification for not using it was given in previous comment)
- If any further changes are required pls let me know

lib/Interpreter/InterOp.cpp
```diff
@@ -2165,5 +2165,46 @@ namespace InterOp {
...
+
+      if (PI->getType()->isFloatingType())
```
vgvassilev, line 2196:
```suggestion
      // FIXME: Floats printed in clang with the precision of their underlying representation
      // and not as written. This is a deficiency in the printing mechanism of clang which we
      // extra work to mitigate. For example float PI = 3.14 is printed as 3.1400000000000001.
      if (PI->getType()->isFloatingType())
```

Can you delete the fixme above and use this one instead?

Smit1603: I have made changes in FIXME @vgvassilev as mentioned by you.
Are there any further changes/improvements required pls let me know so that I could start working on it

vgvassilev: Again, single commit message which reads well.

Smit1603: - I hope its according to your requirements

include/clang/Interpreter/InterOp.h
```diff
@@ -164,6 +164,8 @@ namespace InterOp {
   CallFuncWrapper_t GetFunctionCallWrapper(TInterp_t interp,
                                            TCppFunction_t func);
 
+  std::string GetFunctionArgDefault(TCppFunction_t func, TCppIndex_t param_index);
```
vgvassilev, line 167:
```suggestion
  /// Returns the default argument value as string.
  std::string GetFunctionDefaultArgValueAsString(TCppFunction_t func, TCppIndex_t param_index);
```

Smit1603: I have made this changes as mentioned by you sir
I hope this is correct now

vgvassilev:
Please rebase this PR and we will see if the tests pass.

vgvassilev:
Please remove the merge commit.

vgvassilev:
Do you need some help with that?

vgvassilev (approved):
LGTM!

Smit1603:
- @vgvassilev thank you so much for helping me with that , could you just explain me about the steps that you followed so that  I could understand it correctly and will be helpful for me in future
- I tried various steps but wasnt able to understand it so.....
- I am Sry that I wasnt able to perform that task correctly but will definitely learn about it for future purpose

PR #33, opened 2023-03-08
Add test case to GetVariableOffset for const global vars

sudo-panda:
When GetVariableOffset is run on a const global variable it is not able to get it's address.


sudo-panda:
@vgvassilev can you provide a fix?

PR #35, opened 2023-03-11
Variablereflectionutkarsh

utkarshraj779:
@vgvassilev in this function 
we need to create three return type vector dimension
  GetAllSubDecls(Decls[0], SubDecls);
  GetAllSubDecls(Decls[1], SubDecls);
  GetAllSubDecls(Decls[2], SubDecls);
which gonna check the dimension of three given array after that 
this function gonna check is dimension of the given 
if false other wise return the 
equal(arr_dims.begin(), arr_dims.end(), truth_vals.begin());
value

### DIFF
--- a/lib/Interpreter/InterOp.cpp
+++ b/lib/Interpreter/InterOp.cpp
@@ -782,34 +782,26 @@ namespace InterOp {
 
   intptr_t GetVariableOffset(TInterp_t interp, TCppScope_t var)
   {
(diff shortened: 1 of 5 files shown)


vgvassilev:
This pull request does changes only tests, does not implement anything. Let's close it now and please feel free to re-open if there is an implementation fixing a concrete open issue.

PR #46, opened 2023-03-23
Add support for ArrayTypes that are not ConstantArrayType in GetDimensions

sudo-panda:
Uses an enum `DimensionValue::UNKNOWN_SIZE` to denote the size of these types.

I have used a `namespace` instead of an `enum class` for `DimensionValue` so that it can be implicitly converted to and from long int.


vgvassilev:
This change seems to have regressed cppyy from passing 120 to 118 tests: https://github.com/compiler-research/cppyy-backend/actions/runs/4504956608/jobs/7931040048

PR #47, opened 2023-03-31
Add support for virtual bases in GetBaseClassOffset

sudo-panda:
Fixes Base class not found error.

### DIFF
--- a/unittests/InterOp/ScopeReflectionTest.cpp
+++ b/unittests/InterOp/ScopeReflectionTest.cpp
@@ -540,20 +540,34 @@ TEST(ScopeReflectionTest, IsSubclass) {
 TEST(ScopeReflectionTest, GetBaseClassOffset) {
   std::vector<Decl *> Decls;
   std::string code = R"(
-    class A { int a; };
-    class B { int b; };
-    class C : public A, public B { int c; };
-    class D : public A, public B, public C { int d; };
+    class A { int m_a; };
+    class B { int m_b; };
+    class C : virtual A, virtual B { int m_c; };
+    class D : virtual A, virtual B, public C { int m_d; };
+    class E : public A, public B { int m_e; };
   )";
+
+  class A { int m_a; };
+  class B { int m_b; };
+  class C : virtual A, virtual B { int m_c; };
+  class D : virtual A, virtual B, public C { int m_d; };
+  class E : public A, public B { int m_e; };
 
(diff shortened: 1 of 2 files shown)


unittests/InterOp/ScopeReflectionTest.cpp
```diff
@@ -544,6 +544,7 @@ TEST(ScopeReflectionTest, GetBaseClassOffset) {
     class B { int b; };
     class C : public A, public B { int c; };
     class D : public A, public B, public C { int d; };
+    class E : virtual A { int e; };
```
vgvassilev, line 547: Our test should include at least two base classes with a common base. Then in the check we should ensure that there is only one base.

vgvassilev: Perhaps the case `D` above is better to check.

unittests/InterOp/ScopeReflectionTest.cpp
```diff
@@ -554,6 +555,7 @@ TEST(ScopeReflectionTest, GetBaseClassOffset) {
   EXPECT_EQ(InterOp::GetBaseClassOffset(S, Decls[3], Decls[0]), 0);
   EXPECT_EQ(InterOp::GetBaseClassOffset(S, Decls[3], Decls[1]), 4);
   EXPECT_EQ(InterOp::GetBaseClassOffset(S, Decls[3], Decls[2]), 8);
+  EXPECT_EQ(InterOp::GetBaseClassOffset(S, Decls[4], Decls[0]), 12);
```
vgvassilev, line 558: Why here it is 12?

unittests/InterOp/ScopeReflectionTest.cpp
sudo-panda:
```c++
class F : virtual A { int f; }
class G : public F, public E { int g; }
```
Like this?

unittests/InterOp/ScopeReflectionTest.cpp
sudo-panda: https://godbolt.org/z/3db7M9sM6

I don't know why but it checks out

vgvassilev: Ok, fair enough...

vgvassilev:
This is quite nicer! Many thanks!

PR #63, opened 2023-04-19
Add changes to support enums in Cppyy

sudo-panda:


lib/Interpreter/InterOp.cpp
```diff
@@ -204,6 +217,15 @@ namespace InterOp {
     return {};
   }
 
+  TCppType_t GetEnumConstantType(TCppScope_t handle) {
+    auto *D = (clang::Decl *)handle;
```
vgvassilev, line 221: Likewise.

PR #66, opened 2023-04-21
Add to the returned methods the methods of its base classes.

vgvassilev:


vgvassilev:
As we discussed we should not be including the base class methods in the GetMethods interface.

PR #87, opened 2023-06-04
Add initial setup for doxygen documentation

Krishna-13-cyber:
Signed-off-by: <krishnanarayanan132002@gmail.com> Krishna Narayanan


CMakeLists.txt
```diff
@@ -345,5 +345,13 @@ if (TARGET clang-headers)
   list(APPEND LLVM_COMMON_DEPENDS clang-headers)
 endif()
 
+# Generate docs for CppInterOp
+option(INTEROP_INCLUDE_DOCS "Generate build targets for the Cppinterop docs.")
+option(INTEROP_ENABLE_DOXYGEN "Use doxygen to generate Cppinterop interal API documentation.")
+option(INTEROP_ENABLE_SPHINX "Use sphinx to generage Cppinterop user documentation")
```
vgvassilev, line 351:
```suggestion
option(INTEROP_INCLUDE_DOCS "Generate build targets for the CppInterOp docs.")
option(INTEROP_ENABLE_DOXYGEN "Use doxygen to generate CppInterOp interal API documentation.")
option(INTEROP_ENABLE_SPHINX "Use sphinx to generage CppInterOp user documentation")
```

vgvassilev:
@Krishna-13-cyber ping.

docs/conf.py
```diff
@@ -0,0 +1,39 @@
...
+
+# -- Options for HTML output -------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
+
+html_theme = 'alabaster'
+html_static_path = ['_static']
+
+INTEROP_ROOT = '..'
+
+# html_extra_path = [INTEROP_ROOT + '/build/docs/doxygen/html']
```
parth-07, line 32: Why is this commented out?
And also, `html_extra_path` directory probably should not contain `index.html` file. Otherwise,
readthedocs will get confused which page should be used when visiting the website root.

Krishna-13-cyber: In my next patch to fix the issue of doxygen, I have added the ```html_extra_path``` https://github.com/compiler-research/CppInterOp/compare/main...Krishna-13-cyber:CppInterOp:CppInterOp-Doxygen-Implementation#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R54 but still the issue persists .
The index.html for sphinx is being used as website root as we can see the output. The build directory of doxygen which consists of index.html is implicitly generated by default, so are you suggesting to modify the config to remove the index generation?

Current Status including the  ```html_extra_path```:
https://readthedocs.org/projects/cppinterop-docs/builds/21074694/

parth-07: I am suggesting to make the `html_extra_path` to something like, `INTEROP_ROOT + '/build/docs'`. This way we can
access internal documentation at `${WebsiteRoot}/doxygen/html`.

PR #88, opened 2023-06-07
Fix GetNamed not working for typedef'ed parents

sudo-panda:
Some trivial fixes are included as well.


lib/Interpreter/InterOp.cpp
```diff
@@ -494,12 +494,31 @@ namespace InterOp {
     return GetScope(name.substr(start, end), curr_scope);
   }
 
+  namespace {
```
vgvassilev, line 497: Can we make this function file static and remove the anon namespace?

PR #93, opened 2023-06-16
Add test for lookup of typedefs in struct

sudo-panda:
@vgvassilev can you look at this bug.


vgvassilev:

PR #94, opened 2023-06-16
Support typedefs in GetScopeFromType

sudo-panda:


lib/Interpreter/InterOp.cpp
```diff
@@ -463,7 +463,7 @@ namespace InterOp {
   }
 
   static Decl *GetScopeFromType(QualType QT) {
-    if (auto* Type = QT.getTypePtrOrNull()) {
+    if (auto* Type = QT.getCanonicalType().getTypePtrOrNull()) {
```
vgvassilev, line 466: Can you add a test case?

PR #95, opened 2023-06-16
Add back support of typedefs in GetScope

sudo-panda:


lib/Interpreter/InterOp.cpp
```diff
@@ -502,14 +502,13 @@ namespace InterOp {
...
-    ND = llvm::dyn_cast_or_null<NamedDecl>(GetUnderlyingScope(ND));
-
     if (!ND || ND == (NamedDecl *) -1)
       return 0;
 
     if (llvm::isa<NamespaceDecl>(ND)     ||
         llvm::isa<RecordDecl>(ND)        ||
-        llvm::isa<ClassTemplateDecl>(ND))
+        llvm::isa<ClassTemplateDecl>(ND) ||
+        llvm::isa<TypedefNameDecl>(ND))
```
vgvassilev, line 511: We should add a test and update the documentation. 

My feeling is that we should add a fixme to rethink the `GetScope` interface as it seems to be very randomly returning declarations. We would just need to use more interfaces in the users of this function. The fixme should be part of this PR.

sudo-panda: I have added a fixme to make a general-purpose lookup and filter function (like GetScope but could be used to lookup and filter any kind of decl).

PR #99, opened 2023-06-28
Update documentation setup configuration

Krishna-13-cyber:


vgvassilev:
Why do we need llvm 10? CppInterOp will require some work to enable llvm 10 if possible at all.

Krishna-13-cyber:
> Why do we need llvm 10? CppInterOp will require some work to enable llvm 10 if possible at all.

The readthedocs supports llvm-version only till llvm-10 and also prior cmake versions.So I had to use these versions for readthedocs compatibilty.

vgvassilev:
Do we really need to build the codebase with llvm-10? I think if we specified the doxygen build flag we should probably ignore the llvm version checks as we could be only building the doxygen config.

davidlange6:
Where can we see this llvm-10 support? I don't immediately find it

Krishna-13-cyber:
> Do we really need to build the codebase with llvm-10? I think if we specified the doxygen build flag we should probably ignore the llvm version checks as we could be only building the doxygen config.

We just need llvm to pass the cmake checks, it is nothing to do with docs generation. I will try to get some alter method  for this

Krishna-13-cyber:
> Where can we see this llvm-10 support? I don't immediately find it

The llvm-10 support was working for clad but I had upgraded it for interop to llvm-14 but it didnt seem to work well.
Yes, I found it too while building the docs in readthedocs build setup, I will just share a screen capture of build failure (llvm version incompatible).[](url)
build

davidlange6:
whose cmake checks?

davidlange6:
ok, I see your graphic now..

davidlange6:
did you consider using conda to do the installs instead?

parth-07:
@vgvassilev @davidlange6 If I understand correctly, the pull request is not actually building CppInterOp in the readthedocs configuration. `llvm-10` is only required to run CMake on the CppInterOp. CMake will generate documentation targets. These documentation targets are used to generate sphinx and doxygen docs.

> I think if we specified the doxygen build flag we should probably ignore the llvm version checks as we could be only building the doxygen config.

Yes, this can be done. For this, we will have to add an option such as, 
`CPPINTEROP_BUILD_ONLY_DOCS'. This option can modify CMake to explicitly ignore all configurations
related to building the project. But I don't think this option is required for this change, as we can use `llvm-10`
to bypass the CMake checks.

> did you consider using conda to do the installs instead?

CppInterOp is not being built / installed. Only documentation is being built.

CMakeLists.txt
```diff
@@ -116,9 +116,9 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
     find_package(LLVM ${LLVM_VERSION} REQUIRED CONFIG ${llvm_search_hints} NO_DEFAULT_PATHS)
   endif()
 
-  if (NOT LLVM_FOUND AND DEFINED LLVM_DIR)
-    find_package(LLVM REQUIRED CONFIG ${llvm_search_hints} NO_DEFAULT_PATH)
-  endif()
+  # if (NOT LLVM_FOUND AND DEFINED LLVM_DIR)
```
parth-07, line 119: I think this change can be reverted now.

CMakeLists.txt
```diff
@@ -148,9 +148,9 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
     find_package(Clang ${CLANG_VERSION} REQUIRED CONFIG ${clang_extra_hints} NO_DEFAULT_PATH)
   endif()
 
-  if (NOT Clang_FOUND AND DEFINED Clang_DIR)
-    find_package(Clang REQUIRED CONFIG ${clang_extra_hints} NO_DEFAULT_PATH)
-  endif()
+  # if (NOT Clang_FOUND AND DEFINED Clang_DIR)
```
parth-07, line 151: This can be reverted as well.

parth-07 (changes requested):

vgvassilev:
> @vgvassilev @davidlange6 If I understand correctly, the pull request is not actually building CppInterOp in the readthedocs configuration. `llvm-10` is only required to run CMake on the CppInterOp. CMake will generate documentation targets. These documentation targets are used to generate sphinx and doxygen docs.
> 
> > I think if we specified the doxygen build flag we should probably ignore the llvm version checks as we could be only building the doxygen config.
> 
> Yes, this can be done. For this, we will have to add an option such as, `CPPINTEROP_BUILD_ONLY_DOCS`. This option can modify CMake to explicitly ignore all configurations related to building the project. But I don't think this option is required for this change, as we can use `llvm-10` to bypass the CMake checks.
> 
> > did you consider using conda to do the installs instead?
> 
> CppInterOp is not being built / installed. Only documentation is being built.

Yes, but that change would make it look like we support llvm-10 which we do not. So users with llvm-10 might trigger a build which would then fail. That’s what I am trying to avoid here. The best would be to install the right llvm in the container. The second best would be to use a special cmake flag.

parth-07:
> Yes, but that change would make it look like we support llvm-10 which we do not.

Oh, yes you are right. 

> The best would be to install the right llvm in the container. The second best would be to use a special cmake flag.

Yes, this seems right. @Krishna-13-cyber can you please add the special CMake flag for this if we cannot install
right llvm versions?

Krishna-13-cyber:
> Yes, this seems right. @Krishna-13-cyber can you please add the special CMake flag for this if we cannot install right llvm versions?

Yes, I will get it done!

davidlange6:
"llvm-10 is only required to run CMake on the CppInterOp." - this doesn't make much sense to me. Whose limitation is this?

davidlange6:
ok, I guess this is because build.image is basically obsolete. I'd suggest we follow build.os instead.. 
https://docs.readthedocs.io/en/stable/tutorial/#adding-a-configuration-file

Krishna-13-cyber:
> ok, I guess this is because build.image is basically obsolete. I'd suggest we follow build.os instead.. https://docs.readthedocs.io/en/stable/tutorial/#adding-a-configuration-file

I have this approach and works perfect. Thanks!

vgvassilev:
Can you configure your editor to remove trailing white space?

Krishna-13-cyber:
> Can you configure your editor to remove trailing white space?

Yes, I will do this but I don't see any trailing spaces here in this PR!

docs/conf.py
```diff
@@ -25,15 +25,33 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
 
 html_theme = 'alabaster'
-html_static_path = ['_static']
 
-INTEROP_ROOT = '..'
+html_theme_options = {
+    "github_user": "vgvassilev",
```
vgvassilev, line 30: We should probably use compiler-research.

vgvassilev (approved):
LGTM!

PR #102, opened 2023-07-01
Properly implement placement new via InterOp::Construct.

vgvassilev:
This patch should fix the valgrind issues.

cc: @sudo-panda


sudo-panda:
Can you cheery pick the commit in #103

PR #103, opened 2023-07-02
Update CI and add valgrind testing

sudo-panda:


vgvassilev:
That actually looks good. There is another valgrind issue that I am looking into. I'd propose to merge my PR and then this one.

PR #106, opened 2023-07-08
Add test for GetFunctionUsingArgs

sudo-panda:


sudo-panda:
@wlav `InterOp::GetFunctionUsingArgs` (yet to be implemented) is going to substitute `Cppyy::GetMethodTemplate`. Can you suggest any more test cases that could be added to the test?

vgvassilev:

PR #111, opened 2023-07-19
Add test to check base virtual method call

sudo-panda:
When using a derived class object with a base class virtual method the derived class virtual method is being called.

cc: @vgvassilev


vgvassilev:

PR #113, opened 2023-07-30
Running clang-tidy on PR

Smit1603:
- Sry previous pull request was discarded while rebasing owing to which had to create new pull request
- Currently integrated API for clang tidy now checking whether its running or not
- Here we need to take different approach from clad implementation of clang-tidy as we have llvm-project as our requiremet for clang and cling hence tried a slight different approach


vgvassilev:
I am confused. How is clang-tidy related to code coverage?

Smit1603:
> I am confused. How is clang-tidy related to code coverage?

No sir its not related, I just found that for code coverage part also 'coverage' varible had to be implemented in matrix hence added it now. I think i might have missed earlier hence introduced it now

vgvassilev:

PR #119, opened 2023-08-01
Add support for multidimensional arrays in GetUnderlyingType

sudo-panda:


vgvassilev:
@sudo-panda, can you make clang-format happy?

PR #128, opened 2023-08-08
Consolidate the template instantiation logic.

vgvassilev:
This patch offers a single interface for instantiation of class, function and variable templates. That would simplify user code where we do not need to care what is the underlying template pattern (most of the time).


maximusron:
This looks good to go for the incoming template fix patch in 
- #215

maximusron:
@vgvassilev looks like all the bots are failing only on the cppyy and xeus builds due to the pending update in cppyy-backend https://github.com/compiler-research/cppyy-backend/pull/94

I believe this PR can be merged now

PR #130, opened 2023-08-10
Run clang-tidy on PR

Smit1603:
@vgvassilev could you suggest a way to resolve the CMake error, already tried sudo command to install all required libraries yet facing the same issue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
             cling: Off
             cppyy: Off
             coverage: true
+
           - name: ubu22-gcc9-clang13-cling
             os: ubuntu-22.04
             compiler: gcc-9
@@ -97,6 +98,7 @@ jobs:
         conda install -y -q -c conda-forge \
           distro \
           pytest
+
     - name: Restore Cache LLVM/Clang runtime build directory
       uses: actions/cache/restore@v3
       id: cache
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -224,6 +237,24 @@ jobs:
         echo "CPPINTEROP_DIR=$CPPINTEROP_DIR" >> $GITHUB_ENV
         echo "LLVM_BUILD_DIR=$LLVM_BUILD_DIR" >> $GITHUB_ENV
         echo "CPLUS_INCLUDE_PATH=$CPLUS_INCLUDE_PATH" >> $GITHUB_ENV
+
+    - name: Run clang-tidy
+      if: ${{ (matrix.clang-tidy == true) }}
+      uses: ZedThree/clang-tidy-review@v0.13.2
```
vaithak, line 243: Hi @Smit1603, this action runs the `cmake` command in a docker container, so any dependencies that you install in the host machine won't be available to the container while building. 
So, to fix this, you should install these dependencies inside the clang-tidy step itself, you can use the `apt_packages` option to install these dependencies ([reference](https://github.com/ZedThree/clang-tidy-review)).

Smit1603: Ohhh yeah, I understood, actually I was plaining to create a seperate docker container to resolve but this might be best I feel. Will try this

Smit1603:
@vgvassilev I think now the clang tidy part works properly , after merging this PR

vgvassilev (approved):
LGTM!

PR #131, opened 2023-08-10
[cmake] Link against libtinfo if llvm is configured so.

vgvassilev:


vgvassilev:
This is solved on the user side in a different way.

Issue #84, opened 2023-05-27
Extra space in function signature.


sudo-panda:
https://github.com/compiler-research/InterOp/blob/5e0884484c579b0b79bec1532e8296586ea2cce5/lib/Interpreter/InterOp.cpp#LL988C32-L988C63

There is a large gap in the function signature.

vgvassilev:
Does that occur elsewhere?

Smit1603:
@vgvassilev This issue can be closed

vgvassilev:
Indeed.

PR #139, opened 2023-08-27
Add and update tests

Smit1603:


Smit1603:
@vgvassilev 
This PR has following changes made
- Added GetEnums test to EnumReflectionTest.cpp
- Removed extra blank space problem as visible earlier hence fixes #84
- As well as found a small error in findBuiltinType function hence resolved it
- Added some of the edge cases for some of the tests so that more code shall be covered

include/clang/Interpreter/CppInterOp.h
```diff
@@ -173,6 +173,9 @@ namespace Cpp {
   /// Checks if the passed value is an enum type or not.
   bool IsEnumType(TCppType_t type);
 
+  /// Returns vector of enums present in given scope
+  std::vector<std::string> GetEnums(TCppScope_t scope);
```
vgvassilev, line 177: I am wondering where is the implementation of `GetEnums`?

Smit1603: Its already implemented in CppInterOp.cpp

vgvassilev, line 177:
```suggestion
  void GetEnums(TCppScope_t scope, std::vector<std::string>& Result);
```

We should change the signature in this way.

include/clang/Interpreter/CppInterOp.h
```diff
@@ -173,6 +173,9 @@ namespace Cpp {
   /// Checks if the passed value is an enum type or not.
   bool IsEnumType(TCppType_t type);
 
+  /// Returns vector of enums present in given scope
```
vgvassilev, line 176: We should update the comment as the function does not return a vector anymore.

vgvassilev (approved):
LGTM!

PR #142, opened 2023-08-27
[ci] Add llvm17 support.

vgvassilev:
Currently it is rc3 and will be released next week.


vgvassilev:
The build fails but the upstream fix appeared recently so maybe it is a hiccup. Let's merge this PR.

PR #149, opened 2023-09-10
Updating CI with macOS build

Smit1603:
- Added steps to build LLVM/clang on macOS and works perfectly
- Errors are being faced while building and testing CppInterop hence needs to be fixed yet

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
             clang-runtime: '17'
             cling: Off
             cppyy: Off
+            extra_cmake_options: '-DLLVM_USE_LINKER=lld'
 
           - name: ubu22-gcc9-clang-repl-16
             os: ubuntu-22.04
@@ -35,6 +36,7 @@ jobs:
             cling: Off
             cppyy: Off
             coverage: true
+            extra_cmake_options: '-DLLVM_USE_LINKER=lld'
 
           - name: ubu22-gcc9-clang13-cling
             os: ubuntu-22.04
@@ -42,12 +44,23 @@ jobs:
             clang-runtime: '13'
             cling: On
             cppyy: Off
+            extra_cmake_options: '-DLLVM_USE_LINKER=lld'
+
           - name: ubu22-gcc9-clang13-cling-cppyy
             os: ubuntu-22.04
             compiler: gcc-9
             clang-runtime: '13'
             cling: On
             cppyy: On
+            extra_cmake_options: '-DLLVM_USE_LINKER=lld'
+
+          - name: osx-clang-clang-repl-16
+            os: macos-latest
+            compiler: clang
+            clang-runtime: '16'
+            cling: Off
+            cppyy: Off
+
     steps:
     - uses: actions/checkout@v3
       with:
@@ -56,14 +69,15 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: '3.10'
+
     - name: Save PR Info
       run: |
         mkdir -p ./pr
         echo ${{ github.event.number }} > ./pr/NR
         echo ${{ github.repository }} > ./pr/REPO
 
-        cling_on="${{ matrix.cling }}"
-        if [[ "${cling_on^^}" == "ON" ]]; then
+        cling_on=$(echo "${{ matrix.cling }}" | tr '[:lower:]' '[:upper:]')
+        if [[ "$cling_on" == "ON" ]]; then
           export CLING_HASH=$(git ls-remote https://github.com/root-project/cling.git HEAD| tr '\t' '-')
         else
           export CLING_HASH="Repl"
@@ -73,6 +87,13 @@ jobs:
         echo "CLING_HASH=$CLING_HASH" >> $GITHUB_ENV
         echo "LLVM_HASH=$LLVM_HASH" >> $GITHUB_ENV
     - uses: nelonoel/branch-name@v1.0.1
+
+    - name: Setup default Build Type on *nux
+      if: runner.os != 'windows'
+      run: |
+        echo "BUILD_TYPE=Release" >> $GITHUB_ENV
+        echo "CODE_COVERAGE=0" >> $GITHUB_ENV
+
     - name: Setup compiler on Linux
       if: runner.os == 'Linux'
       run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -106,6 +140,16 @@ jobs:
           distro \
           pytest
 
+    - name: Install deps on MacOS
+      if: runner.os == 'macOS'
+      run: |
+        # Install deps
+        ##sudo apt update
```
vgvassilev, line 147: That seems irrelevant comment.

.github/workflows/ci.yml
```diff
@@ -106,6 +140,16 @@ jobs:
...
+        brew install git gcc gnupg python@3 
+        conda install -y -q -c conda-forge \
```
vgvassilev, line 149: Can we get away from conda, and use only brew?

.github/workflows/ci.yml
```diff
@@ -164,6 +209,63 @@ jobs:
           cmake --build . --target clang clang-repl --parallel $(nproc --all)
         fi
         cd ../../
+
+    - name: Build LLVM/Cling on macOS if the cache is invalid
```
vgvassilev, line 213: This step is not really different from the step on unix. Can we merge them?

.github/workflows/ci.yml
```diff
@@ -234,6 +339,58 @@ jobs:
         echo "LLVM_BUILD_DIR=$LLVM_BUILD_DIR" >> $GITHUB_ENV
         echo "CPLUS_INCLUDE_PATH=$CPLUS_INCLUDE_PATH" >> $GITHUB_ENV
 
+    - name: Build and Test/Install CppInterOp on macOS
```
vgvassilev, line 342: Likewise.

vgvassilev (commented):
This pull request looks promising. It uncovered issues in clang-repl and the build status of the OSX builds should be marked as expected to fail. That is, we should probably not run the tests, or run only the ones we know work.

.github/workflows/ci.yml
```diff
@@ -56,14 +65,15 @@ jobs:
...
+
     - name: Save PR Info
       run: |
         mkdir -p ./pr
         echo ${{ github.event.number }} > ./pr/NR
         echo ${{ github.repository }} > ./pr/REPO
-
-        cling_on="${{ matrix.cling }}"
-        if [[ "${cling_on^^}" == "ON" ]]; then
+
```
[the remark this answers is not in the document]
Smit1603: Actually made this changes as while running on macOS , I was facing bad substitution error and found that earlier command to make string uppercase wasn't available for macOS. Hence used this new command which is valid and running for both Linux and macOS

.github/workflows/ci.yml
Smit1603: Taken care of it

.github/workflows/ci.yml
Smit1603: Since its python libraries, so used pip command rather than brew.

.github/workflows/ci.yml
Smit1603: Yaa so I tried combining them and faced an unforeseen error where building of llvm got terminated unexpectedly. Hence looking for it.

.github/workflows/ci.yml
Smit1603: Yeah merged them, just a thing to be mentioned is lld linker is not supported on macOS. Hence i removed that flag and now default ld64 linker is being used in every environment. Is this change ok for our project??

vgvassilev: This can be solved by adding a matrix switch to handle the platform-specific cmake options: https://github.com/vgvassilev/clad/blob/59fd672f54d171577857fe75ed9223b91d7902d1/.github/workflows/ci.yml#L93

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -556,6 +558,7 @@ TEST(FunctionReflectionTest, IsVirtualMethod) {
   EXPECT_FALSE(Cpp::IsVirtualMethod(SubDecls[3])); // y()
 }
 
+#if !(__clang__ && __APPLE__)
 TEST(FunctionReflectionTest, JitCallAdvanced) {
```
vgvassilev, line 562:
```suggestion
#ifdef __APPLE__
TEST(FunctionReflectionTest, DISABLED_JitCallAdvanced) {
#else
TEST(FunctionReflectionTest, JitCallAdvanced) {
#endif
```

We should use this pattern everywhere.

Smit1603: Yes made changes according to this pattern everywhere

.github/workflows/ci.yml
```diff
@@ -129,41 +175,59 @@ jobs:
...
                 -DCLANG_ENABLE_ARCMT=OFF                      \
                 -DCLANG_ENABLE_FORMAT=OFF                     \
                 -DCLANG_ENABLE_BOOTSTRAP=OFF                  \
                 ../llvm
-          cmake --build . --target clang clang-repl --parallel $(nproc --all)
+          if [[ "${os}" == "macos*" ]]; then
+            cmake --build . --target clang clang-repl --parallel $(sysctl -n hw.physicalcpu)
+          else
+            cmake --build . --target clang clang-repl --parallel $(nproc --all)
+          fi
```
vgvassilev, line 227: Can you make this somewhere in the settings of the build and then just use ncpus instead of if-stmts everywhere?

```
if [[ "${os}" == "macos*" ]]; then
   ncpus=$(sysctl -n hw.physicalcpu)
else
   ncpus=$(nproc --all)
fi
```

alexander-penev: Maybe we should use hw.ncpu instead hw.physicalcpu

.github/workflows/ci.yml
```diff
@@ -1,9 +1,9 @@
 name: Main
 on:
-  pull_request:
-    branches: [main]
+  # pull_request:
+  #   branches: [main]
```
vgvassilev, line 4: Why did you comment this section?

.github/workflows/ci.yml
Smit1603: - Made an environment variable to store number of cpu cores and replaced hw.physicalcpu to hw.ncpu

vgvassilev (approved):
LGTM!

PR #153, opened 2023-10-14
Load the compiler instance plugins.

vgvassilev:
This patch should resolve an issue in xeus-clang-repl with clad.


vgvassilev:
The failures are unrelated to this PR.

PR #161, opened 2023-10-31
Updating GetBaseClass

Smit1603:


lib/Interpreter/CppInterOp.cpp
```diff
@@ -615,11 +615,7 @@ namespace Cpp {
 
     auto type = (CXXRD->bases_begin() + ibase)->getType();
     if (auto RT = type->getAs<RecordType>()) {
```
vgvassilev, line 617:
```suggestion
    if (auto RT = type->getAs<RecordType>())
```

lib/Interpreter/CppInterOp.cpp
```diff
@@ -615,11 +615,7 @@ namespace Cpp {
...
-      return (TCppScope_t) RT->getDecl();
-    } else if (auto TST = type->getAs<clang::TemplateSpecializationType>()) {
-      TemplateName TN = TST->getTemplateName();
-      if (auto *TD = dyn_cast_or_null<ClassTemplateDecl>(TN.getAsTemplateDecl()))
-        return TD->getTemplatedDecl();
+      return (TCppScope_t)RT->getDecl();
     }
```
vgvassilev, line 619:
```suggestion
```

Smit1603: Made all suggested changes

vgvassilev:
The commit message should say what the change was done for instead of which function it was in. How about something like `Remove dead code since TemplateSpecializationType is a subclass for RecordType`.

PR #178, opened 2024-01-04
Add support for GetIncludePaths

Krishna-13-cyber:
This PR adds GetIncludePaths() to fetch the include paths

### DIFF
--- a/include/clang/Interpreter/CppInterOp.h
+++ b/include/clang/Interpreter/CppInterOp.h
@@ -445,7 +445,7 @@ namespace Cpp {
   const char* GetResourceDir();
 
   /// Get Include Paths
-  void GetIncludePath(const char* dir, std::vector<std::string>& includePaths);
+  void GetIncludePath(std::vector<std::string>& includePaths);
 
   /// Secondary search path for headers, if not found using the
   /// GetResourceDir() function.
--- a/include/clang/Interpreter/CppInterOp.h
+++ b/include/clang/Interpreter/CppInterOp.h
@@ -444,6 +444,9 @@ namespace Cpp {
   /// Returns the resource-dir path (for headers).
   const char* GetResourceDir();
 
+  /// Get Include Paths
+  void GetIncludePath(const char* dir, std::vector<std::string>& includePaths);
+
   /// Secondary search path for headers, if not found using the
   /// GetResourceDir() function.
   void AddIncludePath(const char *dir);
--- a/lib/Interpreter/CppInterOpInterpreter.h
+++ b/lib/Interpreter/CppInterOpInterpreter.h
@@ -347,6 +347,10 @@ public:
         const_cast<const Interpreter*>(this)->getDynamicLibraryManager());
   }
 
+  /// @brief As a interface to store paths added in AddIncludePaths
+  std::vector<std::string> include;
+  ///
+
   ///\brief Adds multiple include paths separated by a delimter.
   ///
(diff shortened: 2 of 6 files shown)


lib/Interpreter/CppInterOpInterpreter.h
```diff
@@ -383,6 +383,28 @@ class Interpreter {
...
+  ///
+  ///\param[in] includePaths - Store Path(s)
+  ///\param[in] PathsStr - Path(s)
+  ///\param[in] Delim - Delimiter to separate paths or NULL if a single path
+  ///
+  void GetIncludePaths(std::vector<std::string>& includePaths,
+                       llvm::StringRef PathsStr, const char* Delim = ":") {
+    const clang::CompilerInstance* CI = getCompilerInstance();
+    clang::HeaderSearchOptions& HOpts =
+        const_cast<clang::HeaderSearchOptions&>(CI->getHeaderSearchOpts());
```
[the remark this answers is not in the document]
Krishna-13-cyber: Hi @vgvassilev, 
Could you have a look at this, should we try using //NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast) or 
using false for the setup mentioned here (https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-const-cast.html)

include/clang/Interpreter/CppInterOp.h
```diff
@@ -444,6 +444,9 @@ namespace Cpp {
   /// Returns the resource-dir path (for headers).
   const char* GetResourceDir();
 
+  /// Get Include Paths
+  void GetIncludePath(const char* dir, std::vector<std::string>& includePaths);
```
vgvassilev, line 448:
```suggestion
  void GetIncludePath(std::vector<std::string>& includePaths);
```

This function should only have a single output parameter.

Krishna-13-cyber: Oh okay, I will make the required changes.

Krishna-13-cyber:
Hi @vgvassilev,
Pinging you for a review on this updated version, so that we can complete this.

lib/Interpreter/CppInterOpInterpreter.h
```diff
@@ -347,6 +347,10 @@ class Interpreter {
         const_cast<const Interpreter*>(this)->getDynamicLibraryManager());
   }
 
+  /// @brief As a interface to store paths added in AddIncludePaths
+  std::vector<std::string> include;
```
vgvassilev, line 351: Why do we need this. The include paths are stored in the HeaderSearch already.

Krishna-13-cyber: So we know that AddincludePaths() is fetching the paths and giving it to HeaderSearch, now GetIncludePaths() had the aim of exposing all the paths which were added(not present in HeaderSearch) to the user, for this we had to get/store all those paths which were being added to HeaderSearch with a vector ( as an interface) as AddIncludePaths() does'nt store any paths (makes it available for HeaderSearch).

Initially, I tried other methods to store all the paths from AddIncludePaths() itself but for that we require a change in function definition and would not require another function GetIncludePaths(). I also tried creating wrappers around AddIncludePaths() but to maintain the function signature I could'nt find a way to store the paths.

Vipul-Cariappa:
Closed through #311.
Feel free to reopen if I have missed something.

Issue #175, opened 2024-01-02
Clang-repl=ON issue for both clang 16 & 17 on Apple Silicon for cppyy


mcbarton:
Currently if you build CppInterOp with Clang-Repl=ON , you encounter the following error when performing the command python -c "import cppyy" (with both a patched clang 16 or 17 on Apple Silicon)
```
<built-in>:1:10: fatal error: 'new' file not found
    1 | #include "new"
      |          ^~~~~
In file included from <<< inputs >>>:1:
input_line_2:1:10: fatal error: 'iostream' file not found
    1 | #include <iostream>
      |          ^~~~~~~~~~
Failed to parse via ::process:Parsing failed.
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/wf5617/Library/Python/3.12/lib/python/site-packages/cppyy/__init__.py", line 80, in <module>
    from ._cpython_cppyy import *
  File "/Users/wf5617/Library/Python/3.12/lib/python/site-packages/cppyy/_cpython_cppyy.py", line 152, in <module>
    gbl.std =  _backend.CreateScopeProxy('std')
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'std' is not a known C++ class
```
I get no errors when I use CppInterOp with Cling. After a fix to this issue is found then a ci for osx on arm should be able to pass to the same level as ubuntu and osx on x86, once the free runners become available. My initial attempts to try and find the cause of this issue on my local machine have failed. @vgvassilev @alexander-penev any suggestions on what the issue may be?

vgvassilev:
The problem is that Cling has a robust way to find the relevant include paths to pass to the interpreter by asking the underlying compiler: https://github.com/root-project/root/blob/798b9b079f7b6c136bdaaf6787729a1bef865429/interpreter/cling/lib/Interpreter/CIFactory.cpp#L109

One solution would be to get that code and add it right after we create the clang-repl instance.

mcbarton:
@vgvassilev Until I can find the time add add that code you suggest I worked around the issue by appending to CPLUS_INCLUDE_PATH the paths it needed to find the header files. After doing so I a new error arose
```
M0988-Mac:~ wf5617$ python -c "import cppyy"
[GetVariableOffset] Failed to execute PTU:Duplicate definition of symbol '__ZN5cling7runtime6gClingE'
```
Any suggestions on how to fix this issue? 

[...]

PR #181, opened 2024-01-15
Various windows fixes

fsfod:
Fix some of the windows build errors, It still won't fully build with just these changes, but it gets it much closer. 
I don't know much of the llvm internal API, but i did switched from some Linux specific function that were failing to build on windows with llvm API like SearchForAddressOfSymbol and real_path.

### DIFF
--- a/cmake/modules/GoogleTest.cmake
+++ b/cmake/modules/GoogleTest.cmake
@@ -7,14 +7,16 @@ set(_gtest_byproducts
   ${_gtest_byproduct_binary_dir}/lib/libgmock_main.a
   )
 
+set(gtestbuild "Release")
 if(MSVC)
+  set(gtestbuild $<CONFIG>)
   set(EXTRA_GTEST_OPTS
     -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_DEBUG:PATH=${_gtest_byproduct_binary_dir}/lib/
     -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_MINSIZEREL:PATH=${_gtest_byproduct_binary_dir}/lib/
     -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELEASE:PATH=${_gtest_byproduct_binary_dir}/lib/
     -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELWITHDEBINFO:PATH=${_gtest_byproduct_binary_dir}/lib/
     -Dgtest_force_shared_crt=ON
-    BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config Release)
+    BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config $<CONFIG>)
 elseif(APPLE)
   set(EXTRA_GTEST_OPTS -DCMAKE_OSX_SYSROOT=${CMAKE_OSX_SYSROOT})
 endif()
@@ -32,7 +34,7 @@ ExternalProject_Add(
   #            -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELEASE:PATH=ReleaseLibs
   #            -Dgtest_force_shared_crt=ON
   CMAKE_ARGS -G ${CMAKE_GENERATOR}
-                -DCMAKE_BUILD_TYPE=Release
+                -DCMAKE_BUILD_TYPE=${gtestbuild}
                 -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
                 -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
                 -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
--- a/cmake/modules/GoogleTest.cmake
+++ b/cmake/modules/GoogleTest.cmake
@@ -59,7 +59,11 @@ file(MAKE_DIRECTORY ${GTEST_INCLUDE_DIR} ${GMOCK_INCLUDE_DIR})
 
 # Libraries
 ExternalProject_Get_Property(googletest binary_dir)
-set(_G_LIBRARY_PATH ${binary_dir}/lib/)
+if(WIN32)
+  set(_G_LIBRARY_PATH  ${_gtest_byproduct_binary_dir}/lib)
+else()
+  set(_G_LIBRARY_PATH ${binary_dir}/lib/)
+endif()
 
 # Use gmock_main instead of gtest_main because it initializes gtest as well.
 # Note: The libraries are listed in reverse order of their dependancies.
--- a/cmake/modules/GoogleTest.cmake
+++ b/cmake/modules/GoogleTest.cmake
@@ -7,9 +7,7 @@ set(_gtest_byproducts
   ${_gtest_byproduct_binary_dir}/lib/libgmock_main.a
   )
 
-set(gtestbuild "Release")
 if(MSVC)
-  set(gtestbuild $<CONFIG>)
   set(EXTRA_GTEST_OPTS
     -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_DEBUG:PATH=${_gtest_byproduct_binary_dir}/lib/
     -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_MINSIZEREL:PATH=${_gtest_byproduct_binary_dir}/lib/
@@ -34,7 +32,7 @@ ExternalProject_Add(
   #            -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELEASE:PATH=ReleaseLibs
   #            -Dgtest_force_shared_crt=ON
   CMAKE_ARGS -G ${CMAKE_GENERATOR}
-                -DCMAKE_BUILD_TYPE=${gtestbuild}
+                -DCMAKE_BUILD_TYPE=$<CONFIG>
                 -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
                 -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
                 -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
--- a/lib/Interpreter/Paths.cpp
+++ b/lib/Interpreter/Paths.cpp
@@ -17,7 +17,9 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 
+#if defined(LLVM_ON_UNIX)
 #include <dlfcn.h>
+#endif
 
 namespace Cpp {
 namespace utils {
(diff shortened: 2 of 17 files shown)


lib/Interpreter/Paths.cpp
```diff
@@ -91,15 +91,13 @@ bool GetSystemLibraryPaths(llvm::SmallVectorImpl<std::string>& Paths) {
...
-#define PATH_MAXC (PATH_MAX+1)
-
 std::string NormalizePath(const std::string& Path) {
- char Buf[PATH_MAXC];
-  if (const char* Result = ::realpath(Path.c_str(), Buf))
-    return std::string(Result);
 
-  ::perror("realpath");
-  return std::string();
+  clang::SmallVector<char, MAX_PATH> Buffer;
```
vgvassilev, line 96:
```suggestion
  llvm::SmallVector<char, MAX_PATH> Buffer;
```

vgvassilev, line 96:
```suggestion
  llvm::SmallVector<char, MAX_PATH> Buffer;
```

vgvassilev:
Can you rebase this PR?

mcbarton:
@fsfod In one of your future commits can you fix a mistake I made in the readme build instructions in my last PR. The following line
https://github.com/fsfod/CppInterOp/blob/50a7565a9c10deea39735ff02a2e0912e53ce874/README.md?plain=1#L82
should say
```
git apply -v ../patches/llvm/clang17-*.patch
```
In my haste I copied from the ci and forgot to change it.

fsfod:
sure

fsfod:
Its now just failing to find google test libs.

mcbarton:
@fsfod I also see this error in the CI. LINK : fatal error LNK1104: cannot open file 'pthread.lib' . pthreads seem to be unix only, so this also needs to be solved.

vgvassilev:
> Its now just failing to find google test libs.

Probably here we need to fix the library extensions: https://github.com/compiler-research/CppInterOp/blob/aa3521b00af1aab1ae36df60865a998ceb9d1dd7/cmake/modules/GoogleTest.cmake#L3-L7

vgvassilev:
> Where does the lib path come from for linking thoses libs because it doesn't match that list. Locally it fails with `fatal error LNK1104: cannot open file '..\googletest-prefix\src\googletest-build\lib\\gtest.lib' `

It is probably in `\Debug|Release\lib`?

mcbarton:
> Where does the lib path come from for linking thoses libs because it doesn't match that list. Locally it fails with `fatal error LNK1104: cannot open file '..\googletest-prefix\src\googletest-build\lib\\gtest.lib' `

I believe the gtest library location is being set here on Windows. We are using the MSVC compiler in the CI, so I believe this is where the library path is being set. I'm not 100% certain though.

if(MSVC)
  set(EXTRA_GTEST_OPTS
    -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_DEBUG:PATH=${_gtest_byproduct_binary_dir}/lib/
    -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_MINSIZEREL:PATH=${_gtest_byproduct_binary_dir}/lib/
    -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELEASE:PATH=${_gtest_byproduct_binary_dir}/lib/
    -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELWITHDEBINFO:PATH=${_gtest_byproduct_binary_dir}/lib/
    -Dgtest_force_shared_crt=ON
    BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config Release)

vgvassilev:
> > Where does the lib path come from for linking thoses libs because it doesn't match that list. Locally it fails with `fatal error LNK1104: cannot open file '..\googletest-prefix\src\googletest-build\lib\\gtest.lib' `
> 
> It is probably in `\Debug|Release\lib`?

Maybe something like: `${_gtest_byproduct_binary_dir}/${CMAKE_CFG_INTDIR}/lib/`

fsfod:
> > Where does the lib path come from for linking thoses libs because it doesn't match that list. Locally it fails with `fatal error LNK1104: cannot open file '..\googletest-prefix\src\googletest-build\lib\\gtest.lib' `
> 
> I believe the gtest library location is being set here on Windows. We are using the MSVC compiler in the CI, so I believe this is where the library path is being set. I'm not 100% certain though.
> 
> [fsfod/CppInterOp@`b59af22`/cmake/modules/GoogleTest.cmake#L10C1-L17C74](https://github.com/fsfod/CppInterOp/blob/b59af225385cda1bf1936507eea8be07e2d71240/cmake/modules/GoogleTest.cmake#L10C1-L17C74)

All those paths for point to `CppInterOp/build/Clang-Release/downloads/googletest-prefix/src/googletest-build/lib/` where the libs are for me.

mcbarton:
The cling build of CppInterOp is now getting a large number of 'unresolved external symbol' errors after the last commit. Its not clear to me what may be causing it.

mcbarton:
All the gtest import paths are being set here. 
https://github.com/fsfod/CppInterOp/blob/b59af225385cda1bf1936507eea8be07e2d71240/cmake/modules/GoogleTest.cmake#L81C1-L81C1

Maybe try changing in all the 4 paths 
${_G_LIBRARY_PATH}
to 
${_gtest_byproduct_binary_dir}

fsfod:
I just figured that out, also ${CMAKE_STATIC_LIBRARY_PREFIX} is empty for windows

fsfod:
I don't get any of those linker errors locally that there failing with now. Are the tests relying on symbol visibility for shared libraries being public? because for windows it private default and class or functions need to be explicitly annotated __declspec(dllexport) or the linker has to be passed file of mangled symbol names to be exported.

vgvassilev:
Yes. LLVM has export_executable_symbols() but I think that’s mostly for executables.

vgvassilev:
How is the symbol export solved for the llvm libraries on windows?

fsfod:
Maybe the CI llvm build could be rebuilt with -DBUILD_SHARED_LIBS=OFF for windows. 
Some other alternatives are:
- Decorate the functions declared in CppInterOp.h/CppInterOp.cpp with macros that are turned in to __decl(dllexport) and __decl(dllimport) for windows, but this doesn't solve import errors for clang::Value.
- Use libclang or even add a new public API in CppInterOp to get the mangled names of functions thats used in a script to gather functions declared in CppInterOp.h to create a [linker def file](https://learn.microsoft.com/en-us/cpp/build/reference/exports?view=msvc-170) that contains the mangled export list.

vgvassilev:
> Maybe the CI llvm build could be rebuilt with -DBUILD_SHARED_LIBS=OFF for windows.

That makes sense. CppInterOp should encapsulate all symbols and not dynamically link anything llvm/clang-specific. I've cleaned the cache.

Unrelated, could you @mcbarton and @fsfod reach out to me by email off-list?

mcbarton:
If the ci build passes now it will fix more than just Windows builds. I have done some checks locally and this PR fixes https://github.com/compiler-research/CppInterOp/issues/175 . I am currently checking to see whether this error still occurs or not https://github.com/compiler-research/xeus-clang-repl/issues/78 . If it fixes this issue too then a ci should be able to pass the same level on arm as it does on x86 . I shall know shortly.

mcbarton:
Unfortunately this PR didn't manage to fix https://github.com/compiler-research/xeus-clang-repl/issues/78 too, but this PR does still fix https://github.com/compiler-research/CppInterOp/issues/175 in its current state, which is an added bonus.

vgvassilev:
We might need some inspiration from maybe here: https://github.com/root-project/root/blob/7a853eb2bdee6d9d247adfded6712d7202f0adbe/cmake/modules/SearchInstalledSoftware.cmake#L1930-L1941

cmake/modules/GoogleTest.cmake
```diff
@@ -32,7 +34,7 @@ ExternalProject_Add(
   #            -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELEASE:PATH=ReleaseLibs
   #            -Dgtest_force_shared_crt=ON
   CMAKE_ARGS -G ${CMAKE_GENERATOR}
-                -DCMAKE_BUILD_TYPE=Release
+                -DCMAKE_BUILD_TYPE=${gtestbuild}
```
vgvassilev, line 37:
```suggestion
                -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
```

It's better for unix as well.

fsfod: I just assumed google test was slow to run when built in debug config so it was always set to release

vgvassilev: Probably you are right but maybe we should be consistent across platforms...

vgvassilev:
We can probably disable the failing tests for now and consider this PR success...

fsfod:
The DynamicLibraryManager test mostly just needed different binary output paths in cmake to fix it.

mcbarton:
I also agree that I think its probably best to comment out the failing tests, renaming this PR as fix Windows builds issues, and raise an issue about the failing tests. If you think you can solve the test issues on your local machine, put in a PR which fixes that issue, once they are all passing locally.

fsfod:
I think all of the test executables are crashing early with an assert in llvm code so the tests executables can't be run by windows builds in the CI if you want it to pass atm.

vgvassilev:
Where do you see this assert? Tomorrow I will run clang-format on each commit individually and merge the pr. We can continue addressing the rest of the problems separately.

fsfod:
https://github.com/compiler-research/CppInterOp/actions/runs/7588907781/job/20672453834?pr=181#step:21:248 
`Assertion failed: DC->getLexicalParent() == CurContext && "The next DeclContext should be lexically contained in the current one.", file D:\a\CppInterOp\CppInterOp\llvm-project\clang\lib\Sema\SemaDecl.cpp, line 1347`

vgvassilev:
We might need something like https://github.com/llvm/llvm-project/blob/cd05ade13a66d4fb2daff489b2c02ac1ae2070d1/clang/unittests/Interpreter/InterpreterTest.cpp#L282

fsfod:
That does seem to be implicitly set for me locally, i guess that could be why lots of the tests pass for me locally.

fsfod:
I wonder target triple the tests default to in the windows CI build.

vgvassilev:
> I wonder target triple the tests default to in the windows CI build.

I don't know but we could add a `-v` to the interpreter arguments and will be probably print it out.

mcbarton:
Probably not the best place to put this message, but sometime later today I will put in a PR to fix a few typos and mistakes in the README.md .

vgvassilev:
> > I wonder target triple the tests default to in the windows CI build.
> 
> I don't know but we could add a `-v` to the interpreter arguments and will be probably print it out.

Yes, let's merge this PR because it started to hit the cache limits in github and will be slow to make progress since it will start evicting caches for the builds...

PR #186, opened 2024-01-20
Specify the build configuration when building on Windows

vgvassilev:
The windows cache artifacts are bigger than the unix ones. This is likely because when building on windows we always need to specify the build configuration.


vgvassilev:
After this PR the caches become 700 mb instead of 3.7 gb.

PR #187, opened 2024-01-21
[ci] Allow rerunning gh actions with debug logging.

vgvassilev:
In this case we open an ssh session for 30 mins.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -781,16 +781,12 @@ jobs:
         tail -n1 test_xfailed.log
         echo "Return Code: ${RETCODE}"
         exit $RETCODE
+
     - name: Show debug info
-      if: ${{ failure() }}
+      if: ${{ failure() && (runner.os != 'windows') }}
       run: |
         export
         echo $GITHUB_ENV
-    - name: Setup tmate session
-      if: ${{ failure() }}
-      uses: mxschmitt/action-tmate@v3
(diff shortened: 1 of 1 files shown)


mcbarton:
@vgvassilev The follow error needs to be solved with the Windows build CI (occurs when build CppInterOp). 

CMake Error: Invalid value used with --config

Could you please fix with this PR?

My suggestion is to try changing
https://github.com/vgvassilev/CppInterOp/blob/6d0f691da73e8dad01e3ea2ad1bd8a754722c965/.github/workflows/ci.yml#L309

to
 
```
echo "BUILD_TYPE=Release" >> $env:GITHUB_ENV
echo "CODE_COVERAGE=0" >>  $env:GITHUB_ENV
```

vgvassilev:
> @vgvassilev The follow error needs to be solved with the Windows build CI (occurs when build CppInterOp).
> 
> CMake Error: Invalid value used with --config
> 
> Could you please fix with this PR?
> 
> My suggestion is to try changing https://github.com/vgvassilev/CppInterOp/blob/6d0f691da73e8dad01e3ea2ad1bd8a754722c965/.github/workflows/ci.yml#L309
> 
> to
> 
> ```
> echo "BUILD_TYPE=Release" >> $env:GITHUB_ENV
> echo "CODE_COVERAGE=0" >>  $env:GITHUB_ENV
> ```

Thanks for the hint. Let's try it.

Issue #191, opened 2024-01-27
Disable the failing windows build


vgvassilev:
We should disable or ignore the result of the windows builds as they are expected to fail at that moment, until #188 is resolved.

mcbarton:
This has been fixed in https://github.com/compiler-research/CppInterOp/pull/313

PR #192, opened 2024-01-27
Increase Code Coverage

SAtacker:
It does not seem like this function is used.


vgvassilev (approved):
Nice catch! LGTM!

vgvassilev:
The failing windows platforms are expected to fail according to #191.

PR #193, opened 2024-01-27
Replace std::string pointer by reference and remove unused function DLSym

SAtacker:
- Add test for DLOpen

### DIFF
--- a/lib/Interpreter/Paths.cpp
+++ b/lib/Interpreter/Paths.cpp
@@ -106,52 +106,37 @@ namespace platform {
     return std::string(Buffer.str());
   }
 
-#if defined(LLVM_ON_UNIX)
-  static void DLErr(std::string* Err) {
-    if (!Err)
-      return;
-    if (const char* DyLibError = ::dlerror())
-      *Err = DyLibError;
-  }
-
-  void* DLOpen(const std::string& Path, std::string* Err /* = nullptr */) {
-    void* Lib = ::dlopen(Path.c_str(), RTLD_LAZY | RTLD_GLOBAL);
-    DLErr(Err);
-    return Lib;
-  }
-
-  void* DLSym(const std::string& Name, std::string* Err /* = nullptr*/) {
-    if (void* Self = ::dlopen(nullptr, RTLD_GLOBAL)) {
-      // get dlopen error if there is one
-      DLErr(Err);
-      void* Sym = ::dlsym(Self, Name.c_str());
-      // overwrite error if dlsym caused one
-      DLErr(Err);
-      // only get dlclose error if dlopen & dlsym haven't emited one
-      DLClose(Self, Err && Err->empty() ? Err : nullptr);
-      return Sym;
-    }
-    DLErr(Err);
-    return nullptr;
-  }
-
-  void DLClose(void* Lib, std::string* Err /* = nullptr*/) {
-    ::dlclose(Lib);
-    DLErr(Err);
-  }
-#elif defined(_WIN32)
-
-  void* DLOpen(const std::string& Path, std::string* Err /* = nullptr */) {
-    auto lib = llvm::sys::DynamicLibrary::getLibrary(Path.c_str(), Err);
+#if defined(_WIN32)
+  void* DLOpen(const std::string& Path, std::string& Err) {
+    auto lib = llvm::sys::DynamicLibrary::getLibrary(Path.c_str(), &Err);
     return lib.getOSSpecificHandle();
   }
 
-  void DLClose(void* Lib, std::string* Err /* = nullptr*/) {
+  void DLClose(void* Lib, std::string& Err) {
     auto dl = llvm::sys::DynamicLibrary(Lib);
     llvm::sys::DynamicLibrary::closeLibrary(dl);
-    if (Err) {
-      *Err = std::string();
+    if (Err.empty()) {
+      Err = std::string();
     }
+  }
+#elif defined(LLVM_ON_UNIX)
+  static void DLErr(std::string& Err) {
+    if (const char* DyLibError = ::dlerror())
+      Err = std::string(DyLibError);
+  }
+
+  void* DLOpen(const std::string& Path, std::string& Err) {
+    void* Lib = ::dlopen(Path.c_str(), RTLD_LAZY | RTLD_GLOBAL);
+    if (Lib == nullptr) {
+      DLErr(Err);
+      return nullptr;
+    }
+    return Lib;
+  }
+
+  void DLClose(void* Lib, std::string& Err) {
+    ::dlclose(Lib);
+    DLErr(Err);
   }
 #endif
 
--- a/unittests/CppInterOp/CMakeLists.txt
+++ b/unittests/CppInterOp/CMakeLists.txt
@@ -22,7 +22,7 @@ export_executable_symbols(CppInterOpTests)
 
 unset(LLVM_LINK_COMPONENTS)
 
-add_cppinterop_unittest(DynamicLibraryManagerTests DynamicLibraryManagerTest.cpp)
+add_cppinterop_unittest(DynamicLibraryManagerTests DynamicLibraryManagerTest.cpp ${CMAKE_SOURCE_DIR}/lib/Interpreter/Paths.cpp)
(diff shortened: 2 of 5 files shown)


unittests/CppInterOp/DynamicLibraryManagerTest.cpp
```diff
@@ -54,3 +56,18 @@ TEST(DynamicLibraryManagerTest, Sanity) {
...
+  auto dlopen_handle = Cpp::utils::platform::DLOpen(
+      "./unittests/CppInterOp/libTestSharedLib.dylib", err);
+#elif defined(_WIN32)
+  auto dlopen_handle = Cpp::utils::platform::DLOpen(
+      "./unittests/CppInterOp/libTestSharedLib.dll", err);
+#else
+  auto dlopen_handle = Cpp::utils::platform::DLOpen(
+      "./unittests/CppInterOp/libTestSharedLib.so", err);
+#endif
+  (void)dlopen_handle;
```
vgvassilev, line 72:
```suggestion
  EXPECT_TRUE(dlopen_handle);
  EXPECT_TRUE(err.empty());

  // Check the error case:
  Cpp::utils::platform::DLOpen("missing", err);
  // Check the output of err...
```

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
```diff
@@ -54,3 +59,30 @@ TEST(DynamicLibraryManagerTest, Sanity) {
...
+  auto dlopen_handle = Cpp::utils::platform::DLOpen(
+      CPP_INTEROP_STRINGIFY(
+          CPPINTEROP_LIB_TestSharedLib_DIR_PREFIX) "./libTestSharedLib.so",
+      err);
+#endif
+  EXPECT_TRUE(dlopen_handle);
+  EXPECT_TRUE(err.empty());
+  Cpp::utils::platform::DLOpen("missing", err);
+  EXPECT_TRUE(err.find("no such file") != std::string::npos ||
+              err.find("No such file") != std::string::npos);
```
vgvassilev, line 87:
```suggestion
  EXPECT_FALSE(err.empty());
```

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
```diff
@@ -54,3 +59,30 @@ TEST(DynamicLibraryManagerTest, Sanity) {
   // invalidated...
   // EXPECT_FALSE(Cpp::GetFunctionAddress("ret_zero"));
 }
+
+TEST(UtilsPlatform, DLTest) {
+  std::string err = "";
+  // CPPINTEROP_LIB_TestSharedLib_DIR_PREFIX specified by cmake though target
```
vgvassilev, line 65: Is that supposed to compute the full path?

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
SAtacker: I feel `no such file` is more verbose and the exact error we expect.

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
SAtacker: > Is that supposed to compute the full path?

That depends what `${CMAKE_BINARY_DIR}/unittests/bin/$<CONFIG>/` gives, we can assume that it gives the full path.

vgvassilev: Why is this necessary? Can't we get the fullpath to the test executable and find back the library?

SAtacker: I tried to do it but it seemed to vary across arch and platform and configurations.

lib/Interpreter/Paths.cpp
```diff
@@ -106,53 +106,38 @@ namespace platform {
...
-    return Lib;
+  void DLClose(void* Lib, std::string& Err) {
+    auto dl = llvm::sys::DynamicLibrary(Lib);
+    llvm::sys::DynamicLibrary::closeLibrary(dl);
+    if (Err.empty()) {
+      Err = std::string();
+    }
+  }
+#elif defined(LLVM_ON_UNIX)
+  static void DLErr(std::string& Err) {
```
vgvassilev, line 123: Why not returning a string here?

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
SAtacker: If you do not want macros in that we can have a config.h.in file and have cmake configure it.

vgvassilev: I am trying to avoid that but not at the cost of introducing `config.h.in`. I was wondering if we should not try to look for these files in a few places. Like check if the file exists in these couple of places and pass it to the system.

SAtacker: > I am trying to avoid that but not at the cost of introducing `config.h.in`. I was wondering if we should not try to look for these files in a few places. Like check if the file exists in these couple of places and pass it to the system.

We can have a small utility program that uses c++ 17's filesystem and searches for it. Would that work?

lib/Interpreter/Paths.cpp
SAtacker: I did not change that but we can if that is necessary.

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
vgvassilev: I agree, however, no need to C++ 17 to implement that.

lib/Interpreter/Paths.cpp
vgvassilev: ok, let's keep it as is.

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
SAtacker: Alrighty, I shall do it for POSIX and windows separately.

vgvassilev: You can use llvm facilities from llvm::fs and llvm::sys that would avoid this.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ export_executable_symbols(CppInterOpTests)
 
 unset(LLVM_LINK_COMPONENTS)
 
-add_cppinterop_unittest(DynamicLibraryManagerTests DynamicLibraryManagerTest.cpp)
+add_cppinterop_unittest(DynamicLibraryManagerTests DynamicLibraryManagerTest.cpp ${CMAKE_SOURCE_DIR}/lib/Interpreter/Paths.cpp)
```
vgvassilev, line 25: Why did we add this here? That should be provided by the shared object file.

SAtacker: It was because it wasn't getting linked on one of the platforms

SAtacker: PTAL https://github.com/compiler-research/CppInterOp/actions/runs/7944478504/job/21690038529?pr=193#step:20:172 @vgvassilev

lib/Interpreter/Paths.cpp
```diff
@@ -106,53 +106,38 @@ namespace platform {
...
+    auto lib = llvm::sys::DynamicLibrary::getLibrary(Path.c_str(), &Err);
+    return lib.getOSSpecificHandle();
   }
 
-  void* DLOpen(const std::string& Path, std::string* Err /* = nullptr */) {
-    void* Lib = ::dlopen(Path.c_str(), RTLD_LAZY | RTLD_GLOBAL);
-    DLErr(Err);
-    return Lib;
+  void DLClose(void* Lib, std::string& Err) {
+    auto dl = llvm::sys::DynamicLibrary(Lib);
```
vgvassilev, line 116: Can we add a test where we `dlopen` a library with standard `dlopen` and not `DLOpen` and try to close it with `DLClose`?

Issue #188, opened 2024-01-22
What is required for loading a static libraries to provide symbols to the JIT.


fsfod:
If you try to use any of MSVC STL library on windows you end up with missing symbol errors from the JIT.  The failing symbols come from the c++ runtime static library that still needed even if you use the shared runtime library. 
I see there were patches for patches for it using Orc runtime https://reviews.llvm.org/D134615, but it looks like it was never merged.

vgvassilev:
It was probably due to inactivity on the submitter side. Could you resubmit it as a PR against the llvm-project and add me as a reviewer? Also are we sure this patch fixes our problem?

vgvassilev:
Hi @sunho, can you give us a hand -- we are trying to use the orc jit infrastructure on windows and we get missing symbol errors. You can take a look [here](https://github.com/compiler-research/CppInterOp/actions/runs/7603850481/job/20724508805).

vgvassilev:
@fsfod, is that here a good workaround that would help us make progress with this issue: https://github.com/root-project/root/blob/12ebada154cf640fcf02a39dd0f17412dc59230d/core/metacling/src/CMakeLists.txt#L138-L189

vgvassilev:
Hi @bellenot, apologies for summoning you on a random issue but we are stuck with our CppInterOp library on windows. In principle the problem should exist in ROOT as well but we somehow solve it... Could you give us a hint how to export symbols from MSVC STL library? You can take a look how things fail [here](https://github.com/compiler-research/CppInterOp/actions/runs/7603850481/job/20724508805).

mcbarton:
@vgvassilev I came across the following which defines how you export symbols on Windows using MSVC https://ms-iot.github.io/ROSOnWindows/Porting/SymbolVisibility.html . Microsoft recommends using Visibility Control Headers to control which symbols are exported from my reading. An alternative way they mention is is a cmake property called WINDOWS_EXPORT_ALL_SYMBOLS which may be a workaround for now. Cmake page for reference https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html . @fsfod Can you try this?

fsfod:
The cling workaround does work, but its brittle to the MSVC version used, like for me the `endl` symbol it exports doesn't exist in my MSVC runtime libs. I don't know if it was added for cling specific reasons because it seems to be declared in a header for MSVC's STL.  
[...]

PR #195, opened 2024-01-28
Add ability to compile CppInterOp and tests on Windows with different compiler

mcbarton:
@vgvassilev This PR allows for those trying to solve Windows build issues to use a different compiler to MSVC . With these changes I was able to compile CppInterOp and its tests using the Clang compiler. Currently I'm not passing the tests as the header files cannot be found, but I believe this has something to do with the environment variables I have set, and am investigating.

Fixes https://github.com/compiler-research/CppInterOp/issues/175 (provides proof that earlier PR fixed it)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,6 +202,66 @@ jobs:
           #  cling: On
           #  cling-version: '1.0'
           #  cppyy: On
+          - name: osx14-arm-clang-clang-repl-17
+            os: macos-14
+            compiler: clang
+            clang-runtime: '17'
+            cling: Off
+            cppyy: Off
+          - name: osx14-arm-clang-clang-repl-17-cppyy
+            os: macos-14
+            compiler: clang
+            clang-runtime: '17'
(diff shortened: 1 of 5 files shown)


vgvassilev:
Can we use `continue-on-error` in the yml file instead?

mcbarton:
@vgvassilev So just to be certain you want me to reactivate the other Windows builds, but add continue-on-error: true to the 'Build and Test/Install CppInterOp on Windows systems' part of the jobs?

vgvassilev:
Yes, assuming we still get some green/ or different than X build sign.

mcbarton:
@vgvassilev After that change the Windows builds currently get a green tick. I have added a comment to this section to comment that tests are expected to fail until https://github.com/compiler-research/CppInterOp/issues/188 is resolved.

mcbarton:
@vgvassilev Can this PR be merged now that is gets a green tick for all the jobs?

vgvassilev (commented):
LGTM! Thank you!

PR #196, opened 2024-02-04
Windows test fixes

fsfod:
Not all tests pass yet, but they shouldn't crash early with a fatal assert from Clang any more. The library tests could fail in assert builds before because an Error value for a from a Coff symbol in Dyld::BuildBloomFilter was not handled, those errors were actually revealing another issue that search path to load shared libraries to find symbols should be much more restricted instead of PATH environment variable. 
I also added the exporting of some MSVC runtime symbols based on the code from Clings cmake files to try and fix STL symbol errors, but tests still fail because of missing emulated TLS symbols `__emutls_get_address and __emutls_v._Init_thread_epoch` .

Current failing tests are:
```
 EnumReflectionTest.GetIntegerTypeFromEnumScope
 EnumReflectionTest.GetIntegerTypeFromEnumType
 FunctionReflectionTest.GetFunctionAddress
 FunctionReflectionTest.Construct
 FunctionReflectionTest.Destruct
```

### DIFF
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -2463,6 +2463,10 @@ namespace Cpp {
     std::vector<const char *> ClingArgv = {"-resource-dir", ResourceDir.c_str(),
                                            "-std=c++14"};
     ClingArgv.insert(ClingArgv.begin(), MainExecutableName.c_str());
+#ifdef _WIN32
+    // FIXME : Workaround Sema::PushDeclContext assert on windows
+    ClingArgv.push_back("-fno-delayed-template-parsing");
+#endif
     ClingArgv.insert(ClingArgv.end(), Args.begin(), Args.end());
     // To keep the Interpreter creation interface between cling and clang-repl
     // to some extent compatible we should put Args and GpuArgs together. On the
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -373,6 +373,55 @@ option(CPPINTEROP_ENABLE_DOXYGEN "Use doxygen to generate CppInterOp interal API
 option(CPPINTEROP_ENABLE_SPHINX "Use sphinx to generage CppInterOp user documentation")
 option(CPPINTEROP_ENABLE_TESTING "Enables the testing infrastructure." ON)
 
+if(MSVC)
+
+  set(MSVC_EXPORTLIST
+   ??_7type_info@@6B@
+   ?nothrow@std@@3Unothrow_t@1@B
+   _Init_thread_header
+   _Init_thread_footer
+   ??_7type_info@@6B@
+   ?_Facet_Register@std@@YAXPEAV_Facet_base@1@@Z
+  )
+
+  set(MSVC_EXPORTLIST ${MSVC_EXPORTLIST}
+   ??2@YAPEAX_K@Z
+   ??3@YAXPEAX@Z
+   ??3@YAXPEAX_K@Z
+   ??_U@YAPEAX_K@Z
+   ??_V@YAXPEAX@Z
+   ??_V@YAXPEAX_K@Z
+   ??2@YAPEAX_KAEBUnothrow_t@std@@@Z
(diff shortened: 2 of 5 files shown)


lib/Interpreter/CppInterOp.cpp
```diff
@@ -2463,6 +2463,10 @@ namespace Cpp {
     std::vector<const char *> ClingArgv = {"-resource-dir", ResourceDir.c_str(),
                                            "-std=c++14"};
     ClingArgv.insert(ClingArgv.begin(), MainExecutableName.c_str());
+#ifdef _WIN32
+    // FIXME : Workaround Sema::PushDeclContext assert on windows
+    ClingArgv.push_back("-fno-delayed-template-parsing");
```
vgvassilev, line 2468: Do we need `-femulated-tls` for the TLS related issues? We do this here in cling:

https://github.com/root-project/root/blob/fda4630190de40d817b850efa240c4ed72a20b8e/interpreter/cling/lib/Interpreter/IncrementalJIT.cpp#L388-L390

fsfod: I tried -fno-emulated-tls there and still got the same emulated TLS symbol errors. It looks like its defaulted to on from JITTargetMachineBuilder's constructor.

vgvassilev: Do you think we need a patch in upstream clang?

fsfod: Maybe I don't really know how it should work for windows.

vgvassilev:
@fsfod, forgot to say - can you apply clang-format on per-commit basis or shall I?

mcbarton:
> LGTM! @mcbarton, after this PR we might be able to remove from the CI the make-it-all-green for windows and disable only the failing tests.

@vgvassilev This would not be possible if it was merged right now, as when CppInterOp uses Cling on Windows it currently aborts in the following tests in the CI
- InterpreterTest.Process
- DynamicLibraryManagerTest.Sanit

fsfod:
There are two tests that fail that are not export related:
```
[ RUN      ] EnumReflectionTest.GetIntegerTypeFromEnumScope
EnumReflectionTest.cpp(127): error: Expected equality of these values:
  Cpp::GetTypeAsString(Cpp::GetIntegerTypeFromEnumScope(Decls[4]))
    Which is: "int"
  "unsigned int"
[ RUN      ] EnumReflectionTest.GetIntegerTypeFromEnumType
EnumReflectionTest.cpp(183): error: Expected equality of these values:
  get_int_type_from_enum_var(Decls[10])
    Which is: "int"
  "unsigned int"
```
MSVC defaults to [int as the base type for enums](https://learn.microsoft.com/en-us/cpp/build/reference/zc-enumtypes?view=msvc-170#remarks) ;

vgvassilev:
> There are two tests that fail that are not export related:
> 
> ```
> [ RUN      ] EnumReflectionTest.GetIntegerTypeFromEnumScope
> EnumReflectionTest.cpp(127): error: Expected equality of these values:
>   Cpp::GetTypeAsString(Cpp::GetIntegerTypeFromEnumScope(Decls[4]))
>     Which is: "int"
>   "unsigned int"
> [ RUN      ] EnumReflectionTest.GetIntegerTypeFromEnumType
> EnumReflectionTest.cpp(183): error: Expected equality of these values:
>   get_int_type_from_enum_var(Decls[10])
>     Which is: "int"
>   "unsigned int"
> ```
> 
> MSVC defaults to [int as the base type for enums](https://learn.microsoft.com/en-us/cpp/build/reference/zc-enumtypes?view=msvc-170#remarks) ;

Ok, so the tests need to be adjusted for windows...

fsfod:
> Any idea what is causing these tests to fail (and subsequently abort the test runs) in the Cling version , but not the Clang-Repl version?
Maybe memory was double free'ed in cling::Value or it was uninitialized data being free'ed?

vgvassilev:
Let’s move forward here and address the remaining issues in a subsequent PR.

PR #197, opened 2024-02-06
Remove redundant ci jobs

mcbarton:
@vgvassilev The non cppyy or xeus-clang-repl jobs for osx and Ubuntu are redundant. They don't do anything that the cppyy and xeus-clang-repl jobs don't (there was one small valgrind dependence for Ubuntu but this has been removed in this PR). I have removed them so that free Github runner minutes are not wasted, and so we can run through the osx jobs quicker (the maximum number of concurrent osx jobs allowed is 5, so we are currently have to wait longer than necessary for osx PR jobs to run).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,13 +21,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: ubu22-x86-gcc12-clang-repl-17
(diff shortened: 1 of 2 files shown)


mcbarton:
@alexander-penev can you review this PR for me?

alexander-penev (approved):

Issue #183, opened 2024-01-18
Add WASM CI Support


vgvassilev:
The wasm support should be compatible to what's already done for the [xeus-cpp](https://github.com/compiler-research/xeus-cpp) project where I would like to eventually adopt CppInterOp.

mcbarton:
@vgvassilev I'm just waiting for the Windows CppInterOp build to pass in https://github.com/compiler-research/CppInterOp/pull/181 and be merged, and then I plan add on a Windows cppyy build to the PR. I'll try to address this WASM support in that PR.

vgvassilev:
Ok, awesome!

vgvassilev:
Looks like the windows work is almost done. Maybe a good time to start the WASM work.

PR #198, opened 2024-02-12
[CI] Add wasm support to CI for osx and Linux

mcbarton:
The purpose of this PR is to add wasm support to CppInterOp
Fixes https://github.com/compiler-research/CppInterOp/issues/183

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -433,18 +433,21 @@ jobs:
         mkdir build
         if [[ "${cling_on}" == "ON" ]]; then
           cd build
-          cmake -DLLVM_ENABLE_PROJECTS=clang                  \
-                -DLLVM_EXTERNAL_PROJECTS=cling                \
-                -DLLVM_EXTERNAL_CLING_SOURCE_DIR=../../cling  \
-                -DLLVM_TARGETS_TO_BUILD="host;NVPTX"          \
-                -DCMAKE_BUILD_TYPE=Release                    \
-                -DLLVM_ENABLE_ASSERTIONS=ON                   \
-                -DLLVM_ENABLE_LLD=ON                          \
-                -DCLANG_ENABLE_STATIC_ANALYZER=OFF            \
-                -DCLANG_ENABLE_ARCMT=OFF                      \
-                -DCLANG_ENABLE_FORMAT=OFF                     \
-                -DCLANG_ENABLE_BOOTSTRAP=OFF                  \
+          cmake -DLLVM_ENABLE_PROJECTS="clang;lld"               \
+                -DLLVM_EXTERNAL_PROJECTS=cling                     \
+                -DLLVM_EXTERNAL_CLING_SOURCE_DIR=../../cling       \
+                -DLLVM_TARGETS_TO_BUILD="WebAssembly;host;NVPTX"   \
+                -DCMAKE_BUILD_TYPE=Release                         \
+                -DLLVM_ENABLE_ASSERTIONS=ON                        \
+                -DCLANG_ENABLE_STATIC_ANALYZER=OFF                 \
+                -DCLANG_ENABLE_ARCMT=OFF                           \
+                -DCLANG_ENABLE_FORMAT=OFF                          \
+                -DCLANG_ENABLE_BOOTSTRAP=OFF                       \
+                -DLLVM_ENABLE_ZSTD=OFF                             \
+                -DLLVM_ENABLE_TERMINFO=OFF                         \
+                -DLLVM_ENABLE_LIBXML2=OFF                          \
                 ../llvm
+          cmake --build . --target lld --parallel ${{ env.ncpus }}
           cmake --build . --target clang --parallel ${{ env.ncpus }}
           cmake --build . --target cling --parallel ${{ env.ncpus }}
           # Now build gtest.a and gtest_main for CppInterOp to run its tests.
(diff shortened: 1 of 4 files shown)


vgvassilev:
Isn't there some emscripten forge thing where we can pull some dependencies in?

mcbarton:
> Isn't there some emscripten forge thing where we can pull some dependencies in?

Likely, but I couldn't find a way searching online. I'm not too familiar with emscripten, so that is the reason I tried to build my own wasm zlib, and find it that way like I did locally. That doesn't seem to be working on the Github runner though.

mcbarton:
@vgvassilev I'll try again later using an environment yaml file like in xeus-cpp. Maybe https://repo.mamba.pm/emscripten-forge might have everything we will need, since it has zlib , which is the current stumbling block.

vgvassilev:
> @vgvassilev I'll try again later using an environment yaml file like in xeus-cpp. Maybe https://repo.mamba.pm/emscripten-forge might have everything we will need, since it has zlib , which is the current stumbling block.

Can’t we try to get the llvm binaries from there?

mcbarton:
> > @vgvassilev I'll try again later using an environment yaml file like in xeus-cpp. Maybe https://repo.mamba.pm/emscripten-forge might have everything we will need, since it has zlib , which is the current stumbling block.
> 
> Can’t we try to get the llvm binaries from there?

The only reason I didn't is I because of the patches on clang 16 and 17 that CppInterOp uses.. Is this not an issue? I can add this to the environment yml file if not, as this appears to be working now. I just need to understand why it can't find the wasm installed dependencies. I can hardcode in the paths for now until I understand why.

mcbarton:
@vgvassilev Now that I have the configure and build stage for the wasm build passing, can you delete the llvm cache for this PR? I want to see if I can get the wasm and non wasm builds using the same llvm cache. There doesn't appear to be any reason they can't use the same one.

vgvassilev:
The cache for this PR is deleted now.

mcbarton:
> The cache for this PR is deleted now.

My cache was not deleted. I just went to rebuild, and it restored a llvm build from cache.

vgvassilev:
> > The cache for this PR is deleted now.
> 
> @vgvassilev My cache was not deleted. I just went to rebuild the cache, and it restored a llvm build from cache.

Strange. Now I deleted everything.

mcbarton:
@vgvassilev This PR is ready for review.

vgvassilev:
Thank you! I am afraid I will need help to review this one. Maybe @anutosh491 can help?

mcbarton:
@anutosh491 are you able to review this PR for me?

anutosh491:
Ahh sorry for the delay on this, going through it now.

.github/workflows/ci.yml
```diff
@@ -907,3 +912,275 @@ jobs:
...
+        path: |
+          llvm-project
+          ${{ matrix.cling=='On' && 'cling' || '' }}
+        key: ${{ steps.cache.outputs.cache-primary-key }}
+        
+    - name: Emscripten build of CppInterOp on Unix systems
+      if: ${{ runner.os != 'windows' }}
+      shell: bash -l {0}
+      run: |
+        emsdk activate latest
```
anutosh491, line 1116: Let's use emscripten `3.1.45` ? 
I guess we should only use `3.1.45` ([reference](https://github.com/search?q=repo%3Aemscripten-forge%2Frecipes+3.1.45&type=code))
(ie the same one as in `emscripten-forge`) . All repos like [xeus-cpp](https://github.com/compiler-research/xeus-cpp/blob/216dbed46e358d82fd8fe3065b4533f3e1c49119/.github/workflows/main.yml#L156), [xeus-python](https://github.com/jupyter-xeus/xeus-python/blob/1c55a64050ad43e02611e0a77f2bd1a2a6a00bf5/.github/workflows/main.yml#L167) etc follow that as we have dependencies based on emscripten-forge.

CMakeLists.txt
```diff
@@ -371,7 +371,15 @@ endif()
 option(CPPINTEROP_INCLUDE_DOCS "Generate build targets for the CppInterOp docs.")
 option(CPPINTEROP_ENABLE_DOXYGEN "Use doxygen to generate CppInterOp interal API documentation.")
 option(CPPINTEROP_ENABLE_SPHINX "Use sphinx to generage CppInterOp user documentation")
-option(CPPINTEROP_ENABLE_TESTING "Enables the testing infrastructure." ON)
+
+
+if(EMSCRIPTEN)
+    message("Build with emscripten")
+    option(CPPINTEROP_ENABLE_TESTING "Enables the testing infrastructure." OFF)
```
anutosh491, line 378: Any concrete reason, we're not planning to enable testing here ?

.github/workflows/ci.yml
mcbarton: I will make this change sometime today.

CMakeLists.txt
mcbarton: This was done for no other reason than that was what was done in xeus-cpp (i.e. testing turned off by default when emscripten is on) https://github.com/compiler-research/xeus-cpp/blob/216dbed46e358d82fd8fe3065b4533f3e1c49119/CMakeLists.txt#L77C1-L77C34 .

anutosh491: Works can skip that for now !

vgvassilev:
> @vgvassilev The jobs failing after the suggested change to emscripten 3.1.45 are due to a cache issue as far as I can tell. Can you clear the cache and manually reactivate the workflow? It should get all green checks after this.

@mcbarton, the cache is cleaned. Maybe you can restart?

mcbarton:
@anutosh491 do you have any comments on the PR, or do you approve it?

anutosh491 (approved):
I think this looks good. Maybe the hardcoding of PREFIX path can be looked into in the future.

vgvassilev:
Thanks @anutosh491. Well done @mcbarton. Now we need to figure out how to plug this into the conda recipe [here](https://github.com/conda-forge/cppinterop-feedstock)...

PR #218, opened 2024-04-05
Update GetFunctionReturnType API for template functions

maximusron:


lib/Interpreter/CppInterOp.cpp
```diff
@@ -822,6 +822,10 @@ namespace Cpp {
         return FD->getReturnType().getAsOpaquePtr();
     }
 
+    if (auto* FD = llvm::dyn_cast_or_null<clang::FunctionTemplateDecl>(D)) {
```
vgvassilev, line 825: Can D be null here? If not we need to drop the _or_null part.

vgvassilev (approved):
LGTM!

PR #222, opened 2024-04-07
Only run clang-tidy-review on request in PR's

maximusron:
Prevents the actions bot from spamming PR's with comments. This migrates the clang-tidy-review to be performed only when requested by detecting "clang-tidy-review" in the comment body of the PR.


vgvassilev:
Shouldn't we suppress the false positives for the project that's issued by clang-tidy and address all of the rest of its comments?

PR #223, opened 2024-04-08
Make running emscripten_wasm dependent on sucessfull run of build in ci

mcbarton:
This PR makes running emscripten_wasm dependent on sucessfull run of build in ci. This should reduce ci resources being used uncessarily, to be used by other PRs in the organisation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -977,6 +977,7 @@ jobs:
       timeout-minutes: 30
 
   emscripten_wasm:
+    needs: [build]
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
     strategy:


mcbarton:
@vgvassilev This is a minor change to the way the ci works to help alleviate some of the issues which occur when multiple PRs are put in. It stops the wasm build being check unless the non warm build passes sucessfully.

vgvassilev:
That probably lifts the cache issue and replaces it with longer build times. I think it is probably fine. FWIW, your CI magic becomes quite advanced and I am not sure if I can review it ;) but it looks good to me.

@maximusron what do you think?

maximusron:
@mcbarton @vgvassilev this looks good and should help. I think we can eventually figure out similar conditional additions that can run less builds that are expected to fail, if things that precede the dependecy chain don't build. This way we can aim for a more ideal build chain, and combined with reusing the cache, should optimise the CI overall

PR #227, opened 2024-04-09
[ci] Reduce bot lag by building cache first

mcbarton:
@vgvassilev This should reduce the bot lag as much as https://github.com/compiler-research/CppInterOp/pull/226 while keeping all the jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,443 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  build_cache:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: ubu22-x86-gcc12-clang-repl-18
+            os: ubuntu-22.04
+            compiler: gcc-12
+            clang-runtime: '18'
+            cling: Off
+            cppyy: On
+            coverage: true
+          - name: ubu22-x86-gcc12-clang-repl-17
+            os: ubuntu-22.04
+            compiler: gcc-12
+            clang-runtime: '17'
+            cling: Off
+            cppyy: On
+            coverage: true
+          - name: ubu22-x86-gcc9-clang-repl-16
+            os: ubuntu-22.04
+            compiler: gcc-9
+            clang-runtime: '16'
+            cling: Off
+            cppyy: On
+            coverage: true
+          - name: ubu22-x86-gcc9-clang13-cling
+            os: ubuntu-22.04
+            compiler: gcc-9
+            clang-runtime: '13'
+            cling: On
+            cling-version: '1.0'
+            cppyy: On
+            coverage: true
+          #FIXME: Windows CppInterOp tests expected to fail
+          #until https://github.com/compiler-research/CppInterOp/issues/188 is solved
+          - name: win2022-msvc-clang-repl-18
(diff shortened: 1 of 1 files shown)


mcbarton:
image

The CI is now in 3 stages. The first builds the caches if it cannot find them. It then builds the non wasm builds, and if they pass all tests, it then goes onto build the wasm builds. Eventually I hope to be able to make use of this https://github.com/github/roadmap/issues/826 so that we are not dependent on a cache at all, and can work with vms which have the patched llvms included.

vgvassilev:
Do we really need the arm builds? Maybe arm for the latest clang maybe not even that…

mcbarton:
@vgvassilev My suggestion is to keep the osx arm builds, but drop the x86 ones. If it works on the arm then it is likely to work on the x86. The issue with the osx builds is that you can only have 5 running simultaneously over the whole organisation. If we drop osx x86 and merge the xeus-clang-repl  and cppyy jobs, then the number of osx jobs drop in CppInterOp reduces to a quarter of its current value.

mcbarton:
osx arm is more important that osx x86 going forward now that Apple has moved to an Arm based architechture. In a few years news osx operating systems won't even support x86.

mcbarton:
@vgvassilev Will changing the build type for the llvm from Release to MinSizeRel make much difference to the size it takes up in the cache? I don't have much an idea of how much the different build types take up.

vgvassilev:
> @vgvassilev Will changing the build type for the llvm from Release to MinSizeRel make much difference to the size it takes up in the cache? I don't have much an idea of how much the different build types take up.

It will make a difference if we build only the clang and clang-repl targets as done in my PR.

mcbarton:
> > @vgvassilev Will changing the build type for the llvm from Release to MinSizeRel make much difference to the size it takes up in the cache? I don't have much an idea of how much the different build types take up.
> 
> It will make a difference if we build only the clang and clang-repl targets as done in my PR.

I already made these changes in  this PR (with the also needed lld which I mention in the other PR)

mcbarton:
After @maximusron comment in the other PR about suggesting against merging the xeus-clang-repl and cppyy jobs I won't be doing this now, but I still think its better to drop the osx x86 instead of the osx arm ones.

maximusron:
Looks good to me, one comment would be that I don't think its a good idea to combine them since logs can get cluttered and we usually would prefer a cleaner test environment. This can also entangle dependencies in the long run and have something fail because their requirements clash.

They can happen at the same stage in the CI but should be seperate jobs

maximusron:
@mcbarton yeah the OS X builds seem to be excessive, reducing the number of osx x86 builds seems like a good idea to me

mcbarton:
> @mcbarton yeah the OS X builds seem to be excessive, reducing the number of osx x86 builds seems like a good idea to me (maybe a few as a sanity check)

Tomorrow morning i'll add another commit to this PR removing all but one of the osx x86 builds as the sanity check (probably the latest llvm, with the cpppyy option as the xeus-clang-repl only currently checks the config file isn't wrong, which is done on the arm side).

vgvassilev:
There is a problem with the wasm builds. Shall we merge to start rebuilding the caches in the master and reuse them?

mcbarton:
@vgvassilev clear the cache, merge this PR and then I will look at the output of the ci to determine the problem with the wasm builds.

vgvassilev:
Done

PR #228, opened 2024-04-10
Fix issue with wasm build cache missing

mcbarton:
@vgvassilev This PR should fix the cache issue with the one wasm job. Not sure why it has only just started failing, but every job should pass now,

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,9 +35,9 @@ jobs:
             cling: Off
             cppyy: On
             coverage: true
-          - name: ubu22-x86-gcc9-clang-repl-16
+          - name: ubu22-x86-gcc12-clang-repl-16
             os: ubuntu-22.04
-            compiler: gcc-9
+            compiler: gcc-12
             clang-runtime: '16'
             cling: Off
             cppyy: On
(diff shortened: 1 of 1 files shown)


maximusron:
@mcbarton Looks good, CI is working great now, really like the update! The 2 stage separation and the cache usage makes it easier to look at and faster compared to earlier. If I understand correctly the staged builds for cppyy and xeus show up right after the LLVM build/cache restore job.

I was wondering if we can add the InterOp build to those jobs (tagged without cppyy or xeus) and then use that library instance for all the cppyy/xeus/wasm builds. In theory with any pull request we only need a single platform and clang specific build for InterOp. What I mean to say is that with one ubuntu clang 18 InterOp build we can run all 3 jobs since cppyy, xeus and wasm can use the same library installation (we may have to set some paths). @vgvassilev is this feasible?

I can try to implement this but it might take some time...

mcbarton:
> @mcbarton Looks good, CI is working great now, really like the update! The 2 stage separation and the cache usage makes it easier to look at and faster compared to earlier. If I understand correctly the staged builds for cppyy and xeus show up right after the LLVM build/cache restore job.
> 
This is correct.

> I was wondering if we can add the InterOp build to those jobs (tagged without cppyy or xeus) and then use that library instance for all the cppyy/xeus/wasm builds. In theory with any pull request we only need a single platform and clang specific build for InterOp. What I mean to say is that with one ubuntu clang 18 InterOp build we can run all 3 jobs since cppyy, xeus and wasm can use the same library installation (we may have to set some paths). @vgvassilev is this feasible?
> 
@maximusron I would suggest against dropping down to only one llvm for the cppyy tests, due to issues like https://github.com/compiler-research/xeus-clang-repl/issues/78 which only appear on one llvm. 

I have more ideas on how to speed up the CI further, such as having a stage to build CppInterOp for each llvm, between the cache build and the xeus-clang-repl/cppyy builds so that this work isn't duplicated. I will put these PRs in over the coming week.

maximusron:
Yeah I meant a single build per clang version. I think we are converging on the same idea that is to build CppInterOp on all unique combinations  of platforms and LLVM versions only once.

vgvassilev:
Shall we merge this one?

mcbarton:
Yes this PR is ready to merge.

PR #229, opened 2024-04-11
Reduce cache size

mcbarton:
@vgvassilev can you clear the cache, and then manually rerun the workflow of this PR? It should pass, while having a reduced llvm cache, based on the workflow in xeus-clang-repl.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -365,7 +365,9 @@ jobs:
                 ../llvm
           cmake --build . --target clang clang-repl lld --parallel ${{ env.ncpus }}
         fi
-        cd ../../
+        cd ../
+        rm -rf $(find . -maxdepth 1 ! -name "build" ! -name "llvm" ! -name "clang" ! -name ".")
+        cd ../
 
     - name: Build LLVM/Cling on Windows systems if the cache is invalid
       if: ${{ runner.os == 'windows' && steps.cache.outputs.cache-hit != 'true' }}


vgvassilev:
Why are we still building a cache for clang-18? Should we not get it via the package manager?

mcbarton:
> Why are we still building a cache for clang-18? Should we not get it via the package manager?

We can get rid of it via a package manager, but not for all systems yet (osx Homebrew doesn't currently have llvm 18, but its in progress). Its also not available through Emscripten forge yet for the web assembley. I think we'll be able to remove the llvm 18 cache, but not for a few weeks. This PR should significantly reduce the size of the cache if it works. The cache will also be reduce even more when I put in my PR which removes the osx x86 jobs.

mcbarton:
@vgvassilev I won't be able to be certain this PR works until the cache is cleared and the workflow is run manually.

vgvassilev:
Yes, just waiting for a job to finish and I will merge it.

mcbarton:
> Yes, just waiting for a job to finish and I will merge it.

I wouldn't merge this one yet until its been able to be tested. I'm not 100 % sure this PR works.

vgvassilev:
> > Yes, just waiting for a job to finish and I will merge it.
> 
> I wouldn't merge this one yet until its been able to be tested. I'm not 100 % sure this PR works.

I reran with the cache cleared and indeed you are right we might need to improve this PR.

vgvassilev:
Done.

mcbarton:
@vgvassilev This PR is ready for merging. I recommend clearing the cache before merging.

PR #230, opened 2024-04-12
Tests for static GetClassDecls

maximusron:
This tests both interfaces (GetClassMethods, GetFunctionTemplatedDecls) that use the GetClassDecls static function simultaneously under the same scope


vgvassilev (approved):
LGTM!

PR #233, opened 2024-04-12
[ci] Cppyy tests - fix pytest failing

maximusron:


mcbarton:
@maximusron The cppyy tests on osx arm still don't run after this change, but the osx x86 do (discluding the cling based one). It shouldn't be an issue, but can you turn off the parallel running of the tests for a run, so we can rule that out as an issue?

vgvassilev:
When a test crashes it crashes it cannot run the subsequent test cases. The pydist did that in a distributed way so that each test case gets run in isolation and if it crashes then we still get to see if others succeeded...

maximusron:
@vgvassilev Maybe once the checks are done we can merge this so that we can see the test stats on main

mcbarton:
@maximusron can you rebase and fix the comment I made about apt not being present on osx in this PR? Without it the packges you added for tests on Ubuntu won't be present on osx https://github.com/compiler-research/CppInterOp/pull/231

.github/workflows/ci.yml
```diff
@@ -1156,9 +1161,6 @@ jobs:
       continue-on-error: true
```
mcbarton, line 1161: @maximusron Maybe worth removing this line as part of this PR, so its clearer when this PR is a sucess without consulting the ci output.

vgvassilev:
I do not understand what’s the intent of removing pydist. The other changes seem fine.

maximusron:
> I do not understand what’s the intent of removing pydist. The other changes seem fine.

pydist is no longer being removed, that commit has been dropped

.github/workflows/ci.yml
maximusron: This will probably error out very fast, even if everything runs smoothly because the failing cppyy tests output std::error's to the console.

vgvassilev: Ok, let's do that in a separate PR then..

PR #236, opened 2024-04-14
Reduce warnings ci

mcbarton:
@vgvassilev can you merge this PR? It reduces the warnings outputted by the ci by updating to the latest version of various Github actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,8 +133,6 @@ jobs:
         echo "CLING_HASH=$CLING_HASH" >> $GITHUB_ENV
         echo "LLVM_HASH=$LLVM_HASH" >> $GITHUB_ENV
 
-    - uses: nelonoel/branch-name@v1.0.1
-
     - name: Save PR Info on Windows systems
       if: ${{ runner.os == 'windows' }}
       run: |
@@ -162,8 +160,6 @@ jobs:
         
         echo "CLING_HASH=$env:CLING_HASH" >> $GITHUB_ENV
         echo "LLVM_HASH=$env:LLVM_HASH" >> $GITHUB_ENV
-        
-    - uses: nelonoel/branch-name@v1.0.1
 
     - name: Setup default Build Type on *nux
       if: runner.os != 'windows'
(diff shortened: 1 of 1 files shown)


mcbarton:
@vgvassilev can you take a look at this PR? The changes are needed due to Githubs move from node 12 to node 16 (don't know exactly what this means). Without these changes the workflow will eventually fail to run. The remaining warnings regarding this issue will be solved in a subsequent PR. Its not clear which is the best strategy to deal with the issue in regards of the Github action nelonoel/branch-name@v1.0.1 . This action seems to have been abandoned so won't be updated to fix the issue. Given it simply returns the branch name then its probably just worth replacing with our own action to do this, or integrate this into one of the other actions in the workflow.

alexander-penev:
> @vgvassilev ... Its not clear which is the best strategy to deal with the issue in regards of the Github action nelonoel/branch-name@v1.0.1 . This action seems to have been abandoned so won't be updated to fix the issue. Given it simply returns the branch name then its probably just worth replacing with our own action to do this, or integrate this into one of the other actions in the workflow.

I'm not sure, but this BRANCH_NAME seems to be long gone. If it is still needed, perhaps it can be taken from GITHUB_REF, if of course such a variable exists in the environment, similar to what they do in the abandoned action: [nelonoel/branch-name](https://github.com/nelonoel/branch-name/blob/f2d36c338e7878c908e40e376f0df2c4a5cadf69/index.js#L5C26-L5C37)

mcbarton:
> > @vgvassilev ... Its not clear which is the best strategy to deal with the issue in regards of the Github action nelonoel/branch-name@v1.0.1 . This action seems to have been abandoned so won't be updated to fix the issue. Given it simply returns the branch name then its probably just worth replacing with our own action to do this, or integrate this into one of the other actions in the workflow.
> 
> I'm not sure, but this BRANCH_NAME has been out of use for a long time. If it is still needed, perhaps it can be taken from GITHUB_REF, if of course such a variable exists in the environment, similar to what they do in the abandoned action: [nelonoel/branch-name](https://github.com/nelonoel/branch-name/blob/f2d36c338e7878c908e40e376f0df2c4a5cadf69/index.js#L5C26-L5C37)

I'll perform a run without this action, to see whether we even need it anymore, and if so where.

mcbarton:
@alexander-penev The workflow ran successfully without the branch-name action. Can you merge this PR?

alexander-penev (approved):

PR #238, opened 2024-04-23
[Interpreter] Forward declare input to process so it's registered in JIT

maximusron:
This PR ensures that previous PTU's as a result of just parsing are registered in the JIT. This fixes failing tests with symbol not found, which was causing the wrapper code in TClingCallFunc to not compile.


vgvassilev (approved):
LGTM!

PR #239, opened 2024-04-24
[ci] Report failure in cppyy tests except runxfail/valgrind

maximusron:


maximusron:
OS X arm jobs break while testing cppyy with ` libc++abi: terminating due to uncaught exception of type int`. On cppyy-backend, we have kept the x86 jobs which atleast run the entire test suite (with 33 failures with libc++ not handling std library types correctly). 

I would propose not spending too much time on the OS X jobs as they seem to be a time sink... This PR fixes the regression of the cppyy workflow with clang 16,17 and 18 on Linux

PR #250, opened 2024-04-30
Update codecov-action to v4 (node 16 to node 20 migration)

mcbarton:
An action was missed when upgrading them in https://github.com/compiler-research/CppInterOp/pull/236 due to the migration of node 16 to node 20 in Github actions


vgvassilev:
Thanks!

PR #253, opened 2024-05-02
[ci] Update OS X and clang 16 exitcodes

maximusron:


mcbarton:
You'll want to rebase this PR, so that future commits go quickier, since you no longer need to do the xeus-clang-repl jobs.

PR #254, opened 2024-05-02
Add llvm libunwind callback to suppress exceptions on apple silicon

aaronj0:
Adding fix from
-  https://github.com/llvm/llvm-project/issues/49036

This adds the `find_dynamic_unwind_sections` callback, used by libunwind to fix the` libc++abi: terminating due to uncaught exception of type int` on apple silicon. This returns fake_mach_header as the header for all JIT'd code.

The checks are performed during library loading and unloading time, and so have been added during interpreter creation and destruction called in `CppInterOp.cpp`, via the leaked `compat::Interpreter` class

### DIFF
--- a/lib/CppInterOp/Compatibility.h
+++ b/lib/CppInterOp/Compatibility.h
@@ -76,9 +76,11 @@ static inline char* GetEnv(const char* Var_Name) {
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/Twine.h"
+#include "llvm/BinaryFormat/MachO.h"
 #include "llvm/Config/llvm-config.h"
 #include "llvm/ExecutionEngine/JITSymbol.h"
 #include "llvm/ExecutionEngine/Orc/LLJIT.h"
+#include "llvm/Object/MachO.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/Path.h"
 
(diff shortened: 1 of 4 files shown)


vgvassilev:
@mcbarton, our cache sizes nowadays allow for having 1-2 osx-x86 builds. Are they easy to bring back?

mcbarton:
> @mcbarton, our cache sizes nowadays allow for having 1-2 osx-x86 builds. Are they easy to bring back?

@vgvassilev Yes, they are easy to bring back. I'll put in a PR which brings them back.

vgvassilev:
> > @mcbarton, our cache sizes nowadays allow for having 1-2 osx-x86 builds. Are they easy to bring back?
> 
> @vgvassilev Yes, they are easy to bring back. I'll put in a PR which brings them back.

Nice! Thanks for the speed-of-light reply and PR!

vgvassilev:
I think there was something that still needed to be done on the apple side. cc: @lhames.

lhames:
From memory this should be sufficient, but it's been a while since I looked at this. I'd check the error logs but it looks like they've expired. Is there a way to re-run the tests?

anutosh491:
> Is there a way to re-run the tests?

I just rebased the PR on latest main if that would help

anutosh491:
The tests on OSX fail with this https://github.com/compiler-research/CppInterOp/actions/runs/14876386113/job/41774415677?pr=254#step:9:457

PR #269, opened 2024-05-06
Update clang format and clang tidy

mcbarton:
Ths updates the workflows for clang format and clang tidy. This is update the actions to use new versions of llvm components, and to use new action versions due to warnings.


vgvassilev:
If we decide to go this way we should do it across all repositories, including clad :(

vgvassilev:
> @vgvassilev the upgrade to the action versions is definitely needed due to the node 16 to node 20 transistion. Without these change the workflow will eventually refuse to run. The change to using the latest llvm is not. I just decided to do that while updating the actions.

Ok, understood!

PR #270, opened 2024-05-06
Speed up build_cache process when not needing to build cache

mcbarton:
This should essentially allow workflow runs to skip over the build_cache matrix stage when the cache exists and doesn't need to be built.


mcbarton:
@vgvassilev This PR takes waiting for the build_cache stage running from 3 1/2 minute to 21 seconds, allowing PRs to run much faster.

vgvassilev:
Nice!

PR #276, opened 2024-05-09
Run wasm build unconditionally in the ci

mcbarton:
Fixes https://github.com/compiler-research/CppInterOp/issues/274
This PR will change it so the wasm builds in the ci occur regardless of the outcome of the non wasm jobs. The reason the dependence was there previously was due to build times being large, but this is no longer the case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1063,7 +1063,7 @@ jobs:
       timeout-minutes: 30
 
   emscripten_wasm:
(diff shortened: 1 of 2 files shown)


mcbarton:
@vgvassilev @maximusron can one of you merge this PR? It just changes the dependence of the matrix which the wasm build is dependent on, so it will run unconditionally. It also fixes the mamba envirnoment such that the was builds pass.

vgvassilev:
Thank you!

PR #277, opened 2024-05-10
Fix number of processors environment variable Windows

mcbarton:
The envirnoment variable for the number of processors on Windows was previous not being calculated correctly, or outputted to the Github envirnoment. This means non of the Windows builds previously were building in parallel. This PR should fix this.


vgvassilev (approved):
LGTM!

PR #280, opened 2024-05-14
Speed up build cache matrix

mcbarton:
There is no reason to checkout the CppInterOp repo in the build cache stage. Removing this should make this stage quicker.


mcbarton:
Closing as I forgot about the patches needed for the llvm in CppInterOp

PR #281, opened 2024-05-14
[tests] Conditionally skip valgrind xfails using llvm::sys

maximusron:


vgvassilev (approved):
Nice!

maximusron:
This doesn't seem to work yet, I think we should hold on merging it

vgvassilev:
Yeah, just noticed that...

maximusron:
strange, this was building locally with no problems

unittests/CppInterOp/Utils.h
```diff
@@ -6,6 +6,14 @@
...
 #include <vector>
 
+#ifndef __APPLE__
+  #include <valgrind/valgrind.h>
+#endif //__APPLE__
+
+#ifdef __APPLE__
+  #define RUNNING_ON_VALGRIND 0
+#endif //__APPLE__
+
```
vgvassilev, line 16:
```suggestion
#include "llvm/Support/Valgrind.h"
```

maximusron: this doesn't work as the preprocessor macro isn't visible anymore. Maybe switching to the original usage of `llvm::sys::RunningOnValgrind` will work, but it uses the same directive to determine if we are running on valgrind..

vgvassilev: Which preprocessor macro? How does this work for llvm then?

maximusron: it's something like:

```cpp
#if HAVE_VALGRIND_VALGRIND_H
#include <valgrind/valgrind.h>
 
bool llvm::sys::RunningOnValgrind() {
  return RUNNING_ON_VALGRIND;
}
 
#else  // !HAVE_VALGRIND_VALGRIND_H
 
bool llvm::sys::RunningOnValgrind() {
  return false;
}
```

Let me try again with the previous usage to see if it works now, with the apt install

vgvassilev: Unfortunately I do not think so. Do we need to rebuild the cached llvm or something?

maximusron:
```cpp
#if __APPLE__

bool TestUtils::RunningOnValgrind() {
  return false;
}

#else  // !__APPLE__
 
#include <valgrind/valgrind.h>
bool TestUtils::RunningOnValgrind() {
  return RUNNING_ON_VALGRIND;
}
 
#endif  // !__APPLE__

```
Would be identical to https://llvm.org/doxygen/Valgrind_8cpp_source.html

maximusron: It could be that we aren't setting `HAVE_VALGRIND_VALGRIND_H` in the llvm config

PR #282, opened 2024-05-14
Fix warning micromamba envirnoment setup ci

mcbarton:
This PR is part of the node 16 to node 20 transition. It should stop the warning appearing when the ci runs due to the provision-with-micromamba action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1069,64 +1069,76 @@ jobs:
             compiler: gcc-12
             clang-runtime: '18'
             cling: Off
+            micromamba_shell_init: bash
           - name: ubu22-x86-gcc12-clang-repl-17-emscripten_wasm
             os: ubuntu-22.04
             compiler: gcc-12
             clang-runtime: '17'
             cling: Off
+            micromamba_shell_init: bash
           - name: ubu22-x86-gcc12-clang-repl-16-emscripten_wasm
             os: ubuntu-22.04
             compiler: gcc-12
             clang-runtime: '16'
             cling: Off
+            micromamba_shell_init: bash
           - name: ubu22-x86-gcc9-clang13-cling-emscripten_wasm
             os: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


mcbarton:
@vgvassilev can this PR be merged? It is part of the node 16 to node 20 transition. After this is merged only the Windows errors will need to be dealth with when the ci runs to have no errors or warnings.

maximusron:
this looks good, but the codecov has dropped very drastically.. we should investigate this...

I tried looking into codecov reports with miromamba and couldn't find much. however. I did notice that both this and the other [PR](https://github.com/compiler-research/CppInterOp/pull/283) have dropped the codecoverage by the same number, see https://app.codecov.io/gh/compiler-research/CppInterOp/pulls

mcbarton:
> this looks good, but the codecov has dropped very drastically.. we should investigate this...
> 
> I tried looking into codecov reports with miromamba and couldn't find much. however. I did notice that both this and the other [PR](https://github.com/compiler-research/CppInterOp/pull/283) have dropped the codecoverage by the same number, see https://app.codecov.io/gh/compiler-research/CppInterOp/pulls

@maximusron The drop in codecov drop has nothing to do with this PR. It dropped when we stopped having coverage turned to true in all Ubuntu jobs. The coverage number on the main branch is not updating from the 88%, after that PR merge seems to be the issue.

mcbarton:
@vgvassilev The issue with the coverage number not updating on the main is due to the token issue just like in xeus-cpp. Can you fix it like you did xeus-cpp repo?
Error message can be seen on the nightly runs of the ci on main.
Error: Codecov token not found. Please provide Codecov token with -t flag.

mcbarton:
@maximusron can you merge this now that I have outlined the issue with the coverage in an issue? It has nothing to do with this PR, but a new condition on coverage report uploading and tokens.

maximusron (approved):

maximusron:
Node.js 16 actions

Issue #262, opened 2024-05-04
[ci] Add a debug version of llvm and a debug build of CppInterOp.


vgvassilev:
See https://github.com/vgvassilev/clad/pull/662

Issue #221, opened 2024-04-05
[CI] Minimal jobs to check API on pull requests


maximusron:
The current CI has an extensive suite of jobs covering every permutation of osx, windows, linux with both clang-repl and cling following every tool that uses CppInterOp. This A) takes up a lot of resources and B) are not really necessary to validate most pull requests changing the InterOp API.

I would propose segmenting a minimal workflow that have 3 sets of jobs each for windows, os x and ubuntu that run the InterOp build and test on the clang 16-18 and only run tools like the xeus products, cppyy and WASM once that is verified. Extensive build coverage can also be migrated to the repos of the tools themselves. Essentially it would be good if we don't tag every job with pull_request but have a set that would serve as a sanity check.

The other jobs that cover every permutation can be ran post merge to check main. This can speedup workflows especially when features are being developed in parallel

@vgvassilev @mcbarton would this be something we can do?

maximusron:
Not sure if this is the case but it seems like each job runner is building LLVM. I was wondering if we can also merge runners that use the same LLVM clang-repl/cling build to test multiple tools if that can make the jobs more efficient. That might make debugging failures harder, but is there any other way we can improve this config?

mcbarton:
Reducing the extensiveness of the CI shouldn't be done in my opinion. It may seem silly to have an extensive PR to test small changes, but small changes can have unexpected consequences, and the CI can help catch them before they end up in the main. 

The main reason I see for the PRs being slow when having multiple PRs is due to limitations of the caching system and number of Github runners which can be used simultaneously accross the organisation.

The llvm cache has to be created for each PR, and cannot be shared with the main branch or between PRs. This will overwhelm the cache storage limit even with the number of jobs reduced. An alternative to the cache of llvm builds may be to custom VMs to run the workflows on. See the beta feature here

https://github.com/github/roadmap/issues/826

I don't see a solution to the number of simultaneous runners, but the custom vm option once available should allow for a dramatic speed up in PR run time.

vgvassilev:
I believe the cache in main can be reused. I think for llvm17 we can just take the binary package as we do not have patches on top iirc. The same for llvm18...

mcbarton:
We apply patches for llvm 17 so I don't believe we can use the binary packages for this version. The binary package for llvm 18 requires no patches, so could be used once a web assembley build is available (don't believe this is currently available on emscripten forge though). 

[...]

PR #283, opened 2024-05-14
Speed up coverage job ci

mcbarton:
Given when coverage is turned on CppInterOp is built with debug options, when cppyy makes use of it its tests run much slower. This PR should fix this issue, by rebuilding CppInterOp as a Release version with no debug options once the coverage report has been uploaded. It takes extra time to build CppInterOp again, but this should be much less time then how much longer it takes to run the cppyy tests when CppInterOp is built with debug options.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -782,7 +782,7 @@ jobs:
       run: |
         sudo apt install lcov
         echo "CODE_COVERAGE=1" >> $GITHUB_ENV
-        echo "BUILD_TYPE=Debug" >> $GITHUB_ENV
+        echo "BUILD_TYPE=RelWithDebInfo" >> $GITHUB_ENV
 
     - name: Build and Test/Install CppInterOp on Unix systems
       if: ${{ runner.os != 'windows' }}


mcbarton:
This PR saves almost 10 minutes off the coverage job.

mcbarton:
> Do we have guidance from codecov if building in Release would produce useful information? We need to relate the coverage information against the source code.

@vgvassilev The release build is done after the code coverage report is uploaded based on the debug version. This change cannot effect the outcome of the coverage report.

mcbarton:
@vgvassilev can this PR be reviewed too?

maximusron:
Hi @mcbarton

Looks good to me! If I understand correctly this PR rebuilds CppInterOp in release mode after the codecov report.

mcbarton:
> Hi @mcbarton
> 
> Looks good to me! If I understand correctly this PR rebuilds CppInterOp in release mode after the codecov report.

That's exactly what its doing. Saves approximately 10 minutes of time for that job due to the speed up in the time to do the cppyy tests.

maximusron:
@mcbarton the environment variables and paths remain the same after the codecov job so this commit should perform the same minimally

maximusron:
@vgvassilev this works for me but the only con would be that when we ssh into the runners to debug, we will have to rebuild in debug mode. Seems like a fair enough trade-off for a faster CI

mcbarton:
> @vgvassilev this works for me but the only con would be that when we ssh into the runners to debug, we will have to rebuild in debug mode. Seems like a fair enough trade-off for a faster CI

@maximusron I have made this change. Are you happy with it?

maximusron:
@mcbarton I think the last commit is not required, since we would probably need to rebuild all the way to cppyy after this which is something we can do manually (these builds also add overheads)

.github/workflows/ci.yml
```diff
@@ -1053,6 +1087,40 @@ jobs:
...
 
+    - name: Rebuild CppInterOp in Debug mode for tmate session
+      if: ${{ runner.os != 'windows' && (matrix.coverage == true) &&  failure() }}
+      run: |
+        cling_on=$(echo "${{ matrix.cling }}" | tr '[:lower:]' '[:upper:]')
+        
+        # Build CppInterOp next to cling and llvm-project.
+        cd build
+        rm -rf ./*
+        cling_on=$(echo "${{ matrix.cling }}" | tr '[:lower:]' '[:upper:]')
```
maximusron, line 1098: we can drop this rebuild

mcbarton: You mentioned that if you have to ssh into the machine it will need to be rebuilt with debug. This saves time by getting the ci to rebuild, and also guards again the case the person sshing into the macine forgets to (we know too, but others may not).

mcbarton:
> @mcbarton I think the last commit is not required, since we would probably need to rebuild all the way to cppyy after this which is something we can do manually (these builds also add overheads)

I'd argue the build is not an overhead since its only rebuilt if the job fails.

.github/workflows/ci.yml
mcbarton: @maximusron I can revert this change if your sure but I think its a good guard against those sshing in..

maximusron:
I realise that the valgrind commands on cppyy and the unittests need CppInterOp to remain in debug mode.

Otherwise don't get any function names or line number information. Using Valgrind on release build finds fewer or no memory issues since llvm has memory pool management of its own. In release mode valgrind will not be able to see the alloc/dealloc mechanism.

mcbarton:
> I realise that the valgrind commands on cppyy and the unittests need CppInterOp to remain in debug mode.
> 
> Otherwise don't get any function names or line number information. Using Valgrind on release build finds fewer or no memory issues since llvm has memory pool management of its own. In release mode valgrind will not be able to see the alloc/dealloc mechanism.


> I realise that the valgrind commands on cppyy and the unittests need CppInterOp to remain in debug mode.
> 
> Otherwise don't get any function names or line number information. Using Valgrind on release build finds fewer or no memory issues since llvm has memory pool management of its own. In release mode valgrind will not be able to see the alloc/dealloc mechanism.

@maximusron Wouldn't this be true of the for llvm versions too since they perform the Valgrind commands too? If yes, do we need a debug version of all the Ubuntu jobs?

.github/workflows/ci.yml
mcbarton: > We can make it conditional on if the runner is run in debug.

@vgvassilev Currently we have no debug flag. I can make it dependent on a debug run when I address https://github.com/compiler-research/CppInterOp/issues/262 soon if that's ok?

mcbarton:
> > @maximusron Wouldn't this be true of the for llvm versions too since they perform the Valgrind commands too? If yes, do we need a debug version of all the Ubuntu jobs? Once this comment is answered i'll close this PR with a link to your comment.
> 
> I think a debug build works but at the same time we seem to be departing from having the CI itself a debug mode testing platform, I do think having the CI run everything in debug mode (for unique build types since we have a multitude of compiler, OS and CPU platform specific errors). This way we catch more stuff and have good logs as a first response. I don't think the need for the builds to be blazing fast warrants this change
> 
> Also the builds have been fast and easy to work since your recent upgrades :)
> 
> @vgvassilev what do you think?

@maximusron @vgvassilev I will address the debug run issue by tackling issue https://github.com/compiler-research/CppInterOp/issues/262 . I will make a PR which adds a debug flag, which does different things if we feels its needed, so its not done all the time.

vgvassilev:
I think `RelWithDebInfo` should be fine for the CI. That essentially turns on assertions and it is still fast. Maybe that's a good common ground. The downside is that if something fails on a particular bot we have to rebuild everything in debug mode but that happens rarely. I guess we can revisit if necessary...

mcbarton:
> I think `RelWithDebInfo` should be fine for the CI. That essentially turns on assertions and it is still fast. Maybe that's a good common ground. The downside is that if something fails on a particular bot we have to rebuild everything in debug mode but that happens rarely. I guess we can revisit if necessary...

@vgvassilev @maximusron This doesn't work, as turning on codecov turns on debug mode, regardless of what we put for the build type (Errors here https://github.com/compiler-research/CppInterOp/blob/791995a7b518226bbf9003ad0e13f35ec39e4705/CMakeLists.txt#L282C1-L282C63). Do you know if codecov works with RelWithDebInfo? If not, I might close this PR and revisit the issue at some point in the future.

maximusron:
> This doesn't work, as turning on codecov turns on debug mode, regardless of what we put for the build type

ah that is a problem... https://github.com/bilke/cmake-modules/blob/1fcf7f4a2179a7b649be15473906882871ef5f0b/CodeCoverage.cmake#L213-L214 this does seem to be the case. We need to have a non optimised build to ensure the best results from codecov

Overall, with valgrind and codecov things point towards us retaining a debug build on the CI. I don't think this is much of a priority given that the builds are much faster and easy to work with now. Maybe we can drop this PR

mcbarton:
> > This doesn't work, as turning on codecov turns on debug mode, regardless of what we put for the build type
> 
> ah that is a problem... https://github.com/bilke/cmake-modules/blob/1fcf7f4a2179a7b649be15473906882871ef5f0b/CodeCoverage.cmake#L213-L214 this does seem to be the case. We need to have a non optimised build to ensure the best results from codecov
> 
> Overall, with valgrind and codecov things point towards us retaining a debug build on the CI. I don't think this is much of a priority given that the builds are much faster and easy to work with now. Maybe we can drop this PR

I'll drop this PR, and only revisit the issue if we feel there is a need to in future. @maximusron can you close this issue https://github.com/compiler-research/CppInterOp/issues/221 , as I feel its been resolved now?

Issue #284, opened 2024-05-15
Code coverage not updating on main


mcbarton:
The code coverage percentage on the main is not updating due to a missing. The same issue outlined here for xeus-cpp https://github.com/compiler-research/xeus-cpp/issues/68.

PR #285, opened 2024-05-15
[ci] Add token to upload codecov reports

maximusron:


maximusron:
Fixes : #284

vgvassilev:
We need to update the commit message, too.

vgvassilev:
This fails because a new valgrind gem has showed up...

maximusron:
> This fails because a new valgrind gem has showed up...

Looks like we see this on all current PR's, with clang 17 and 18. I am rerunning a job on another PR to confirm. New detected uninitialised values in `ScopeReflectionTest_IncludeVector` orginating from `llvm::APInt::setBits`. Could this be because of some new patch in clang-repl?

vgvassilev:
We should see if we can reproduce it in a debug build.

maximusron:
@vgvassilev Looks like it works now, this failure should disappear on all PR's by rebasing

maximusron:
@vgvassilev This is ready to merge

Issue #260, opened 2024-05-03
Fix all bot warnings and enable -Werror


vgvassilev:
See https://github.com/vgvassilev/clad/pull/756 for inspiration.

PR #288, opened 2024-05-15
Treat all warnings as errors

mcbarton:
This PR will add the flag to treat all warnings as errors when building the ci.
Fixes https://github.com/compiler-research/CppInterOp/issues/260
Fixes https://github.com/compiler-research/CppInterOp/issues/234

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -819,6 +819,7 @@ jobs:
                 -DBUILD_SHARED_LIBS=ON                          \
                 -DCODE_COVERAGE=${{ env.CODE_COVERAGE }}        \
                 -DCMAKE_INSTALL_PREFIX=$CPPINTEROP_DIR          \
+                -DLLVM_ENABLE_WERROR=On                         \
                 ../
         else
           cmake -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}    \
@@ -829,6 +830,7 @@ jobs:
                 -DBUILD_SHARED_LIBS=ON                      \
                 -DCODE_COVERAGE=${{ env.CODE_COVERAGE }}    \
                 -DCMAKE_INSTALL_PREFIX=$CPPINTEROP_DIR      \
+                -DLLVM_ENABLE_WERROR=On                     \
                 ../
         fi
         os="${{ matrix.os }}"
@@ -844,7 +846,7 @@ jobs:
         echo "CPPINTEROP_DIR=$CPPINTEROP_DIR" >> $GITHUB_ENV
         echo "LLVM_BUILD_DIR=$LLVM_BUILD_DIR" >> $GITHUB_ENV
         echo "CPLUS_INCLUDE_PATH=$CPLUS_INCLUDE_PATH" >> $GITHUB_ENV
-
+         
     - name: Build and Test/Install CppInterOp on Windows systems
       continue-on-error: true
       if: ${{ runner.os == 'windows' }}
@@ -903,6 +905,7 @@ jobs:
                 -DUSE_REPL=OFF                                  `
                 -DCling_DIR="$env:LLVM_BUILD_DIR\tools\cling"   `
                 -DLLVM_DIR="$env:LLVM_BUILD_DIR" `
+                -DLLVM_ENABLE_WERROR=On                          `
                 -DClang_DIR="$env:LLVM_BUILD_DIR"  -DCODE_COVERAGE=${{ env.CODE_COVERAGE }} -DCMAKE_INSTALL_PREFIX="$env:CPPINTEROP_DIR"  ..\
         }
         else
@@ -911,6 +914,7 @@ jobs:
                 -DUSE_CLING=OFF                             `
                 -DUSE_REPL=ON                               `
                 -DLLVM_DIR="$env:LLVM_BUILD_DIR\lib\cmake\llvm"  `
+                -DLLVM_ENABLE_WERROR=On                          `
                 -DClang_DIR="$env:LLVM_BUILD_DIR\lib\cmake\clang"  -DCODE_COVERAGE=${{ env.CODE_COVERAGE }}  -DCMAKE_INSTALL_PREFIX="$env:CPPINTEROP_DIR"  ..\
         }
         cmake --build . --config ${{ env.BUILD_TYPE }} --target check-cppinterop --parallel ${{ env.ncpus }}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -211,7 +211,7 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
 
   endif()
 
-  ##Replace \ with / in LLVM_DIR (attempt to fix path parsing issue Windows)
+  #Replace \ with / in LLVM_DIR (attempt to fix path parsing issue Windows)
   string(REPLACE "\\" "/" LLVM_DIR "${LLVM_DIR}")
 
   # When in debug mode the llvm package thinks it is built with -frtti.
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -815,6 +815,7 @@ jobs:
                 -DBUILD_SHARED_LIBS=ON                          \
                 -DCODE_COVERAGE=${{ env.CODE_COVERAGE }}        \
                 -DCMAKE_INSTALL_PREFIX=$CPPINTEROP_DIR          \
+                -DCMAKE_COMPILE_WARNING_AS_ERROR=ON            \
```
vgvassilev, line 818: I believe having `-DLLVM_ENABLE_WERROR=On` will set up everything that we need in cmake.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.24)
```
vgvassilev, line 1: Why do we need to upgrade here?

mcbarton: CMAKE_COMPILE_WARNING_AS_ERROR was new in 3.24 . I don't think its really necessary, but it's what I was asked to do when I did this for xeus-cpp.

.github/workflows/ci.yml
mcbarton: I will switch the flag and revert the cmake minimum version change.

CMakeLists.txt
vgvassilev: I do not think we should upgrade cmake to that late versions. That'd leave behind a lot of clients.

vgvassilev: It is probably fine as xeus-cpp is a end-point project not a library that needs to be incorporated in many contexts...

mcbarton:
@maximusron This PR involves more than just ci changes. Can you make more labels such as cmake and codebase?

unittests/CppInterOp/VariableReflectionTest.cpp
```diff
@@ -121,13 +121,17 @@ TEST(VariableReflectionTest, GetVariableOffset) {
...
-          ((unsigned long) &(c.b)) - ((unsigned long) &(c.a)));
+          (reinterpret_cast<unsigned long>(&(c.b)) - reinterpret_cast<unsigned long>(&(c.a))));
   EXPECT_EQ(Cpp::GetVariableOffset(datamembers[2]),
-          ((unsigned long) &(c.c)) - ((unsigned long) &(c.a)));
+          (reinterpret_cast<unsigned long>(&(c.c)) - reinterpret_cast<unsigned long>(&(c.a))));
   EXPECT_EQ(Cpp::GetVariableOffset(datamembers[3]),
-          ((unsigned long) &(c.d)) - ((unsigned long) &(c.a)));
-
+          (reinterpret_cast<unsigned long>(&(c.d)) - reinterpret_cast<unsigned long>(&(c.a))));
+#endif // WIN32
```
vgvassilev, line 134: These do not seem right.

mcbarton: This is something I'm planning to revert. It was from when I was trying to debug the Windows compile issue.

mcbarton: I tried many times to geet this part of the test to work on Windows without sucess.

mcbarton: I've reverted this change now.

mcbarton:
@vgvassilev I'm still resolving issues with osx, but can you review what I have done so far?

unittests/CppInterOp/VariableReflectionTest.cpp
```diff
@@ -91,6 +91,21 @@ TEST(VariableReflectionTest, GetVariableType) {
...
+  class C {                                                     \
+  public:                                                       \
+    int a;                                                      \
+    double b;                                                   \
+    int *c;                                                     \
+    int d;                                                      \
+    static int s_a;                                             \
+  } c;                                                            \
+  int C::s_a = 12;
+#else
```
vgvassilev, line 108: Likewise.

unittests/CppInterOp/VariableReflectionTest.cpp
```diff
@@ -121,13 +137,17 @@ TEST(VariableReflectionTest, GetVariableOffset) {
   EXPECT_TRUE((bool) Cpp::GetVariableOffset(Decls[1])); // N
 
   EXPECT_EQ(Cpp::GetVariableOffset(datamembers[0]), 0);
+//FIXME: Unable to compile these on Windows due to type conversion
+#ifdef _WIN32
```
vgvassilev, line 141: What is the error?

unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -671,6 +682,7 @@ TEST(ScopeReflectionTest, GetBaseClassOffset) {
   class F : public A { int m_f; };                              \
   class G : public F { int m_g; };
 
+#endif
   CODE;
```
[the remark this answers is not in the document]
mcbarton: If you can make a suggestion I'll try it. I tried many ways to get something to pass this test, and duplication was the only way I could get it to work.

unittests/CppInterOp/VariableReflectionTest.cpp
mcbarton: I'd have to rerun it again without this guard, but I think it was something like  &(c.b),&(c.c) and &(c.d) was coming out as int** type and was unable to convert to unsigned long on Windows. &(c.a) was coming out int* and it was unable to convert to unsigned long.

unittests/CppInterOp/ScopeReflectionTest.cpp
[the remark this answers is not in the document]
vgvassilev: https://godbolt.org/z/Wb4cvo58P ?

unittests/CppInterOp/VariableReflectionTest.cpp
vgvassilev: What if we cast to `intptr_t`?

mcbarton: I will try this in a few minutes. I have performed a run with your suggestions, and removed this guard so you can see the exact error incase I am misquoting the error.

mcbarton: @vgvassilev Here is a link to a run with the error https://github.com/compiler-research/CppInterOp/actions/runs/9142123700/job/25137329640?pr=288#step:16:177 . I will now try your suggestion.

unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -671,7 +671,10 @@ TEST(ScopeReflectionTest, GetBaseClassOffset) {
   class F : public A { int m_f; };                              \
   class G : public F { int m_g; };
 
-  CODE;
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-private-field"
+CODE;
+#pragma GCC diagnostic pop
```
vgvassilev, line 677: Let's drop the pragmas and make classes `struct`s. This way the `m_*` will become public and will make the warning go away.

mcbarton: That will fix most warnings, but this won't fix the warning about   const int N = 5; being unused VariableReflectionTest.cpp.

unittests/CppInterOp/VariableReflectionTest.cpp
mcbarton: > What if we cast to `intptr_t`?

This solution worked.

unittests/CppInterOp/ScopeReflectionTest.cpp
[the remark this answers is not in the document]
mcbarton: This has been solved by changing the classes to structs as suggsted.

unittests/CppInterOp/ScopeReflectionTest.cpp
mcbarton: https://github.com/compiler-research/CppInterOp/actions/runs/9142202412/job/25137488586?pr=288#step:15:185 Link to error with const int N = 5; Unless there is a specific reason its being labelled const then I can try fixing it by removing the const?

mcbarton: Testing removing const to fix. Will revert and try something else if needed

mcbarton:
@vgvassilev This PR is ready for further review.

unittests/CppInterOp/VariableReflectionTest.cpp
```diff
@@ -93,7 +93,7 @@ TEST(VariableReflectionTest, GetVariableType) {
 
 #define CODE                                                    \
   int a;                                                        \
-  const int N = 5;                                              \
```
vgvassilev, line 96: We can’t change that because this changes the intent of the test. Maybe use it in the struct definition below to initialize some member.

mcbarton:
@vgvassilev I have tried to address your comment about retaining N as const. This PR is ready for further review.

vgvassilev (approved):
LGTM!

PR #290, opened 2024-05-18
[ci] Update InterOp builds to Ubuntu 24.04

### LINKED ISSUE #265: Ubuntu 24.04 support ci
Currently the Ubuntu covered in the ci is Ubuntu 22.04 . Recently a newer version came out Ubuntu 24.04 . Runners with this os are on the public roadmap https://github.com/orgs/github/projects/4247?pane=issue&itemId=60910253 . Once they are available we can either switch all Ubuntu jobs to 24.04 , or run them alongside the 22.04 .

aaronj0:
Meant to fix #265 

The CI seems to be picking up `cppyy-cling` from master when the backend install fails and uses an include `RTypes.h` used for ROOT. Specifically `strlcpy.h` which was copied to fix issues with Clang parsing gcc headers. These are BSD only string functions and not part of any standard, and declared in `cling/src/core/clib`:

```cpp
#ifndef ROOT_strlcpy
#define ROOT_strlcpy

#include <ROOT/RConfig.hxx>
#ifdef __cplusplus
extern "C" {
#endif

size_t strlcpy(char *dst, const char *src, size_t siz);
size_t strlcat(char *dst, const char *src, size_t siz);

#ifdef __cplusplus
}
```

Ideally we should update the header guards for ubuntu 24 but in our case we don't use `cppyy-cling`, nor does our fork have those files. pip defaults to downloading and using the tarball for upstream cppyy-backend which causes this mess. Defaulting to cmake could fix this


maximusron:
this is picking up `cppyy-cling` from master when the backend install fails and uses an include `RTypes.h` which was meant to provide ROOT specific implementations of system libraries used by `string.h`. Specifically `strlcpy.h` which was copied to fix issues with Clang parsing gcc headers. Ideally we should update the header guards for ubuntu 24 upstream but in our case we don't use `cppyy-cling`, nor does our fork have those files. pip defaults to downloading and using the tarball for upstream cppyy-backend which causes this mess. Defaulting to make could fix this

mcbarton:
@maximusron If this works you'll want to wipe the cache the cache this PR made before merging, and delete the Ubuntu cache on the main too.

mcbarton:
@maximusron can you link this PR to https://github.com/compiler-research/CppInterOp/issues/265 ?

maximusron:
> @maximusron can you link this PR to #265 ?

Yeah sure, this PR was meant to just fix the cppyy-backend issue but if this works then maybe we can just upgrade the ubuntu versions with this. Ideally we shouldn't have to change anything on the cppyy-backend side but I am still investigating that

mcbarton:
@maximusron Is the commit where you removed -DLLVM_ENABLE_WERROR=On  temporary while you debug this PR?

maximusron:
@mcbarton yep its temporary, but we have another set of warnings that emerge on ubuntu 24 that we need to fix. This will happen everytime we change the os/platform/compiler so ideally I would suggest dropping `-DLLVM_ENABLE_WERROR` for good

vgvassilev:
> @mcbarton yep its temporary, but we have another set of warnings that emerge on ubuntu 24 that we need to fix. On this PR, I removed it since we would error out on the build CppInterOp step. This may happen everytime we change the os/platform/gcc/clang

Yes, and as part of this change we will fix them. Otherwise people commit stuff with warnings which are actually real issues...

maximusron:
@vgvassilev yes I have only removed it fix the pypi issue, theres something in the pyproject.toll file that needs to be changed for no-pep to work. Once that worked I thought we can fix the new warnings before merging

.github/workflows/ci.yml
```diff
@@ -985,7 +984,7 @@ jobs:
         # Install cppyy
         git clone --depth=1 https://github.com/compiler-research/cppyy.git
         cd cppyy
-        python -m pip install --upgrade . --no-deps
+        python -m pip install --upgrade . --no-deps --no-use-pep517
```
henryiii, line 987:
```suggestion
        python -m pip install --upgrade . --no-deps --no-build-isolation
```

mcbarton:
@maximusron The ci won't run on this PR until you have resolve the merge conflicts

mcbarton:
Closing as this has been done now for CppInterOp.

PR #291, opened 2024-05-18
Fix bug due to swiftlint osx jobs ci

mcbarton:
@vgvassilev @maximusron can one of you merge this into the main to fix the issue with osx x86 jobs? I know this fix works as I'm currently using it in my PR on treating all warnings as errors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -758,6 +758,10 @@ jobs:
       run: |
         brew update
         brew remove ruby@3.0
+        export ARCHITECHURE=$(uname -m)
+        if [[ "$ARCHITECHURE" == "x86_64" ]]; then
+          brew remove swiftlint
+        fi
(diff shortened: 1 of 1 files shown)


maximusron:
@mcbarton will do once the checks are done

maximusron:
@mcbarton we need the cache from PR #281 since LLVM has to be built after valgrind. I will rerun these jobs to avoid both of them building a cache at the same time

mcbarton:
@maximusron The cache from both PRs can't clash since there isn't a Ubuntu cache on the main anymore. One PR cannot pickup the cache from another PR. This PR also has nothing to do with caching. It is just removing a package which causes issues on osx. Unsure why the ci run had to be cancelled.

maximusron:
That's why I would prefer that this PR doesn't also build a cache at the same time. Once the new cache is built, this PR can just use that and make things more simple. We have no rush to merge this.

maximusron (approved):

PR #297, opened 2024-05-20
Fix unit test failure

maximusron:


maximusron:
Merging, as this fixes the non wasm builds on the CI

PR #304, opened 2024-06-04
Add a libclang-style C API and improve type safety

Gnimuc:
TODO:

- [x] `void AddSearchPath(const char *dir, bool isUser, bool prepend);`
- [x] `const char* GetResourceDir();`
- [x] `void AddIncludePath(const char *dir);`
- [x] `int Declare(const char* code, bool silent);`
- [x] `int Process(const char *code);`
- [x] `intptr_t Evaluate(const char *code, bool *HadError/*=nullptr*/);`
- [x] `std::string LookupLibrary(const char* lib_name);`
- [x] `bool LoadLibrary(const char* lib_stem, bool lookup);`
- [x] `void UnloadLibrary(const char* lib_stem);`
- [x] `std::string SearchLibrariesForSymbol(const char* mangled_name, bool search_system /*true*/);`
- [x] `std::string ObjToString(const char *type, void *obj);`
- [x] `void Destruct(TCppObject_t This, TCppScope_t scope, bool withFree /*=true*/);`
- [x] `TCppFuncAddr_t GetFunctionAddress(const char* mangled_name);`
- [x] `CPPINTEROP_API JitCall MakeFunctionCallable(TCppConstFunction_t func);`
- [x] `intptr_t GetVariableOffset(TCppScope_t var); `
- [x] `bool InsertOrReplaceJitSymbol(const char* linker_mangled_name, uint64_t address);`
- [x] Add C APIs for scope, type, object, and jitcall manipulation
- [ ] Add tests and run clang-format and fix clang-tidy

Update:

Keeping the C++ API untouched is impossible if we want libclang-styled types. As a result, I reimplemented most of the C++ APIs on the C API side.

### DIFF
--- a/include/clang-c/CXCppInterOp.h
+++ b/include/clang-c/CXCppInterOp.h
@@ -59,7 +59,7 @@ TInterp_t clang_interpreter_takeInterpreterAsPtr(CXInterpreter I);
 /**
  * Undo N previous incremental inputs.
  */
-CXErrorCode clang_interpreter_Undo(CXInterpreter I, unsigned int N);
+enum CXErrorCode clang_interpreter_Undo(CXInterpreter I, unsigned int N);
 
 /**
  * Dispose of the given interpreter context.
--- a/include/clang-c/CXCppInterOp.h
+++ b/include/clang-c/CXCppInterOp.h
@@ -5,8 +5,8 @@
 #include "clang-c/CXErrorCode.h"
 #include "clang-c/CXString.h"
 #include "clang-c/ExternC.h"
+#include "clang-c/Index.h"
 #include "clang-c/Platform.h"
-#include "llvm-c/LLJIT.h"
 
 #include <stdbool.h>
 #include <stddef.h>
@@ -256,19 +256,10 @@ clang_interpreter_getFunctionAddressFromMangledName(CXInterpreter I,
  */
 
 /**
- * Describes the kind of entity that a type refers to.
- */
-enum CXQualTypeKind {
-  CXQualType_Unexposed = 0,
-  CXQualType_Invalid = 1,
-  // reserved for future use
-};
-
-/**
  * An opaque pointer representing a type.
  */
 typedef struct {
-  enum CXQualTypeKind kind;
+  enum CXTypeKind kind;
   void* data;
   const void* meta;
 } CXQualType;
--- a/include/clang-c/CXCppInterOp.h
+++ b/include/clang-c/CXCppInterOp.h
@@ -1,0 +1,935 @@
+// NOLINTBEGIN()
+#ifndef LLVM_CLANG_C_CXCPPINTEROP_H
+#define LLVM_CLANG_C_CXCPPINTEROP_H
+
+#include "clang-c/CXErrorCode.h"
+#include "clang-c/CXString.h"
+#include "clang-c/ExternC.h"
+#include "clang-c/Platform.h"
+#include "llvm-c/LLJIT.h"
+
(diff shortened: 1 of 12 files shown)


lib/Interpreter/CppInterOp.cpp
Gnimuc, line 1: I must have done something when configuring my local clang-format.

lib/Interpreter/CXCppInterOp.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+void clang_interpreter_unloadLibrary(CXInterpreter I, const char* lib_stem) {
+  static_cast<Cpp::Interpreter*>(I)->getDynamicLibraryManager()->unloadLibrary(
+      lib_stem);
+}
+
+CXString clang_interpreter_searchLibrariesForSymbol(CXInterpreter I,
+                                                    const char* mangled_name,
+                                                    bool search_system) {
+  return createCXString(
```
Gnimuc, line 148: Is it OK to keep a `StringRef` in the dynamic library manager to store the output? For the C++ interface, returning a `std::string` is fine,  but it's not C-friendly.

Gnimuc:
How to tell clang-tidy to ignore the `CXCppInterOp.h`? It's a C header and should not use any C++ stuff.

lib/Interpreter/CppInterOp.cpp
vgvassilev: You can do only diff-based formatting with `git-clang-format HEAD~`.

vgvassilev:
I feel like this PR is also addressing https://github.com/compiler-research/CPyCppyy/issues/8 to some extent. Maybe we can use the momentum gained here..

include/clang-c/CXCppInterOp.h
```diff
@@ -0,0 +1,643 @@
...
+/**
+ * \defgroup CPPINTEROP_TYPE_MANIP Type manipulations
+ *
+ * @{
+ */
+
+/**
+ * Describes the kind of entity that a type refers to.
+ */
+enum CXQualTypeKind {
```
vgvassilev, line 224: I am wondering if we should not use `clang/include/clang-c/Index.h` for these?

Gnimuc: Yes. I think that's possible.

vgvassilev: ping.

vgvassilev (commented):
This starts looking good. I wonder what's the way to test the C layer? I worry that it can regress without tests...

vgvassilev:
@Gnimuc, is that PR ready to review/merge?

Gnimuc:
> @Gnimuc, is that PR ready to review/merge?

I still need to add more tests. but, I feel that not all C++ APIs need to be exposed. I guess we can keep this in a separate branch a bit longer.

vgvassilev:
> > @Gnimuc, is that PR ready to review/merge?
> 
> I still need to add more tests. but, I feel that not all C++ APIs need to be exposed. I guess we can keep this in a separate branch a bit longer.

Okay, then I am waiting for a signal from you what and when to review.

Gnimuc:
> Hey, @Gnimuc, I was wondering what's the fate of this PR? Looks like we are almost there.

I will revisit this and rebase it on the latest master next month. (I'm still testing the library in https://github.com/Gnimuc/CppCall.jl)

Gnimuc:
I’m closing this because I feel that, instead of wrapping everything into one large PR, it’s better to proceed gradually.  see #337

PR #307, opened 2024-06-10
Fix wasm jobs in ci due to changes in xeus-cpp requirements

mcbarton:
This PR should allow the wasm jobs in the ci to pass again after the changes to the requirements in xeus-cpp


vgvassilev (approved):
LGTM!

PR #308, opened 2024-07-10
Library autoloader

alexander-penev:


include/clang/Interpreter/CppInterOp.h
```diff
@@ -663,6 +663,14 @@ namespace Cpp {
                                    unsigned complete_line = 1U,
                                    unsigned complete_column = 1U);
 
+  /// Set libraries autoload.
+  ///\param[in] autoload - true = libraries autoload is on.
```
vgvassilev, line 667:
```suggestion
  ///\param[in] autoload - true to enable library autoload.
```

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1071,9 +1077,9 @@ namespace Cpp {
     auto FDAorErr = compat::getSymbolAddress(I, mangled_name);
     if (llvm::Error Err = FDAorErr.takeError())
       llvm::consumeError(std::move(Err)); // nullptr if missing
-    else
+    else {
       return llvm::jitTargetAddressToPointer<void*>(*FDAorErr);
-
+    }
```
vgvassilev, line 1082: We do not use braces for single statement if-stmts.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3267,4 +3274,181 @@ namespace Cpp {
                          complete_column);
   }
 
+#define DEBUG_TYPE "autoload"
+
+  static inline std::string DemangleNameForDlsym(const std::string& name) {
+    std::string nameForDlsym = name;
+
+#if defined(R__MACOSX) || defined(R__WIN32)
```
vgvassilev, line 3282: Generally we avoid that to enable cross-compilation. We can ask the target triple for the same information at runtime. Should be something like `Interpreter::getExecutionEngine::getTargetTriple::getOS`.

lib/Interpreter/DynamicLibraryManagerSymbol.cpp
```diff
@@ -1153,8 +1153,8 @@ namespace Cpp {
   std::string Dyld::searchLibrariesForSymbol(StringRef mangledName,
                                              bool searchSystem/* = true*/) {
 #define DEBUG_TYPE "Dyld:searchLibrariesForSymbol:"
-    assert(!llvm::sys::DynamicLibrary::SearchForAddressOfSymbol(mangledName.str()) &&
-           "Library already loaded, please use dlsym!");
+//    assert(!llvm::sys::DynamicLibrary::SearchForAddressOfSymbol(mangledName.str()) &&
+//           "Library already loaded, please use dlsym!");
```
vgvassilev, line 1157: Why do we need to comment out this assert?

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -30,9 +30,24 @@ target_link_libraries(DynamicLibraryManagerTests
   PRIVATE
   clangCppInterOp
   )
+set_source_files_properties(DynamicLibraryManagerTest.cpp PROPERTIES COMPILE_DEFINITIONS
+  "LLVM_BINARY_DIR=\"${LLVM_BINARY_DIR}\""
+  )
+set_source_files_properties(DynamicLibraryManagerTest.cpp PROPERTIES COMPILE_DEFINITIONS
+  "CPPINTEROP_VERSION=\"${CPPINTEROP_VERSION}\""
+  )
```
vgvassilev, line 38: Do we use LLVM_BINARY_DIR or CPPINTEROP_VERSION in DynamicLibraryManagerTest.cpp?

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
```diff
@@ -2,55 +2,216 @@
...
+static inline std::string MangleNameForDlsym(const std::string& name) {
+  std::string nameForDlsym = name;
+#if defined(R__MACOSX) || defined(R__WIN32)
+  if (nameForDlsym[0] != '_')
+    nameForDlsym.insert (0, 1, '_');
+#endif //R__MACOSX
+  return nameForDlsym;
+}
+
+#include "../../lib/Interpreter/CppInterOp.cpp"
```
vgvassilev, line 36: We probably should drop this.

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
```diff
@@ -2,55 +2,216 @@
...
+  // Cover
+  // if (Path.empty()) {
+  //   LLVM_DEBUG(dbgs() << "DynamicLibraryManager::lookupLibMaybeAddExt(): "
+  // TODO: implement test this case
+
+  // Cover
+  // platform::DLClose(dyLibHandle, &errMsg);
+  // if (!errMsg.empty()) {
+  //   LLVM_DEBUG(dbgs() << "DynamicLibraryManager::unloadLibrary(): "
+  // TODO: implement test this case
```
vgvassilev, line 216: We should write these tests, right?

unittests/CppInterOp/TestSharedLib1/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# TODO: Remove when library unload work correctly
```
vgvassilev, line 1: I do not understand why the comment. Does it mean to remove TestSharedLib1 altogether?

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
alexander-penev: These are quite unexpected cases, even some of them are quite difficult to simulate even if you want to, but it's really good to implement these cases for better test coverage.

unittests/CppInterOp/TestSharedLib1/CMakeLists.txt
alexander-penev: Unloading libraries can hardly be implemented the way we think, so this functionality should probably be removed for now. If the unloading does work deterministically then more tests can be implemented.

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
vgvassilev: Yes, if we are not planning to implement these tests as part of this PR we should probably remove the code and keep track of this in a github issue.

unittests/CppInterOp/TestSharedLib1/CMakeLists.txt
vgvassilev: This comment is unclear what needs to be removed. Maybe remove the comment and open a github issue to track this?

unittests/CppInterOp/CMakeLists.txt
alexander-penev: These symbols are necessary so that we can include CppInterOp.cpp (to compile properly) and use from it functions that are not visible through the .h file ie. should be available only for the tests.

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
alexander-penev: We use from it functions that are not visible through the .h file ie. should be available only for the tests.

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
alexander-penev: They are planned unless we rush with the merge...

vgvassilev: Well, there is no rush but this is a blocker for xeus-cpp. If we can implement them within a week or so then that's probably good. Otherwise we can make a subsequent PR.

unittests/CppInterOp/DynamicLibraryManagerTest.cpp
vgvassilev: Okay, fair enough...

lhames:
A couple of quick thoughts for future upstreaming: we'd want to make this compatible with out-of-process execution, and it'd be great if we could re-use parts of it for MachO auto-linking.

We should consider whether `SimpleExecutorDylibManager` can be extended to handle all this.

vgvassilev:
> A couple of quick thoughts for future upstreaming: we'd want to make this compatible with out-of-process execution, and it'd be great if we could re-use parts of it for MachO auto-linking.

Thanks for the feedback, @lhames. That makes sense.

> We should consider whether `SimpleExecutorDylibManager` can be extended to handle all this.

I was not aware such a thing existed. @SahilPatidar perhaps that's interesting in terms of your project, too.

vgvassilev:
@aaronj0, do you have a clue why cppyy fails?

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3356,4 +3362,188 @@ namespace Cpp {
...
+        if (!failedSymbols.empty()) {
+          auto failingMR = R->delegate(failedSymbols);
+          if (failingMR) {
+            (*failingMR)->failMaterialization();
+          }
+        }
+
+        if (!loadedSymbols.empty()) {
+          llvm::cantFail(R->notifyResolved(loadedSymbols));
+          llvm::cantFail(R->notifyEmitted());
```
aaronj0, line 3470: [](url)

aaronj0:
> @aaronj0, do you have a clue why cppyy fails?

cppyy does not fail, InterOp does not build on any of the jobs. This comes back to the point that CppInterOp is only built on the jobs tagged with "cppyy" not the first stage that only builds llvm.

On a preliminary look, it seems like on llvm19 we get:
```
error: no matching function for call to ‘llvm::orc::MaterializationResponsibility::notifyEmitted()’
 3470 |           llvm::cantFail(R->notifyEmitted());
 ```
and on 18 and 17, these "undefined reference to `typeinfo` for `llvm::orc::DefinitionGenerator`"

anutosh491:
Hey @alexander-penev 

What is the progress on this ?

I haven't had the time to looking into the library loading feature yet but we would need this for xeus-cpp-lite and xeus-cpp I suppose. My understanding was that this would supersede the Loading shared objects feature we could achieve with cling's pragma.

PR #309, opened 2024-07-12
Update template function selection to use min required args

maximusron:


vgvassilev (approved):
LGTM, can you address clang-format?

vgvassilev:
@maximusron, can we add a test for this change?

PR #316, opened 2024-08-14
Remove Windows wasm llvm

mcbarton:
PR https://github.com/compiler-research/CppInterOp/pull/298 showed that it is currently not possible to build a wasm version of CppInterOp on Windows. Therefore this PR removes the web assembly build of llvm on Windows from the cache. The cache on the main should be cleared before this PR is merged, and should end up smaller.


vgvassilev (approved):
Thank you, @mcbarton. Lgtm!

PR #319, opened 2024-08-31
Add `GetBinaryOperator`

Vipul-Cariappa:
Implements `GetBinaryOperator` to resolve binary operator in the given scope with specified arguments.


lib/Interpreter/CppInterOp.cpp
```diff
@@ -3161,6 +3161,69 @@ namespace Cpp {
...
+    else if (op == "<")
+      getSema().LookupOverloadedOperatorName(clang::OO_Less, S, lookup);
+    else if (op == "<=")
+      getSema().LookupOverloadedOperatorName(clang::OO_LessEqual, S, lookup);
+    else if (op == "&")
+      getSema().LookupOverloadedOperatorName(clang::OO_Amp, S, lookup);
+    else if (op == "|")
+      getSema().LookupOverloadedOperatorName(clang::OO_Pipe, S, lookup);
+    else if (op == "^")
+      getSema().LookupOverloadedOperatorName(clang::OO_Caret, S, lookup);
```
Vipul-Cariappa, line 3206: What other operators should be here?
Should I include `+=`, `&=`, `&&`, etc?

vgvassilev: We probably should create some enum with the supported operators.

vgvassilev (commented):
We need a test case for this PR.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3161,6 +3161,69 @@ namespace Cpp {
...
+      return nullptr;
+
+    Scope* S = getSema().getScopeForContext(DC);
+    if (!S)
+      return nullptr;
+
+    clang::UnresolvedSet<8> lookup;
+
+    if (op == "+")
+      getSema().LookupOverloadedOperatorName(clang::OO_Plus, S, lookup);
```
vgvassilev, line 3182: Looks like `Sema::LookupBinOp` is a higher-level interface. We should probably use it instead...

include/clang/Interpreter/CppInterOp.h
```diff
@@ -465,7 +483,7 @@ namespace Cpp {
                                                 TCppIndex_t param_index);
 
   ///\returns function that performs operation op on lc and rc
-  TCppFunction_t GetBinaryOperator(TCppScope_t scope, const std::string& op,
+  TCppFunction_t GetBinaryOperator(TCppScope_t scope, enum BinaryOperator op,
```
vgvassilev, line 486: We don't need an elaborated type here. Dropping the enum should be fine.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3178,35 +3178,53 @@ namespace Cpp {
 
     clang::UnresolvedSet<8> lookup;
 
-    if (op == "+")
+    switch (op) {
+    case Plus:
       getSema().LookupOverloadedOperatorName(clang::OO_Plus, S, lookup);
```
vgvassilev, line 3183: Can't we follow the enum ordering in clang. Eg clang::OO_Plus == BinaryOperator.Plus? Then this will be just a single call instead of a switch stmt.

unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -952,3 +952,29 @@ TEST(ScopeReflectionTest, IncludeVector) {
...
+        return MyClass(lhs.x - rhs.x);
+    }
+
+    MyClass operator+(MyClass lhs, MyClass rhs) {
+        return MyClass(lhs.x + rhs.x);
+    }
+  )";
+
+  Cpp::Declare(code.c_str());
+  EXPECT_TRUE(Cpp::GetBinaryOperator(
```
vgvassilev, line 976: Do we get calls to `GetNamed` with things like `operator+`? If so perhaps we can check if the string starts with `operator` and call `GetBinaryOperator`.

include/clang/Interpreter/CppInterOp.h
Vipul-Cariappa: I was thinking of removing our definition of the enum and replacing it with `using clang::BinaryOperatorKind;`. Then we will not require that long `switch ... case ...`  in `CppInterOp.cpp::GetBinaryOperator`. Nor any kind of string comparison.
Any thoughts?

unittests/CppInterOp/ScopeReflectionTest.cpp
Vipul-Cariappa: > Do we get calls to `GetNamed` with things like `operator+`?

I don't think we get such calls.

---

It may not be that simple. What if the string is `operator_something` or `operator123`? Just checking if it starts with `operator` may not be the correct logic here. Also, `GetBinaryOperator` requires the type string of LHS and RHS.

vgvassilev: Fair point.

Vipul-Cariappa:
Hi @vgvassilev, I have made all the changes you suggested. Please have a look.

Valgrind test fails with [`Conditional jump or move depends on uninitialised value`](https://github.com/compiler-research/CppInterOp/actions/runs/10677336267/job/29592320735#step:15:949). I am not sure about the cause here.

unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -952,3 +952,27 @@ TEST(ScopeReflectionTest, IncludeVector) {
   )";
   Interp->process(code);
 }
+
+TEST(ScopeReflectionTest, GetBinaryOperator) {
+  std::string code = R"(
```
aaronj0, line 957:
```suggestion
  if (llvm::sys::RunningOnValgrind())
    GTEST_SKIP() << "XFAIL due to Valgrind report";
  std::string code = R"(
```

include/clang/Interpreter/CppInterOp.h
vgvassilev: We can't leak implementation details for clang. That means we can't include header files of llvm/clang in CppInterOp.h.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3161,6 +3161,44 @@ namespace Cpp {
...
+    if (!S)
+      return nullptr;
+
+    clang::UnresolvedSet<8> lookup;
+
+    getSema().LookupBinOp(S, SourceLocation(), (clang::BinaryOperatorKind)op,
+                          lookup);
+
+    for (NamedDecl* x : lookup) {
+      if (auto* F = llvm::dyn_cast<Decl>(x)) {
```
vgvassilev, line 3185:
```suggestion
    for (NamedDecl* D : lookup) {
      if (auto* FD = llvm::dyn_cast<Decl>(D)) {
```

This is the common convention.

unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -952,3 +952,50 @@ TEST(ScopeReflectionTest, IncludeVector) {
   )";
   Interp->process(code);
 }
+
+TEST(ScopeReflectionTest, GetBinaryOperator) {
+  if (llvm::sys::RunningOnValgrind())
+    GTEST_SKIP() << "XFAIL due to Valgrind report";
```
vgvassilev, line 958: We can drop that as Aaron has added to the suppression file.

vgvassilev (approved):
LGTM! See remaining comments.

unittests/CppInterOp/ScopeReflectionTest.cpp
aaronj0: The InterOp unitttests don't use a a suppression file, they are only for cppyy

vgvassilev: Ok. Then we can ignore this comment.

Vipul-Cariappa:
I have spent more than 3 hours trying to get the following to work.

```C++
// C++ code

class MyClass {
public:
    int x;
    MyClass(int x) : x(x) {}
};

namespace MyScope {

    MyClass operator*(MyClass lhs, MyClass rhs) {
        return MyClass(lhs.x * rhs.x);
    }

} // end namespace MyScope
```

Look up the operator with 
```c++
Cpp::GetBinaryOperator(Cpp::GetScope("MyScope"), Cpp::BinaryOperator::Mul, "MyClass", "MyClass");
```

But it is not working.

---

This is required for the `operator+` for `std::string`; it is defined within the `std` namespace.

PR #320, opened 2024-09-07
GetBinaryOperator returns vector of functions

Vipul-Cariappa:


unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -985,17 +985,16 @@ TEST(ScopeReflectionTest, GetBinaryOperator) {
...
+  Cpp::GetBinaryOperator(Cpp::GetGlobalScope(), Cpp::BinaryOperator::Add, ops);
+  EXPECT_EQ(ops.size(), 3);
+  ops.clear();
+
+  Cpp::GetBinaryOperator(Cpp::GetGlobalScope(), Cpp::BinaryOperator::Sub, ops);
+  EXPECT_EQ(ops.size(), 1);
+  ops.clear();
+
+  Cpp::GetBinaryOperator(Cpp::GetGlobalScope(), Cpp::BinaryOperator::Mul, ops);
+  EXPECT_EQ(ops.size(), 0);
```
vgvassilev, line 999: Da we have an example involving argument dependent lookup?

Vipul-Cariappa: No. I have not implemented the argument dependent lookup with this PR.

I have made changes to take `std::vector&` as an argument and populate it with the operators. Removing the previous string comparisons to resolve the overloaded operator.
We discussed this after the monthly meeting the day before yesterday.

vgvassilev (approved):
Lgtm!

PR #322, opened 2024-09-17
Resolve static attributes of records

Vipul-Cariappa:
Fixes: [test04_class_read_access](https://github.com/Vipul-Cariappa/cppyy-compiler-research/blob/75be0854fbdb830b597b73377a4f95fe8b2108f3/test/test_datatypes.py#L381-L382) and [test05_class_data_write_access](https://github.com/Vipul-Cariappa/cppyy-compiler-research/blob/75be0854fbdb830b597b73377a4f95fe8b2108f3/test/test_datatypes.py#L444-L445) with one-liner change in cppyy.

### DIFF
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -1155,6 +1155,9 @@ namespace Cpp {
         stack_begin.back()++;
       }
 
+      // static field member
+      GetClassDecls<VarDecl>(scope, datamembers);
+
       return datamembers;
     }
 
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -1155,13 +1155,16 @@ namespace Cpp {
         stack_begin.back()++;
       }
 
-      // static field member
-      GetClassDecls<VarDecl>(scope, datamembers);
-
       return datamembers;
     }
 
     return {};
+  }
+
+  std::vector<TCppScope_t> GetStaticDatamembers(TCppScope_t scope) {
+    std::vector<TCppScope_t> datamembers;
+    GetClassDecls<VarDecl>(scope, datamembers);
+    return datamembers;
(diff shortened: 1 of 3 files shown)


Vipul-Cariappa:
@vgvassilev, Should this change be extracted to a separate function `GetStaticDatamembers`? If we extract it to a separate function we will have to modify CPyCppyy slightly.

Vipul-Cariappa:
The failing tests need to be fixed in cppyy with the following patch:
```patch
diff --git a/python/cppyy/__init__.py b/python/cppyy/__init__.py
index 53b7e92..c5d7a3b 100644
--- a/python/cppyy/__init__.py
+++ b/python/cppyy/__init__.py
@@ -133,6 +133,7 @@ def _standard_pythonizations(pyclass, name):
                 return other == -1 or  int(self) == other
             def __ne__(self, other):
                 return other != -1 and int(self) != other
+        del pyclass.__class__.npos          # drop b/c is const data
         pyclass.npos = NPOS(pyclass.npos)
 
     return True
```

Taken from [master](https://github.com/wlav/cppyy/blob/11d990331e7663bf35f1e8c21b35c34124ddb471/python/cppyy/__init__.py#L137).

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1155,15 +1155,18 @@ namespace Cpp {
...
-      // static field member
-      GetClassDecls<VarDecl>(scope, datamembers);
-
       return datamembers;
     }
 
     return {};
   }
 
+  std::vector<TCppScope_t> GetStaticDatamembers(TCppScope_t scope) {
```
vgvassilev, line 1164: We should pass the std::vector as an out parameter.

vgvassilev (approved):
LGTM! @aaronj0, what do you think?

aaronj0 (approved):

Issue #305, opened 2024-06-05
Null access violation when using InstantiateTemplate with variadic class template


deadlocklogic:
I am using the snippet: of the function `static Decl* InstantiateTemplate(TemplateDecl* TemplateD, ArrayRef<TemplateArgument> TemplateArgs, Sema& S)`. It works unless the class template is variadic.
Consider:
```C++
template <typename... T>
class ClassTemplate
{
};
```
Now when analyzing any instantiations of this template like for example: `ClassTemplate<>` I obtain an `ClassTemplateSpecializationDecl`  `CTSD` with `CTSD->getSpecializationKind() == clang::TemplateSpecializationKind::TSK_Undeclared`.
Now when trying to instantiate this class template with:
`InstantiateTemplate(CTSD->getSpecializedTemplate(),
                                            CTSD->getTemplateInstantiationArgs().asArray(), CI.getSema());`
I get null access violation exception.
Any ideas?
Thanks.

vgvassilev:
We have not added support for variadic templates yet. It was at some point in @maximusron's radar.

maximusron:
@deadlocklogic I think we shouldn't be constructing the `TemplateArgumentListInfo` ourselves, can you try using this function : https://github.com/compiler-research/CppInterOp/blob/6a5f5649d2549337456391f672749213118a5b50/lib/Interpreter/CppInterOp.cpp#L2855-L2856 without using `.asArray()` for the args. If it doesn't work I was wondering if we should extract the `clang::PackExpansionType` and use the pattern, but looks like [Sema::DeduceTemplateArguments](https://clang.llvm.org/doxygen/SemaTemplateDeduction_8cpp_source.html#l04437) does that already... @vgvassilev any idea what would be a good way to proceed here?

vgvassilev:
I think we need to check what clang does in these cases and try to replay it in CppInterOp.

vgvassilev:
Can you submit a PR with tests for this change?

deadlocklogic:
Ok, I can submit a PR fixing the issue. I will try a writing a test too.

PR #324, opened 2024-09-19
Test variadic template support

aaronj0:


vgvassilev (commented):
We don't seem to have tests for the new code.

aaronj0:
> We don't seem to have tests for the new code.

There is an added unittest... It is still a draft though.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -589,6 +589,34 @@ TEST(FunctionReflectionTest, InstantiateTemplateMethod) {
   EXPECT_TRUE(TA1.getAsType()->isIntegerType());
 }
 
+TEST(FunctionReflectionTest, InstantiateVariadicFunctionTemplate) {
+  std::vector<Decl*> Decls;
+  std::string code = R"(
+template<typename... Args> void VariadicFnTemplate(Args... args) {}
```
vgvassilev, line 595: Can we include the example of #305?

aaronj0: The current test mimics the same variadic template, I am extending it to initialize using a template pack which should cover the other branch for codecov

vgvassilev: Still worth adding the test from the issue and adding code to support it.

mcbarton:
@aaronj0 to get past the undefined symbols error with the Emscripten jobs, add the symbols in this file https://github.com/aaronj0/CppInterOp/blob/variadic-support/lib/Interpreter/exports.ld , following the pattern already in the file. Eventually we will work out why these symbols are not being exported, fix it, and remove the file, but for now this is what we are doing for the moment when a new llvm or clang symbols is not being exported.

vgvassilev (approved):
LGTM!

PR #329, opened 2024-09-25
using `getFirstDecl` for `TranslationUnitDecl` for consistent pointers

Vipul-Cariappa:
Eight previously failing tests pass without any new failing tests in cppyy.
image

But `EnumReflectionTest.GetEnums` fails from the CppInterOp tests.
It looks like `GetEnums` cannot resolve any `enum`s in the global scope.

### DIFF
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -498,7 +498,7 @@ namespace Cpp {
 
   TCppScope_t GetGlobalScope()
   {
-    return getSema().getASTContext().getTranslationUnitDecl();
+    return getSema().getASTContext().getTranslationUnitDecl()->getFirstDecl();
   }
 
   static Decl *GetScopeFromType(QualType QT) {
@@ -604,8 +604,12 @@ namespace Cpp {
     if (!ParentDC)
       return 0;
 
-    return (TCppScope_t) clang::Decl::castFromDeclContext(
-            ParentDC)->getCanonicalDecl();
+    auto* P = clang::Decl::castFromDeclContext(ParentDC)->getCanonicalDecl();
+
+    if (auto* TU = llvm::dyn_cast_or_null<TranslationUnitDecl>(P))
+      return (TCppScope_t)TU->getFirstDecl();
+
+    return (TCppScope_t)P;
   }
 
   TCppIndex_t GetNumBases(TCppScope_t klass)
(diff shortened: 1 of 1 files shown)


Vipul-Cariappa:
ping @vgvassilev @aaronj0 
Any ideas on what is causing the CppInterOp failure?

vgvassilev:
Looks like some test is failing.

vgvassilev:
> Eight previously failing tests pass without any new failing tests in cppyy. image
> 
> But `EnumReflectionTest.GetEnums` fails from the CppInterOp tests. It looks like `GetEnums` cannot resolve any `enum`s in the global scope.

The problem is that the implementation is insufficient. To fix it, we need to iterate the redeclaration chains of NamespaceDecl and TranslationUnitDecl. That is, we should check if the Decl is Redeclarable and then do `getMostRecentDecl()` and loop until there is `getPreviousDecl()`...

vgvassilev:
> > Eight previously failing tests pass without any new failing tests in cppyy. image
> > But `EnumReflectionTest.GetEnums` fails from the CppInterOp tests. It looks like `GetEnums` cannot resolve any `enum`s in the global scope.
> 
> The problem is that the implementation is insufficient. To fix it, we need to iterate the redeclaration chains of NamespaceDecl and TranslationUnitDecl. That is, we should check if the Decl is Redeclarable and then do `getMostRecentDecl()` and loop until there is `getPreviousDecl()`...

Even better `DeclContext::collectAllContexts` does it for us.

Vipul-Cariappa:
> The problem is that the implementation is insufficient. To fix it, we need to iterate the redeclaration chains of NamespaceDecl and TranslationUnitDecl.

Fixed. 👍🏽 

---

Maybe there are other places where we will need to use `DeclContext::collectAllContexts` to resolve symbols.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3082,23 +3082,26 @@ namespace Cpp {
...
       DC = clang::NamespaceDecl::castToDeclContext(ND);
-      decl = DC->decls_begin();
+      DC->collectAllContexts(DCs);
     } else if (auto *TUD = dyn_cast_or_null<TranslationUnitDecl>(D)) {
       DC = clang::TranslationUnitDecl::castToDeclContext(TUD);
-      decl = DC->decls_begin();
+      DC->collectAllContexts(DCs);
     } else {
       return;
     }
```
vgvassilev, line 3098: We need something:

if(!isa<DeclContext>(D))
  return;
DC = cast<DeclContext>(D);
DC->collect(...);
...

vgvassilev:
> > The problem is that the implementation is insufficient. To fix it, we need to iterate the redeclaration chains of NamespaceDecl and TranslationUnitDecl.
> 
> Fixed. 👍🏽
> 
> Maybe there are other places where we will need to use `DeclContext::collectAllContexts` to resolve symbols.

Hopefully not, because this iteration generally is wrong. This need to be lookup-based approach instead of a brute forcing all namespaces and translation units... Please add a giant FIXME describing that.

lib/Interpreter/CppInterOp.cpp
Vipul-Cariappa: It may not be required.
We already check for [`TagDecl`](https://clang.llvm.org/doxygen/classclang_1_1TagDecl.html), [`NamespaceDecl`](https://clang.llvm.org/doxygen/classclang_1_1NamespaceDecl.html), & [`TranslationUnitDecl`](https://clang.llvm.org/doxygen/classclang_1_1TranslationUnitDecl.html).
And looking at the documentation for them, they inherited `DeclContext`.

Let me know, if you still think it is necessery.

Vipul-Cariappa: I have fixed it. Please have a look.

vgvassilev: Looks good but I am not seeing my fixme comment addressed.

Vipul-Cariappa: Done. Sorry, I missed it earlier.

vgvassilev (approved):
LGTM!

aaronj0:
@Vipul-Cariappa Can you squash into a single commit?

vgvassilev:
The commit does more than it says?

Vipul-Cariappa:
> The commit does more than it says?

The main purpose of the PR is to use `getFirstDecl` for `TranslationUnitDecl`. That breaks `GetEnums` therefore we had to refactor it to use `collectAllContexts`.

Hope the new commit message conveys the same.

PR #330, opened 2024-09-26
Use `getExecutionEngine` to obtain the JIT for Cling > LLVM16

aaronj0:
To also be backported to the root-v1.0 tag

### DIFF
--- a/lib/Interpreter/Compatibility.h
+++ b/lib/Interpreter/Compatibility.h
@@ -85,17 +85,36 @@ inline void maybeMangleDeclName(const clang::GlobalDecl& GD,
   cling::utils::Analyze::maybeMangleDeclName(GD, mangledName);
 }
 
-inline llvm::orc::LLJIT* getExecutionEngine(const cling::Interpreter& I) {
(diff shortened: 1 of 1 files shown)


lib/Interpreter/Compatibility.h
```diff
@@ -85,17 +85,35 @@ inline void maybeMangleDeclName(const clang::GlobalDecl& GD,
...
+#endif
+
+// Note: The offsets changed in LLVM 16 with the introduction of
+#if CLANG_VERSION_MAJOR == 16
+#ifdef __APPLE__
+  const unsigned m_ExecutorOffset = 62;
+#else
+  const unsigned m_ExecutorOffset = 72;
+#endif // __APPLE__
+#endif
```
vgvassilev, line 115: How is this block different from the previous?

aaronj0: The offsets have not yet been updated for cling/llvm 16, will do

aaronj0: @vgvassilev done

vgvassilev (approved):
Lgtm!

PR #331, opened 2024-09-26
[bp] Use `getExecutionEngine` to obtain the JIT for Cling > LLVM16

aaronj0:


aaronj0:
Backporting as #330 has been merged

PR #333, opened 2024-10-21
[ci] Add latest cling versions(LLVM 16, 18)

aaronj0:


mcbarton:
@aaronj0 You'll need to delete the jobs related to cling and llvm 13, and just keep the job related to llvm 18. If you don't the llvm cache will be overwhelmed.

.github/workflows/ci.yml
```diff
@@ -81,74 +89,94 @@ jobs:
...
             os: windows-2022
             compiler: msvc
-            clang-runtime: '13'
+            clang-runtime: '16'
             cling: On
-            cling-version: '1.0'
+            cling-version: 'v1.1'
             cppyy: Off
-          - name: osx14-arm-clang-clang-repl-19
+          - name: osx14-arm-clang-repl-19
```
mcbarton, line 105: With the exception of Windows the names of jobs are currently
'operation system'-architechture-compiler-'cling or clang-repl version'

.github/workflows/ci.yml
```diff
@@ -81,74 +89,94 @@ jobs:
...
+            os: windows-2022
+            compiler: msvc
+            clang-runtime: '18'
+            cling: On
+            cling-version: 'refs/heads/master'
+          - name: win2022-msvc-cling-16
             os: windows-2022
             compiler: msvc
-            clang-runtime: '13'
+            clang-runtime: '16'
```
mcbarton, line 101: Please remove the clang llvm 16 jobs as this will overwhelm the llvm cache. We need to just keep the latest version of llvm with respect to cling.

.github/workflows/ci.yml
aaronj0: I would prefer easier readability. Compiler backend are either clang-repl or cling. 'clang-clang-repl' does not make sense

mcbarton: @aaronj0 That's fine to do but we need to be consistent, so you've need to drop the msvc and gcc12 from the other jobs for consistency.

.github/workflows/ci.yml
aaronj0: We need both. Cling differs with the LLVM-18 upgrade and ROOT requires this check to be made to support releases >6.32 . I would rather drop clang-repl16

mcbarton: @aaronj0 We need to drop one job. If you think it would be better to drop clang-repl-16 then I am ok with them. We can continue to support it in the other repos for now since they don't have any Windows builds filling in the cache.

PR #338, opened 2024-10-27
Refactoring the API for multiple interpreter instance support

Gnimuc:
I feel like we also need to split the implementation and use an internal shared header for interfaces.

### DIFF
--- a/include/clang/Interpreter/CppInterOp.h
+++ b/include/clang/Interpreter/CppInterOp.h
@@ -77,7 +77,8 @@ namespace Cpp {
   /// function, constructor or destructor.
   class JitCall {
   public:
-    friend CPPINTEROP_API JitCall MakeFunctionCallable(TCppConstFunction_t);
+    friend CPPINTEROP_API JitCall
+    MakeFunctionCallable(TInterp_t I, TCppConstFunction_t func);
     enum Kind : char {
       kUnknown = 0,
       kGenericCall,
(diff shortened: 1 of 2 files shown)


lib/Interpreter/CppInterOp.cpp
```diff
@@ -3262,15 +3279,24 @@ namespace Cpp {
     return nullptr;
   }
 
-  void Destruct(TCppObject_t This, TCppScope_t scope, bool withFree /*=true*/) {
-    Decl* Class = (Decl*)scope;
-    if (auto wrapper = make_dtor_wrapper(getInterp(), Class)) {
+  TCppObject_t Construct(TCppScope_t scope, void* arena /*=nullptr*/) {
```
Gnimuc, line 3282: also removed the `Impl` suffix.

include/clang/Interpreter/CppInterOp.h
```diff
@@ -78,6 +78,8 @@ namespace Cpp {
   class JitCall {
   public:
     friend CPPINTEROP_API JitCall MakeFunctionCallable(TCppConstFunction_t);
```
vgvassilev, line 80: Do we need this as a friend as it is now just forwarding?

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1453,6 +1456,10 @@ namespace Cpp {
     }
   }
 
+  QualType GetBuiltinType(const std::string& name, Sema& sema) {
```
vgvassilev, line 1459: This one seems unused and not part of the header. Maybe an oversight from the bigger PR?

lib/Interpreter/CppInterOp.cpp
```diff
@@ -2574,26 +2581,31 @@ namespace Cpp {
     } // namespace
       // End of JitCall Helper Functions
 
-  CPPINTEROP_API JitCall MakeFunctionCallable(TCppConstFunction_t func) {
-    auto* D = (const clang::Decl*)func;
-    if (!D)
-      return {};
+    JitCall MakeFunctionCallableImpl(TInterp_t I, TCppConstFunction_t func) {
+      auto* D = (const clang::Decl*)func;
```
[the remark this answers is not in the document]
vgvassilev: I think that's a good suggestion..

vgvassilev (approved):
Lgtm! Thank you, @Gnimuc!

vgvassilev:
Can you squash the commits to have clear history?

Gnimuc:
> Can you squash the commits to have clear history?

Done!

PR #340, opened 2024-10-28
Change standard clang format to Cpp14

mcbarton:
C++14 is stated as the minimum standard acceptable by roots llvm 13 that we use, so I've updated the clang-format file to reflect this standard.


vgvassilev (approved):
LGTM!

PR #341, opened 2024-10-28
Remove Fortran files from .gitignore

mcbarton:
We do not create any Fortran files during the build process, so seems unneeded to have them in the .gitignore


vgvassilev (approved):
LGTM!

PR #343, opened 2024-10-28
Remove random coverage flags ci

mcbarton:
Coverage was set to try for all llvm cache builds. Didn't do anything to effect results but needs removing. Coverage was set to true for both llvm 18 and llvm 19 builds of CppInterOp. I've removed the true flag from the llvm 18 build so we have consistent coverage report results going forward.


vgvassilev (approved):
LGTM!

Issue #295, opened 2024-05-19
Update Google Test used


mcbarton:
Currently we download version 1.12.1 . After this version Google Test required c++14 which is the minimum standard of this project. I suggest we move to a newer version.

PR #344, opened 2024-10-28
Update gtest used to v1.15.2

mcbarton:
This PR will attempt to update the version of gtest we are using to version 1.15.2 .
Fixes https://github.com/compiler-research/CppInterOp/issues/295

### DIFF
--- a/cmake/modules/GoogleTest.cmake
+++ b/cmake/modules/GoogleTest.cmake
@@ -50,7 +50,6 @@ ExternalProject_Add(
   TIMEOUT 600
   )
 
-
 # Specify include dirs for gtest and gmock
 ExternalProject_Get_Property(googletest source_dir)
 set(GTEST_INCLUDE_DIR ${source_dir}/googletest/include)
--- a/cmake/modules/GoogleTest.cmake
+++ b/cmake/modules/GoogleTest.cmake
@@ -16,7 +16,12 @@ if(WIN32)
     -Dgtest_force_shared_crt=ON
     BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config $<CONFIG>)
 elseif(APPLE)
+  #FIXME: Temporarily remove -pedantic from CMAKE_CXX_FLAGS to avoid issues with building
+  #       latest version of gtest on MacOS (use of -Wc++17-attribute-extensions causes 
+  #       issues due to [[maybe_unused]] in gtest)
   set(EXTRA_GTEST_OPTS -DCMAKE_OSX_SYSROOT=${CMAKE_OSX_SYSROOT})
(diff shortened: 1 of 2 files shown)


cmake/modules/GoogleTest.cmake
```diff
@@ -17,14 +17,15 @@ if(WIN32)
     BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config $<CONFIG>)
 elseif(APPLE)
   set(EXTRA_GTEST_OPTS -DCMAKE_OSX_SYSROOT=${CMAKE_OSX_SYSROOT})
+  string(REPLACE "-Wc++17-attribute-extensions" "" CMAKE_CXX_FLAGS "${CMAKE_Cxx_FLAGS}")
```
vgvassilev, line 20:
```suggestion
  string(REPLACE "-Wc++17-attribute-extensions" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```

vgvassilev (approved):
LGTM!

vgvassilev:
Please squash on merge.

mcbarton:
@vgvassilev This PR needs another review, as I had to take a different approach after your last review revealed an error in my previous approach.

cmake/modules/GoogleTest.cmake
```diff
@@ -16,15 +16,20 @@ if(WIN32)
     -Dgtest_force_shared_crt=ON
     BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config $<CONFIG>)
 elseif(APPLE)
+  #FIXME: Temporarily remove -pedantic from CMAKE_CXX_FLAGS to avoid issues with building
+  #       latest version of gtest on MacOS (use of -Wc++17-attribute-extensions causes 
+  #       issues due to [[maybe_unused]] in gtest)
   set(EXTRA_GTEST_OPTS -DCMAKE_OSX_SYSROOT=${CMAKE_OSX_SYSROOT})
+  set(CMAKE_CXX_FLAGS_TEMP_COPY ${CMAKE_CXX_FLAGS})
+  string(REPLACE "-pedantic" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
vgvassilev, line 24: I am wondering if we need to propagate `CMAKE_CXX_COMPILER` to the `CMAKE_ARGS`. Alternatively in `CMAKE_ARGS` we can append `-no-pedantic` to disable the default mode.

vgvassilev: ping.

mcbarton:
@vgvassilev Rather than add `-no-pedantic` in the osx build I just found where `pedantic` was being applied in the osx builds and removed. Is the PR good now?

PR #349, opened 2024-11-04
[ci] No longer run cppyy with clang-repl16

aaronj0:
This is due to the recently enabled test on cppyy that crashes on clang 16 and passes with clang >=17


mcbarton:
@aaronj0 why only disable cppyy job for clang-16 on Ubuntu and not osx too?

PR #353, opened 2024-11-19
Remove include new flags while creating interpreter

anutosh491:


anutosh491:
@mcbarton and I are interested in this change.

We think it is specific to cppyy. To test it here's a PR that removes it from CppInterOp. We now check if the CppInterOp tests pass. If yes, we could then we can make a matching PR which adds the flags it to cppyy.

anutosh491:
Hey, I can confirm that this change is not required for now. 
Probably the new flag should be removed down the line but we can keep it for now I guess.

Closing till we find a way to do this !

PR #357, opened 2024-11-28
Reduce size of llvm cache by not build webassembly target for llvm<19

mcbarton:
@vgvassilev Due to the work to get the wasm version of xeus-cpp working we will have a wasm version of CppInterOp we will be able to test, and keep on top of from version 19. Although it will build, I suspect our wasm version of CppInterOp for llvm<19 won't work. Therefore I suggest we don't build the web assembly target for llvm<19, and remove the wasm jobs for llvm<19 so that ci runs are quicker. I've cleared the cache for 2 jobs before putting the PR in so that we can test if my changes work. If you're happy with the changes in this PR the cache will need to be cleared before merging. With the space saved in the cache by this change should give us the room for when llvm 20 is released.


anutosh491:
> Due to the work to get the wasm version of xeus-cpp working we will have a wasm version of CppInterOp we will be able to test, and keep on top of from version 19. Although it will build, I suspect our wasm version of CppInterOp for llvm<19 won't work. Therefore I suggest we don't build the web assembly target for llvm<19, and remove the wasm jobs for llvm<19 so that ci runs are quicker.

Hey @mcbarton
I think I know on how to get a proper wasm build for CppInterOp that can be put to use for downstream projects like xeus-cpp (have been playing around locally) . It should also rid of the numerous undefined symbols out of clang as raised here (https://github.com/compiler-research/CppInterOp/issues/334) .

For that I would possibly like 
1) llvm 19.1.5 to be out (and the PR getting clang-repl to work in the browser to be a part of it https://github.com/llvm/llvm-project/pull/117978)
2) CppInterOp to be based on this llvm version.
3) Then I could raise a PR introducing a useable wasm build.

Maybe we could raise an issue (and possibly assign it to me) till then ?

PR #358, opened 2024-11-29
Update cling in ci to version 1.1 using root llvm 16

mcbarton:
This PR will update cling in the ci to version 1.1 using root llvm 16

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,12 +53,12 @@ jobs:
             cppyy: On
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
-          - name: ubu22-x86-gcc9-clang13-cling
+          - name: ubu22-x86-gcc9-clang16-cling
             os: ubuntu-22.04
-            compiler: gcc-9
-            clang-runtime: '13'
+            compiler: gcc-12
+            clang-runtime: '16'
             cling: On
-            cling-version: '1.0'
+            cling-version: '1.1'
             cppyy: On
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
@@ -118,12 +118,12 @@ jobs:
             cppyy: On
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
-          - name: osx14-arm-clang-clang13-cling
+          - name: osx14-arm-clang-clang16-cling
             os: macos-14
             compiler: clang
-            clang-runtime: '13'
+            clang-runtime: '16'
             cling: On
-            cling-version: '1.0'
+            cling-version: '1.1'
             cppyy: On
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
@@ -159,12 +159,12 @@ jobs:
             cppyy: On
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
-          - name: osx13-x86-clang-clang13-cling
+          - name: osx13-x86-clang-clang16-cling
             os: macos-13
             compiler: clang
-            clang-runtime: '13'
+            clang-runtime: '16'
             cling: On
-            cling-version: '1.0'
+            cling-version: '1.1'
             cppyy: On
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
@@ -552,12 +552,12 @@ jobs:
             clang-runtime: '16'
             cling: Off
             cppyy: Off
-          - name: ubu22-x86-gcc9-clang13-cling-cppyy
+          - name: ubu22-x86-gcc9-clang16-cling-cppyy
             os: ubuntu-22.04
-            compiler: gcc-9
-            clang-runtime: '13'
+            compiler: gcc-12
+            clang-runtime: '16'
             cling: On
-            cling-version: '1.0'
+            cling-version: '1.1'
             cppyy: On
           #FIXME: Windows CppInterOp tests expected to fail
           #until https://github.com/compiler-research/CppInterOp/issues/188 is solved
@@ -609,12 +609,12 @@ jobs:
             clang-runtime: '16'
             cling: Off
             cppyy: Off
-          - name: osx14-arm-clang-clang13-cling-cppyy
+          - name: osx14-arm-clang-clang16-cling-cppyy
             os: macos-14
             compiler: clang
-            clang-runtime: '13'
+            clang-runtime: '16'
             cling: On
-            cling-version: '1.0'
+            cling-version: '1.1'
             cppyy: On
           - name: osx13-x86-clang-clang-repl-19-cppyy
             os: macos-13
@@ -640,12 +640,12 @@ jobs:
             clang-runtime: '16'
             cling: Off
             cppyy: Off
-          - name: osx13-x86-clang-clang13-cling-cppyy
+          - name: osx13-x86-clang-clang16-cling-cppyy
             os: macos-13
             compiler: clang
-            clang-runtime: '13'
+            clang-runtime: '16'
             cling: On
-            cling-version: '1.0'
+            cling-version: '1.1'
             cppyy: On
 
     steps:


aaronj0:
@mcbarton this is already ongoing in #333, rebasing that PR should do the job. The reason cling 16 should build now is because the offsets to obtain the JIT was changed. I just never rebased the CI pr after that

mcbarton:
> @mcbarton this is already ongoing in #333, rebasing that PR should do the job. The reason interop should build now is because the offsets to obtain the JIT was changed in #330 . I just never rebased the CI pr after that

@vgvassilev @aaronj0 https://github.com/compiler-research/CppInterOp/pull/333 has issues for me that are easier to resolve by just starting a new PR. We shouldn't have the cling version set to 'refs/heads/master' . This will just cause the cache is fill up quickly. The names have been made inconsistent by dropping the compiler from the job name on osx. Its just brought back lots of jobs that were removed from the ci.

aaronj0:
> > @mcbarton this is already ongoing in #333, rebasing that PR should do the job. The reason interop should build now is because the offsets to obtain the JIT was changed in #330 . I just never rebased the CI pr after that
> 
> @vgvassilev @aaronj0 #333 has issues for me that are easier to resolve by just starting a new PR. We shouldn't have the cling version set to 'refs/heads/master' . This will just cause the cache is fill up quickly. The names have been made inconsistent by dropping the compiler from the job name on osx. Its just brought back lots of jobs that were removed from the ci.

It's not ready yet... The reason you see the old jobs is because it's many commits behind, but that is relatively easy to resolve which I am doing right now. I don't think opening a fresh PR is the solution.

> We shouldn't have the cling version set to 'refs/heads/master'. This will just cause the cache is fill up quickly
As for 'refs/head/master', cling with llvm18 doesn't have a release tag yet. The purpose of this is to have a real-time test of upstream cling

mcbarton:
> > @vgvassilev @aaronj0 #333 has issues for me that are easier to resolve by just starting a new PR. We shouldn't have the cling version set to 'refs/heads/master' . This will just cause the cache is fill up quickly. The names have been made inconsistent by dropping the compiler from the job name on osx. Its just brought back lots of jobs that were removed from the ci.
> 
> It's not ready yet... The reason you see the old jobs is because it's many commits behind, but that is relatively easy to resolve which I am doing right now. I don't think opening a fresh PR is the solution.
> 
> > We shouldn't have the cling version set to 'refs/heads/master'. This will just cause the cache is fill up quickly
> 
> As for 'refs/head/master', cling with llvm18 doesn't have a release tag yet. The purpose of this is to have a real-time test of upstream cling

@aaronj0 We can't have a real time test of the cling based on llvm 18 in the ci as it would be constantly be rebuilding the cache, overwhelming the space we have very quickly. If you drop this from the PR and make it specifically about cling 1.1 and llvm 16, and revert the names back to be consistent then I'll drop this PR.

aaronj0:
> > > @vgvassilev @aaronj0 #333 has issues for me that are easier to resolve by just starting a new PR. We shouldn't have the cling version set to 'refs/heads/master' . This will just cause the cache is fill up quickly. The names have been made inconsistent by dropping the compiler from the job name on osx. Its just brought back lots of jobs that were removed from the ci.
> > 
> > 
> > It's not ready yet... The reason you see the old jobs is because it's many commits behind, but that is relatively easy to resolve which I am doing right now. I don't think opening a fresh PR is the solution.
> > > We shouldn't have the cling version set to 'refs/heads/master'. This will just cause the cache is fill up quickly
> > 
> > 
> > As for 'refs/head/master', cling with llvm18 doesn't have a release tag yet. The purpose of this is to have a real-time test of upstream cling
> 
> @aaronj0 We can't have a real time test of the cling based on llvm 18 in the ci as it would be constantly be rebuilding the cache, overwhelming the little space we have very quickly. If you drop this from the PR and make it specifically about cling 1.1 and llvm 16, and revert the names back to be consistent then I'll drop this PR.

We will cut a tag for it, or use a commit. Either way I don't see the point of duplicating the previous PR

Vipul-Cariappa (commented):
LGTM! Maybe we should also update the README with updated versions for build instructions.

---

I spent some time debugging the failure. It is caused when there is an error in the code we are executing ([source](https://github.com/compiler-research/CppInterOp/blob/f794dee0a636b8c958927aa527440474c28de5b8/unittests/CppInterOp/InterpreterTest.cpp#L69C1-L70C69)).

Stack trace goes through

https://github.com/root-project/cling/blob/b5e06a2a7b0832f034d7e405f42f0770364b607f/lib/Interpreter/IncrementalParser.cpp#L914

Destructor of `llvm::Expected` https://github.com/root-project/llvm-project/blob/cling-llvm16/llvm/include/llvm/Support/Error.h#L551-L552

https://github.com/root-project/llvm-project/blob/cling-llvm16/llvm/include/llvm/Support/Error.h#L714-L715

The logic looks valid in my eyes.

The error msg is
```
Expected<T> must be checked before access or destruction.
Unchecked Expected<T> contained error
```

With the following patch for cling the CppInterOp tests pass

```patch
diff --git a/lib/Interpreter/IncrementalParser.cpp b/lib/Interpreter/IncrementalParser.cpp
index 89cd78d..2aea40d 100644
--- a/lib/Interpreter/IncrementalParser.cpp
+++ b/lib/Interpreter/IncrementalParser.cpp
@@ -911,8 +911,11 @@ namespace cling {
     PP.EnterSourceFile(FID, /*DirLookup*/nullptr, NewLoc);
     m_Consumer->getTransaction()->setBufferFID(FID);
 
-    if (!ParseOrWrapTopLevelDecl())
+    llvm::Expected<bool> res = ParseOrWrapTopLevelDecl();
+    if (!res) {
+      llvm::consumeError(std::move(res.takeError()));
       return kFailed;
+    }
 
     if (PP.getLangOpts().DelayedTemplateParsing) {
       // Microsoft-specific:
```

@vgvassilev may have to take a look once.

vgvassilev:
> LGTM! Maybe we should also update the README with updated versions for build instructions.
> 
> I spent some time debugging the failure. It is caused when there is an error in the code we are executing ([source](https://github.com/compiler-research/CppInterOp/blob/f794dee0a636b8c958927aa527440474c28de5b8/unittests/CppInterOp/InterpreterTest.cpp#L69C1-L70C69)).
> 
> Stack trace goes through
> 
> https://github.com/root-project/cling/blob/b5e06a2a7b0832f034d7e405f42f0770364b607f/lib/Interpreter/IncrementalParser.cpp#L914
> 
> Destructor of `llvm::Expected` https://github.com/root-project/llvm-project/blob/cling-llvm16/llvm/include/llvm/Support/Error.h#L551-L552
> 
> https://github.com/root-project/llvm-project/blob/cling-llvm16/llvm/include/llvm/Support/Error.h#L714-L715
> 
> The logic looks valid in my eyes.
> 
> The error msg is
> 
> ```
> Expected<T> must be checked before access or destruction.
> Unchecked Expected<T> contained error
> ```
> 
> With the following patch for cling the CppInterOp tests pass
> 
> ```diff
> diff --git a/lib/Interpreter/IncrementalParser.cpp b/lib/Interpreter/IncrementalParser.cpp
> index 89cd78d..2aea40d 100644
> --- a/lib/Interpreter/IncrementalParser.cpp
> +++ b/lib/Interpreter/IncrementalParser.cpp
> @@ -911,8 +911,11 @@ namespace cling {
>      PP.EnterSourceFile(FID, /*DirLookup*/nullptr, NewLoc);
>      m_Consumer->getTransaction()->setBufferFID(FID);
>  
> -    if (!ParseOrWrapTopLevelDecl())
> +    llvm::Expected<bool> res = ParseOrWrapTopLevelDecl();
> +    if (!res) {
> +      llvm::consumeError(std::move(res.takeError()));
>        return kFailed;
> +    }
>  
>      if (PP.getLangOpts().DelayedTemplateParsing) {
>        // Microsoft-specific:
> ```
> 
> @vgvassilev may have to take a look once.

Makes sense. Please open a pr.

Vipul-Cariappa:
> Makes sense. Please open a pr.

PR at cling?

vgvassilev:
> > Makes sense. Please open a pr.
> 
> PR at cling?

Yes.

Vipul-Cariappa:
Hi @vgvassilev I have opened the PR at https://github.com/root-project/cling/pull/538.

I was thinking... after merging the PR at Cling, we will have to create a new release, `v1.2`, before we can use that fix here?
Is it better to just include the fix as a patch file in `patches` directory in this repo and apply it?

FYI, with this patch, CppInterOp tests pass, but cppyy tests fail. I will start looking into it.

Any thoughts @mcbarton

vgvassilev:
> Hi @vgvassilev I have opened the PR at [root-project/cling#538](https://github.com/root-project/cling/pull/538).
> 
> I was thinking... after merging the PR at Cling, we will have to create a new release, `v1.2`, before we can use that fix here? Is it better to just include the fix as a patch file in `patches` directory in this repo and apply it?
> 
> FYI, with this patch, CppInterOp tests pass, but cppyy tests fail. I will start looking into it.
> 
> Any thoughts @mcbarton

Yeah, I could release cling...

mcbarton:
@Vipul-Cariappa I have added your patch to the workflow to see it working. I have disabled building cppyy, until you determine what is the issue stopping cppyy from passing with cling 1.1. If this patch works then I will update the documentation.

mcbarton:
@Vipul-Cariappa any idea how to fix the failing osx jobs? The errror is the same as what you fixed for Ubuntu.

Vipul-Cariappa:
Debugging OSX failures at #359.

Looks like `Cpp::CreateInterpreter` fails.
Backtrace:
Screenshot From 2024-12-02 09-02-10

I will need to rebuild cling and llvm in debug. It is going to be time consuming.

Vipul-Cariappa:
Hi @mcbarton.

If I run `git diff` on cling, I do not see my patch applied.
I guess my patch is not applied?

Do we need to rebuild the cache?

Or update the logic of retrieving the cache?
I see some complicated key in the retrieval step
```yaml
path: |
          llvm-project
          ${{ matrix.cling=='On' && 'cling' || '' }}
key: ${{ env.CLING_HASH }}-${{ runner.os }}-${{ matrix.os }}-${{ matrix.compiler }}-clang-${{ matrix.clang-runtime }}.x-patch-${{ hashFiles(format('patches/llvm/clang{0}-*.patch', matrix.clang-runtime)) || 'none' }}
```

I am coming to this conclusion by looking at the CI run of https://github.com/compiler-research/CppInterOp/pull/358/commits/dbc5d38133b0a24d7ab638390ec4fa890355e3e9 ([Update cling16-1-Error-Handling.patch to add blank line](https://github.com/compiler-research/CppInterOp/pull/358/commits/dbc5d38133b0a24d7ab638390ec4fa890355e3e9)) commit. I see that ubuntu-cling still fails for the same reason.
With that patch Ubuntu-Cling should pass, at least it passes in my local machine.

mcbarton:
@Vipul-Cariappa It passes with your patch. See https://github.com/compiler-research/CppInterOp/actions/runs/12100202323/job/33748369555

Vipul-Cariappa:
Weird. Looks like the OSX failure may not be caused in this round of CI run.
I am speculating here., as OSX-arm passed. Waiting for OSX-X86 and Ubuntu to finish.

---

> @Vipul-Cariappa The Ubuntu job passes with your patch. See https://github.com/compiler-research/CppInterOp/actions/runs/12100202323/job/33748369555

Which commit is this?

Any idea why Ubuntu-Cling fails in my PR https://github.com/compiler-research/CppInterOp/actions/runs/12112309439/job/33766010006?pr=359? It fails with the same error, my patch is supposed to fix.

alexander-penev (approved):

alexander-penev (approved):

.github/workflows/ci.yml
```diff
@@ -118,12 +118,12 @@ jobs:
             cppyy: On
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
-          - name: osx14-arm-clang-clang13-cling
+          - name: osx14-arm-clang-clang16-cling
             os: macos-14
             compiler: clang
-            clang-runtime: '13'
```
vgvassilev, line 124: We should probably keep both versions of cling for now...

vgvassilev (changes requested):

mcbarton:
> Hi @mcbarton, could you please reopen these changes. We are now interested to get this merged quickly (I guess, at least I am).
> 
> @aaronj0, @vgvassilev can we skip cling `v1.1` and directly try using [`v1.2`](https://github.com/root-project/cling/releases/tag/v1.2)?

@Vipul-Cariappa I will look into this today. I'll make a fresh PR which adds cling 1.1 for linux to the ci (since we know that works, and the issues is  macos) and i'll open up a separate PR for cling 1.2

Vipul-Cariappa:
> I'll make a fresh PR which adds cling 1.1 for linux to the ci (since we know that works, and the issues is macos) and i'll open up a separate PR for cling 1.2

Just `1.2` should be sufficient, I don't want to deal with old versions of LLVM. Unless Aaron or Vassil wants to support `1.1`.

mcbarton:
> > I'll make a fresh PR which adds cling 1.1 for linux to the ci (since we know that works, and the issues is macos) and i'll open up a separate PR for cling 1.2
> 
> Just `1.2` should be sufficient, I don't want to deal with old versions of LLVM. Unless Aaron or Vassil wants to support `1.1`.

I think its best if we have the 1.1 job in the ci for linux. It would look strange if we support 1.0 and 1.2, but not 1.1 at some level at least (with at least meaning just linux). If Aaron and Vassil agree with you though that's fine. We would probably need some sort of note to that effect though in the documentation, if that was the case.

Vipul-Cariappa:
> I think its best if we have the 1.1 job in the ci for linux. It would look strange if we support 1.0 and 1.2, but not 1.1 at some level at least (with at least meaning just linux).

Once we support `1.2`, we plan on dropping `1.0`. So it should be ok if we don't support `1.1`.
Maybe we can have this discussion later, once we have something.
Please note that I will be prioritizing `1.2` for now.

mcbarton:
> > I'll make a fresh PR which adds cling 1.1 for linux to the ci (since we know that works, and the issues is macos) and i'll open up a separate PR for cling 1.2
> 
> Just `1.2` should be sufficient, I don't want to deal with old versions of LLVM. Unless Aaron or Vassil wants to support `1.1`.

It should be noted that for clang repl, where we couldn't get a feature working for an older llvm, we just skip over the test. The point of having the old llvms in the ci is more to check that an existing feature that already works isn't broken by a patch

vgvassilev:
I think supporting the latest released cling is fine.

PR #360, opened 2024-12-02
Make -DUSE_REPL=On default

### LINKED ISSUE #255: Make -DUSE_REPL=On default and remove the defines of USE_REPL.

faze-geek:
Fixes #255


faze-geek:
Will test this locally soon.

CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
   endif()
 
   option(USE_CLING "Use Cling as backend" OFF)
-  option(USE_REPL "Use clang-repl as backend" OFF)
+  option(USE_REPL "Use clang-repl as backend" ON)
```
anutosh491, line 46: As the block below points out the following
```
  if (USE_CLING AND USE_REPL)
    message(FATAL_ERROR "We can only use Cling (USE_CLING=On) or Repl (USE_REPL=On), but not both of them.")
  endif()
```

What can be done if we are using REPL as default ....  is to make cling and REPL mutually exclusive. Which means we only use 1 of these by default.
1) REPL by default , cling off by default
2) And if cling is made on .... you can `set(USE_CLING ON)` and `unset(USE_REPL)` or anything related in cmake.

So that if we encounter something like (the case you've not addressed yet)

https://github.com/compiler-research/CppInterOp/blob/f794dee0a636b8c958927aa527440474c28de5b8/.github/workflows/ci.yml#L873-L874

We can just use USE_CLING and we don't need to explicitly set REPL to off.

anutosh491: So we just end up with either the default case (which is supplying nothing, understood that repl is on and cling is off) or we just use cling=ON (understood repl would be off)

lib/Interpreter/Compatibility.h
```diff
@@ -183,7 +183,6 @@ inline void codeComplete(std::vector<std::string>& Results,
 
 #endif // USE_CLING
 
-#ifdef USE_REPL
```
vgvassilev, line 186: These checks should be turned into `#ifndef USE_CLING`.

vgvassilev: You should find all such cases and rework them just like this one.

mcbarton:
@faze-geek Once you get this working you will also need to update the documentation to match this change. This includes both within the docs folder as well as the README.md .

lib/Interpreter/Compatibility.h
anutosh491: Sure. Thanks !

faze-geek: I have followed this norm for all definitions now.

faze-geek:
> @faze-geek Once you get this working you will also need to update the documentation to match this change. This includes both within the docs folder as well as the README.md .

I have tested the pull request locally now. Once you guys are satisfied with the changes, I'll make the documentation change quick.

.github/workflows/ci.yml
```diff
@@ -1580,7 +1578,6 @@ jobs:
         else
           emcmake cmake -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}    \
                 -DUSE_CLING=OFF                             \
-                -DUSE_REPL=ON                               \
```
anutosh491, line 1583: please make the similar changes in this file too 

https://github.com/compiler-research/CppInterOp/blob/main/.github/workflows/deploy-pages.yml

.github/workflows/ci.yml
```diff
@@ -1191,7 +1191,6 @@ jobs:
         else
           cmake -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}     \
                 -DUSE_CLING=OFF                              \
```
mcbarton, line 1193: @faze-geek If use_repl has been made the default then we should be able to remove all the uses of use_cling=off, since it should be off by default.

.github/workflows/ci.yml
mcbarton: I suggest he goes through all workflows in the repo.

faze-geek: Thanks for letting me know.

mcbarton:
@faze-geek I just tried to rebase your PR, but it says it cannot be due to conflicts. Can you fix these conflicts and then rebase?

vgvassilev (approved):
Lgtm!

PR #362, opened 2024-12-02
Stale issue and PR workflow

mcbarton:
This PR makes it so issues and PRs which have had no activity for 90 days will be marked as stale, and will be closed 14 days after that if no further activity.


vgvassilev (approved):
Let’s see how that would work out.

PR #365, opened 2024-12-03
Minimal example of a bug from cppyy

Vipul-Cariappa:
That needs to be fixed at CppInterOp
reference: look at comments at compiler-research/cppyy-backend#116

---

I have written a test case that is a minimal example of a failing test at cppyy (suspected).
I am not sure how to fix this. More experienced people here could take a look at it.

### DIFF
--- a/unittests/CppInterOp/VariableReflectionTest.cpp
+++ b/unittests/CppInterOp/VariableReflectionTest.cpp
@@ -139,6 +139,39 @@ TEST(VariableReflectionTest, DatamembersWithAnonymousStructOrUnion) {
 #endif
 }
 
+TEST(VariableReflectionTest, GetTypeAsString) {
+  if (llvm::sys::RunningOnValgrind())
+    GTEST_SKIP() << "XFAIL due to Valgrind report";
+
+  std::string code = R"(
+  namespace my_namespace {
+  
+  struct Container {
+    int value;
+  };
+  
+  struct Wrapper {
+    Container item;
+  };
+
+  }
+  )";
+
+  Cpp::CreateInterpreter();
+  EXPECT_EQ(Cpp::Declare(code.c_str()), 0);
+
+  Cpp::TCppScope_t wrapper =
+      Cpp::GetScopeFromCompleteName("my_namespace::Wrapper");
+  EXPECT_TRUE(wrapper);
+
+  std::vector<Cpp::TCppScope_t> datamembers;
+  Cpp::GetDatamembers(wrapper, datamembers);
+  EXPECT_EQ(datamembers.size(), 1);
+
+  EXPECT_EQ(Cpp::GetTypeAsString(Cpp::GetVariableType(datamembers[0])),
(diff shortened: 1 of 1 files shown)


unittests/CppInterOp/VariableReflectionTest.cpp
```diff
@@ -139,6 +139,39 @@ TEST(VariableReflectionTest, DatamembersWithAnonymousStructOrUnion) {
...
+  Cpp::TCppScope_t wrapper =
+      Cpp::GetScopeFromCompleteName("my_namespace::Wrapper");
+  EXPECT_TRUE(wrapper);
+
+  std::vector<Cpp::TCppScope_t> datamembers;
+  Cpp::GetDatamembers(wrapper, datamembers);
+  EXPECT_EQ(datamembers.size(), 1);
+
+  EXPECT_EQ(Cpp::GetTypeAsString(Cpp::GetVariableType(datamembers[0])),
+            "my_namespace::Container");
```
Vipul-Cariappa, line 172: The assertion fails here with
```
"Container" != "my_namespace::Container"
```
I need the fully qualified type name.

vgvassilev: Wouldn't `GetQualifiedName` help?

mcbarton:
@Vipul-Cariappa has this bug been fixed now?

unittests/CppInterOp/VariableReflectionTest.cpp
faze-geek: Well the following diff in `GetVariableType` passed the above test for me. I may be wrong due to lack of context.

```
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -1216,7 +1216,10 @@ namespace Cpp {
     auto D = (Decl *) var;
 
     if (auto DD = llvm::dyn_cast_or_null<DeclaratorDecl>(D)) {
-      return DD->getType().getAsOpaquePtr();
+      QualType QT = DD->getType();
+
+      // Ensure the type includes the namespace.
+      QT = QT.getCanonicalType();
+      return QT.getAsOpaquePtr();
     }

```

vgvassilev: If the change makes the test pass that's a good sign.

faze-geek: Well, let me run all the tests and get back to you.

Vipul-Cariappa: @faze-geek, I did not test your changes locally. But if all existing tests and the test from this PR pass, you can open a PR of this branch (create a new branch from this branch). I will take a look at it. We may extend the test with other combinations.

Thanks for looking into this. 😄

faze-geek: Sure ! Will make the changes soon.

PR #367, opened 2024-12-03
Fix assert for external interpreter

aaronj0:


unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -200,7 +200,7 @@ TEST(InterpreterTest, ExternalInterpreterTest) {
 #endif
 
   EXPECT_NE(ExtInterp, nullptr);
-  Cpp::UseExternalInterpreter(ExtInterp);
+  ASSERT_DEATH(Cpp::UseExternalInterpreter(ExtInterp), "Existing Interpreter has been set");
```
vgvassilev, line 203:
```suggestion
#if defined(NDEBUG) || !GTEST_HAS_DEATH_TEST
  ASSERT_DEATH(Cpp::UseExternalInterpreter(ExtInterp), "Existing Interpreter has been set");
#endif
```

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -200,7 +203,10 @@ TEST(InterpreterTest, ExternalInterpreterTest) {
 #endif
 
   EXPECT_NE(ExtInterp, nullptr);
-  Cpp::UseExternalInterpreter(ExtInterp);
+
+#if defined(NDEBUG) || !GTEST_HAS_DEATH_TEST
```
vgvassilev, line 207: This needs to become `#if !defined(NDEBUG) || !GTEST_HAS_DEATH_TEST`

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -200,7 +203,13 @@ TEST(InterpreterTest, ExternalInterpreterTest) {
 #endif
 
   EXPECT_NE(ExtInterp, nullptr);
-  Cpp::UseExternalInterpreter(ExtInterp);
+
+#if !defined(NDEBUG) || !GTEST_HAS_DEATH_TEST
+  // death tests don't work right on windows
+  #ifndef _WIN32
```
vgvassilev, line 209: I thought this was covered by `GTEST_HAS_DEATH_TEST`. Did anything in particular fail?

aaronj0: Yes, it doesn't hit the assertion:
`  1: Actual msg:
  1: [  DEATH   ] Unable to find target for this triple (no targets are registered)
  `

aaronj0: Shouldn't we do `#if !defined(NDEBUG) && !GTEST_HAS_DEATH_TEST`?

vgvassilev: We do that, we do not need the `_WIN32` thing...

> Yes, it doesn't hit the assertion:
> 1: Actual msg: 1: [  DEATH   ] Unable to find target for this triple (no targets are registered)
 
 Seems a different problem...

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -200,7 +203,10 @@ TEST(InterpreterTest, ExternalInterpreterTest) {
 #endif
 
   EXPECT_NE(ExtInterp, nullptr);
-  Cpp::UseExternalInterpreter(ExtInterp);
+
+#if !defined(NDEBUG) && GTEST_HAS_DEATH_TEST
+    EXPECT_DEATH(Cpp::UseExternalInterpreter(ExtInterp), "sInterpreter already in use!");
```
vgvassilev, line 208:
```suggestion
#ifndef _WIN32 // Windows seems to fail to die...
    EXPECT_DEATH(Cpp::UseExternalInterpreter(ExtInterp), "sInterpreter already in use!");
#endif // _WIN32
```

vgvassilev (approved):
Lgtm!

PR #370, opened 2024-12-04
Update UsingCppInterOp.rst to not reference use of patches + small other updates

mcbarton:
This PR makes some small changes to UsingCppInterOp.rst to reflect the fact that CppInterOp now generates shared library on more than just Linux, and removes reference to patches, instead pointing user to installation instructions.


vgvassilev (approved):
Lgtm!

Issue #243, opened 2024-04-29
Doxygen documentation regeneration


vgvassilev:
Looks like our documentation does not pick up the doxygen style comments: https://cppinterop.readthedocs.io/en/latest/build/html/index.html

I am not sure if that's a feature or we are just not regenerating the documentation.

mcbarton:
I think this could be added as part of the nightly builds on the ci. We could add something that builds the documentation, and pushes the changes to this webpage.

vgvassilev:
> I think this could be added as part of the nightly builds on the ci. We could add something that builds the documentation, and pushes the changes to this webpage.

I believe that was part of the readthedocs website infrastructure. I just was not sure if the problem was the way we build our documentation or if we are not triggering new builds. 

In #246 we landed a new interface and it seems to be showing there which means we regenerate the documentation but we do not take the comments...

vgvassilev:
We seem to have hit some bug because all interface documentation is attached to the __attribute__() : https://cppinterop.readthedocs.io/en/latest/build/html/namespaceCpp.html#a3ef57f5fc1b6bf7e59bc40e340fd2be6

mcbarton:
This was fixed by https://github.com/compiler-research/CppInterOp/pull/381

PR #372, opened 2024-12-04
Try to fix developer documentation deployment

mcbarton:
@vgvassilev This PR may or may not provide a fix for https://github.com/compiler-research/CppInterOp/issues/243 . It fixes the link on my local machine, but that may not necessarily work when deployed.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,12 +5,12 @@ sphinx:
   builder: html
 
 build:
-  os: "ubuntu-22.04"
+  os: "ubuntu-24.04"
   tools:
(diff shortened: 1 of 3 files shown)


vgvassilev (approved):
Lgtm!

vgvassilev:
@mcbarton, it seems this broke doxygen completely: https://cppinterop.readthedocs.io/en/latest/build/html/index.html

mcbarton:
@vgvassilev I'll put in a PR to revert the changes to `.readthedocs.yaml` and `docs/conf.py` since they will be the things that will have caused the breaking.

PR #375, opened 2024-12-05
Use a Shared build for cppinterop while building for wasm

anutosh491:
Git diff can sometimes look to be pretty challenging to understand.

But what has been done is simply 

```
if(EMSCRIPTEN)
  // new logic for shared build
 else()
   // all logic on master moved here
endif()
```


anutosh491:
**Another major reason as to why I am not in favour of using different configs while building LLVM (clang, lld) for webassembly.**

Focus on the error here 
```
wasm-ld: error: unknown file type: /home/runner/work/CppInterOp/CppInterOp/llvm-project/build/lib/libclangInterpreter.a(DeviceOffload.cpp.o)
em++: error: '/home/runner/work/CppInterOp/CppInterOp/emsdk/upstream/bin/wasm-ld @/tmp/emscripten_y2obwybd.rsp.utf-8' failed (returned 1)
make[2]: *** [lib/Interpreter/CMakeFiles/clangCppInterOp.dir/build.make:245: lib/libclangCppInterOp.so.19.1] Error 1
make[1]: *** [CMakeFiles/Makefile2:351: lib/Interpreter/CMakeFiles/clangCppInterOp.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

What is happening is the default archiver being used is llvm-ar (which shouldn't be the case while building for webassembly using emscripten)

1) CppInterOp's CI is not using emcmake or emmake otherwise we would have seen this as a warning even before failing. (replicating the build locally but with emcmake you would see this)
```
em++: warning: /Users/anutosh491/work/llvm-project/build/lib/libclangInterpreter.a: archive is missing an index; Use emar when creating libraries to ensure an index is created [-Wemcc]
em++: warning: /Users/anutosh491/work/llvm-project/build/lib/libclangInterpreter.a: adding index [-Wemcc]
```
2) Now if you don't build using emar but using llvm-ar, the compiled objects are not what we want.
```
// so if we extract object files first
> emar x libclangInterpreter.a 

// then check file type
// the file type is wrong.
// expected is WebAssembly object file
> file DeviceOffload.cpp.o
DeviceOffload.cpp.o: Mach-O 64-bit object arm64
```
3) So basically while building llvm we need to supply the correct toolchain

i) Like what I did locally 
```
-DCMAKE_TOOLCHAIN_FILE=/Users/anutosh491/micromamba/envs/xeus-cpp-wasm-build/lib/python3.13/site-
packages/emsdk/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake
```
ii) What emscripten-forge does (https://github.com/emscripten-forge/recipes/actions/runs/12139306429/job/33846640952#step:9:216)

anutosh491:
Xeus-cpp's CI works perfectly now 
1) I made the exact same changes (added as a patch) on cppinterop's recipe on emscripten-forge to basically attempt a shared build (https://github.com/emscripten-forge/recipes/pull/1454)
2) Then used it in xeus-cpp's CI and as expected the undefined symbols occur no more (https://github.com/compiler-research/xeus-cpp/pull/190)

As I said above we should stick to 1 llvm build config here.

1) I would do it exactly as we did it for xeus-cpp.
2) We need something that sticks to emscripten stands (using emcmake, emmake and the emscripten toolchain) which is the whole point behind hosting stuff on emscripten-forge.
3) I would just remove `zlib` and add `llvm 19.1.5` here (https://github.com/compiler-research/CppInterOp/blob/main/environment-wasm.yml) and use it for the wasm build for cppinterop if given a choice.

mcbarton (approved):

PR #390, opened 2024-12-12
Add Emscripten Build Instructions

mcbarton:
# Description

This PR will add instructions on how to build a wasm version of CppInterOp and xeus-cpp.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/Emscripten-build-instructions.md
+++ b/Emscripten-build-instructions.md
@@ -99,7 +99,7 @@ emcmake cmake -DCMAKE_BUILD_TYPE=Release    \
                 -DCMAKE_INSTALL_PREFIX=$PREFIX         \
                 ../
 emmake make -j $(nproc --all)
-```bash
+```
 
 Once this finishes building we need to take note of where we built CppInterOp. This can be done by executing the following
 
--- a/Emscripten-build-instructions.md
+++ b/Emscripten-build-instructions.md
@@ -73,7 +73,6 @@ emcmake cmake -DCMAKE_BUILD_TYPE=Release \
                         -DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4" \
                         ../llvm
 emmake make clang -j $(nproc --all)
-emmake make clang-repl -j $(nproc --all)
 emmake make lld -j $(nproc --all)
 ```
 
@@ -83,8 +82,7 @@ Once this finishes building we need to take note of where we built our llvm buil
 export LLVM_BUILD_DIR=$PWD
 ```
 
-We can move onto building the wasm  
-version of CppInterOp. To do this execute the following  
+We can move onto building the wasm version of CppInterOp. To do this execute the following  
 
 ```bash
 cd ../../CppInterOp/
@@ -103,7 +101,7 @@ emcmake cmake -DCMAKE_BUILD_TYPE=Release    \
 emmake make -j $(nproc --all)
 ```bash
 
-Once this finishes building we need to take note of where we built our llvm build. This can be done by executing the following
+Once this finishes building we need to take note of where we built CppInterOp. This can be done by executing the following
 
 ```bash
 export CPPINTEROP_BUILD_DIR=$PWD
--- a/Emscripten-build-instructions.md
+++ b/Emscripten-build-instructions.md
@@ -1,0 +1,159 @@
+# Wasm Build Instructions
+
+It should be noted that the wasm build of CppInterOp is still experimental and subject to change to change.  
+
+## CppInterOp Wasm Build Instructions
+
+This document first starts with the instructions on how to build a wasm build of CppInterOp. Before we start it should be noted that  
+unlike the non wasm version of CppInterOp we currently only support the Clang-REPL backend using llvm>19 for osx and Linux.  
+We will first make folder to build our wasm build of CppInterOp. This can be done by executing the following command
+
+```bash
+mkdir CppInterOp-wasm
+```
+
+Now move into this directory using the following command
+
+```bash
+cd ./CppInterOp-wasm
+```
+
+To create a wasm build of CppInterOp we make use of the emsdk toolchain. This can be installed by executing (we only currently  
+support version 3.1.45)
+```bash
+git clone https://github.com/emscripten-core/emsdk.git
+cd emsdk
+./emsdk install  3.1.45
+```
+
+Go back upto the top level build directory using
+
+```bash
+cd ../..
+```
+
+and activate the emsdk environment
+
+```bash
+./emsdk/emsdk activate 3.1.45
+source ./emsdk/emsdk_env.sh
+```
+
+Now clone the 19.x release of the LLVM project repository and CppInterOp (the building of the emscripten version of llvm can be
+avoided by executing micromamba install llvm -c <https://repo.mamba.pm/emscripten-forge> and setting the LLVM_BUILD_DIR appropriately)
+
+
+```bash
+git clone --depth=1 --branch release/19.x https://github.com/llvm/llvm-project.git
+git clone --depth=1 https://github.com/compiler-research/CppInterOp.git
+```
+
+Now move into the cloned llvm-project folder and apply a required patch
+
+```bash
+cd ./llvm-project/
+git apply -v ../CppInterOp/patches/llvm/emscripten-clang${{ matrix.clang-runtime }}-*.patch
+```
+
+We are now in a position to build an emscripten build of llvm by executing the following
+```bash
+mkdir build
+cd build
+emcmake cmake -DCMAKE_BUILD_TYPE=Release \
+                        -DLLVM_HOST_TRIPLE=wasm32-unknown-emscripten \
+                        -DLLVM_ENABLE_ASSERTIONS=ON                        \
+                        -DLLVM_TARGETS_TO_BUILD="WebAssembly" \
+                        -DLLVM_ENABLE_LIBEDIT=OFF \
+                        -DLLVM_ENABLE_PROJECTS="clang;lld" \
+                        -DLLVM_ENABLE_ZSTD=OFF \
+                        -DLLVM_ENABLE_LIBXML2=OFF \
+                        -DCLANG_ENABLE_STATIC_ANALYZER=OFF \
+                        -DCLANG_ENABLE_ARCMT=OFF \
+                        -DCLANG_ENABLE_BOOTSTRAP=OFF \
+                        -DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4" \
+                        ../llvm
+emmake make clang -j $(nproc --all)
+emmake make clang-repl -j $(nproc --all)
+emmake make lld -j $(nproc --all)
+```
+
+Once this finishes building we need to take note of where we built our llvm build. This can be done by executing the following
+
+```bash
+export LLVM_BUILD_DIR=$PWD
+```
+
+We can move onto building the wasm  
+version of CppInterOp. To do this execute the following  
+
+```bash
+cd ../../CppInterOp/
+mkdir build
+cd ./build/
+emcmake cmake -DCMAKE_BUILD_TYPE=Release    \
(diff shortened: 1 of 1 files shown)


Emscripten-build-instructions.md
```diff
@@ -0,0 +1,167 @@
...
+python -m pip install jupyterlite-xeus
+jupyter lite build --XeusAddon.prefix=$PREFIX
+```
+
+Once the Jupyter Lite site has built you can test the website locally  
+by executing
+
+```bash
+jupyter lite serve --XeusAddon.prefix=$PREFIX
+```
```
anutosh491, line 167: I don't think there is any need to paste the xeus-cpp build again. Just provide a link from xeus-cpp's master for the same. This would ensure we have one place to go to when interested and the changes also need to be done at one place only.

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,167 @@
...
+
+```bash
+mkdir CppInterOp-wasm
+```
+
+Now move into this directory using the following command
+
+```bash
+cd ./CppInterOp-wasm
+```
```
anutosh491, line 20: Name for a build directory is weird. Just use build.

anutosh491: Also order looks weird 

1) setup emsdk first
2) any cloning that's required
3) create the build directory
4) The build configs using cmake
5) make step

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,167 @@
...
+emmake make clang-repl -j $(nproc --all)
+emmake make lld -j $(nproc --all)
+```
+
+Once this finishes building we need to take note of where we built our  
+llvm build. This can be done by executing the following
+
+```bash
+export LLVM_BUILD_DIR=$PWD
+```
```
anutosh491, line 86: This is okay but in the same place ....
**I would really like to make it as easy as possible for the user to try out**

Building LLVM against emscripten would itself take greater an 1 hour. 
Now that we're using the same configs as emscripten-forge, I would just do 
1) micromamba create an environment
2) micromamba install llvm -c https://repo.mamba.pm/emscripten-forge
3) use env prefix as `llvm_build_dir`, so basically LLVM_DIR = prefix/lib/cmake/llvm 

This would get the above task done in 2 mins. 

But as you've done ... for the curious user, we also provide a way to build llvm so that they can play around with configs if interested.

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,167 @@
...
+
+```bash
+cd ../../CppInterOp/
+mkdir build
+cd ./build/
+emcmake cmake -DCMAKE_BUILD_TYPE=Release    \
+                -DUSE_CLING=OFF                             \
+                -DUSE_REPL=ON                               \
+                -DLLVM_DIR=$LLVM_BUILD_DIR/lib/cmake/llvm      \
+                -DLLD_DIR=$LLVM_BUILD_DIR     \
```
anutosh491, line 99: lld should have its own dir `LLVM_BUILD_DIR/lib/cmake/lld`

Emscripten-build-instructions.md
anutosh491: We should just mention if cppinterop is built from source then we need to provide this config 
`-DCppInterOp_DIR="$CPPINTEROP_BUILD_DIR/lib/cmake/CppInterOp"  \`

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,167 @@
...
+                ../
+emmake make -j $(nproc --all)
+```bash
+
+Once this finishes building we need to take note of where we built our  
+llvm build. This can be done by executing the following
+
+```bash
+export CPPINTEROP_BUILD_DIR=$PWD
+```
```
anutosh491, line 113: Again .... we should provide a way to build cppinterop from source, but these would be the exact same configs on emscripten-forge, hence it would just be easy for anyone to fetch it from there, rather than building the whole toolchain.

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,167 @@
...
+```bash
+./emsdk/emsdk activate 3.1.45
+source ./emsdk/emsdk_env.sh
+```bash
+
+Now clone the 19.x release of the LLVM project repository 
+and CppInterOp
+
+```bash
+git clone --depth=1 --branch release/19.x https://github.com/llvm/llvm-project.git
```
anutosh491, line 47: image

This looks weird in preview.

anutosh491:
Overall formatting looks a bit weird, I don't know why that happens but everything seems left aligned.
Can't we have the docs to fit the page ?


image

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,167 @@
...
+                        -DLLVM_ENABLE_LIBXML2=OFF \
+                        -DCLANG_ENABLE_STATIC_ANALYZER=OFF \
+                        -DCLANG_ENABLE_ARCMT=OFF \
+                        -DCLANG_ENABLE_BOOTSTRAP=OFF \
+                        -DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4" \
+                        ../llvm
+emmake make clang -j $(nproc --all)
+emmake make clang-repl -j $(nproc --all)
+emmake make lld -j $(nproc --all)
+```
```
anutosh491, line 79: We have `cmake` in our `envoronment-wasm-build.yml` and I think we that should be setup and activate otherwise the emcmake cmake would fail !

mcbarton: It will not fail. This is evident through the ci since it just builds our llvm with the non conda provided cmake.We are going to assume that for these build instructions

mcbarton:
@anutosh491 I am aware if the weird formatting and will fix before it gets merged when we are happy with the contents

Emscripten-build-instructions.md
mcbarton: These instructions are here not just for those who want to fownload but also for those who want to help wuth development. I will add a link to the emscripten forge packages if you have one though

Emscripten-build-instructions.md
mcbarton: Will fix

Emscripten-build-instructions.md
mcbarton: We are not quite the same (not just yet anyway 😉). These are there to provide the imquistitive user too in case they want to play with some different flags. Gives them a starting point. These instructions are for developers as well as users

Emscripten-build-instructions.md
mcbarton: This is a top levell build directory to keep everything contained if the development build goes wrong. For those who simple want to experiment with it then I shalll point then tk the emscripten forge package

Emscripten-build-instructions.md
mcbarton: I plan to keep these instructions here rather than point to xeus-cpp instructions. This is because the instructions are slightly different if testing xeus-cpp against a CppInterOp built from source.

Emscripten-build-instructions.md
mcbarton: We don't really need envoronment-wasm-build.yml since we can use the systems conda. See llvm build in the ci which does just that.

anutosh491: Hmmm wait but we should atleast ask the user to get cmake from conda-forge using micromamba correct. 
I think you're right and the CI won't need it but if you do a micromamba install, I don't think you get cmake in your base env by default. 

Atleast I get a failure locally when I run a cmake command without having cmake installed

mcbarton: I'll add it for now as I'm not too fussed about this.

mcbarton: Just a note though that of course you'll get an error if you don't have cmake installed locally before executing the command.

alexander-penev (approved):

anutosh491:
I shall review this soon !

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,159 @@
...
+
+Once this finishes building we need to take note of where we built our llvm build. This can be done by executing the following
```
anutosh491, line 80: image

Formatting error !

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,159 @@
...
+                        -DLLVM_ENABLE_ZSTD=OFF \
+                        -DLLVM_ENABLE_LIBXML2=OFF \
+                        -DCLANG_ENABLE_STATIC_ANALYZER=OFF \
+                        -DCLANG_ENABLE_ARCMT=OFF \
+                        -DCLANG_ENABLE_BOOTSTRAP=OFF \
+                        -DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4" \
+                        ../llvm
+emmake make clang -j $(nproc --all)
+emmake make clang-repl -j $(nproc --all)
+emmake make lld -j $(nproc --all)
```
anutosh491, line 77: I am not fully convinced here (not sure it matters in the build) but 
1) if we use `LLVM_ENABLE_PROJECTS="clang;lld"` that means we want to build all static libraries in this cases coming out of clang and lld
2) Our focus is not to get clang-repl (the approach we use to get clang-repl working in the browser doesn't need `clang-repl.js` or `clang-repl.wasm` rather everything works through `libclangInterpreter.a`)
3) Even if we want to build clang-repl (which has dependency on lldwasm and lldcommon, I guess this would invoke "make lld" already, as required to get clang-repl ..... which in turn makes the below command redundant.

Basically what I am saying is we need a simple `emmake make -j $(nproc --all)` here and nothing else. I had raised a similar point for the CI not sure if that was addressed.

Emscripten-build-instructions.md
mcbarton: @anutosh491 Can you eloborate? I'm not  sure what the formstting error is.

Emscripten-build-instructions.md
```diff
@@ -0,0 +1,159 @@
...
+micromamba activate xeus-lite-host
+python -m pip install jupyterlite-xeus
+jupyter lite build --XeusAddon.prefix=$PREFIX
+```
+
+Once the Jupyter Lite site has built you can test the website locally  by executing
+
+```bash
+jupyter lite serve --XeusAddon.prefix=$PREFIX
+```
```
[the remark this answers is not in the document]
mcbarton: @anutosh491 I'll wait until you make the change to get xeus-cpp-lite working. I'll change accordingly if necessary.

Emscripten-build-instructions.md
vgvassilev: If we just need `libclangInterpreter.a` we should call that -- it should be a target : `make clangInterpreter` or something like that...

Emscripten-build-instructions.md
anutosh491: Open the preview and you would realized what I am talking about !

Emscripten-build-instructions.md
anutosh491: Ahhh sorry I meant, we need libclangIntepreter.a (and all its respective transitive dependencies like lldwasm, lldcommon from lld ,  clangast, clangbasic etc from clang and llvmsupport etc from llvm)

I think just doing `emmake make clang -j $(nproc --all)` would get the job done

Emscripten-build-instructions.md
mcbarton: @anutosh491 I think this is fixed now.

Emscripten-build-instructions.md
mcbarton: I think this is fixed now. I removed the clang-repl target, leaving just the clang and lld ones.

anutosh491 (approved):
xeus-cpp-lite build instructions need to be edited but otherwise looks good. Merging after CI passes.

PR #394, opened 2024-12-16
Update `GetBinaryOperator` to `GetOperator`

Vipul-Cariappa:
# Description

Resolving unary operators and operators defined within namespaces.
Required for `+` op in `std::string`.

## Fixes (issue)

Three issues at cppyy. Requires a patch for both CPyCppyy and cppyy-backend.

## Type of change

- [x] Bug fix
- [x] New feature

## Testing

I have added in tests cases for unary operator lookup and lookup of operator within a namespace.

## Checklist

### DIFF
--- a/include/clang/Interpreter/CppInterOp.h
+++ b/include/clang/Interpreter/CppInterOp.h
@@ -37,41 +37,56 @@ namespace Cpp {
   using TInterp_t = void*;
   using TCppObject_t = void*;
 
-  enum BinaryOperator {
-    PtrMemD = 0,
-    PtrMemI,
-    Mul,
-    Div,
-    Rem,
-    Add,
-    Sub,
-    Shl,
-    Shr,
-    Cmp,
-    LT,
-    GT,
-    LE,
-    GE,
-    EQ,
-    NE,
-    And,
-    Xor,
-    Or,
-    LAnd,
-    LOr,
-    Assign,
-    MulAssign,
-    DivAssign,
-    RemAssign,
-    AddAssign,
-    SubAssign,
-    ShlAssign,
-    ShrAssign,
-    AndAssign,
-    XorAssign,
-    OrAssign,
-    Comma,
+  enum Operator {
+    OP_None,
+    OP_New,
+    OP_Delete,
+    OP_Array_New,
+    OP_Array_Delete,
+    OP_Plus,
+    OP_Minus,
+    OP_Star,
+    OP_Slash,
+    OP_Percent,
+    OP_Caret,
+    OP_Amp,
+    OP_Pipe,
+    OP_Tilde,
+    OP_Exclaim,
+    OP_Equal,
+    OP_Less,
+    OP_Greater,
+    OP_PlusEqual,
+    OP_MinusEqual,
+    OP_StarEqual,
+    OP_SlashEqual,
+    OP_PercentEqual,
+    OP_CaretEqual,
+    OP_AmpEqual,
+    OP_PipeEqual,
+    OP_LessLess,
+    OP_GreaterGreater,
+    OP_LessLessEqual,
+    OP_GreaterGreaterEqual,
+    OP_EqualEqual,
+    OP_ExclaimEqual,
+    OP_LessEqual,
+    OP_GreaterEqual,
+    OP_Spaceship,
+    OP_AmpAmp,
+    OP_PipePipe,
+    OP_PlusPlus,
+    OP_MinusMinus,
+    OP_Comma,
+    OP_ArrowStar,
+    OP_Arrow,
+    OP_Call,
+    OP_Subscript,
+    OP_Conditional,
+    OP_Coawait,
   };
+
+  enum OperatorArity { kUnary = 1, kBinary, kBoth };
 
   /// A class modeling function calls for functions produced by the interpreter
   /// in compiled code. It provides an information if we are calling a standard
@@ -513,9 +528,13 @@ namespace Cpp {
   CPPINTEROP_API std::string GetFunctionArgName(TCppFunction_t func,
                                                 TCppIndex_t param_index);
 
-  ///\returns function that performs operation op on lc and rc
-  void GetBinaryOperator(TCppScope_t scope, enum BinaryOperator op,
-                         std::vector<TCppFunction_t>& operators);
+  ///\returns arity of the operator or kNone
+  OperatorArity GetOperatorArity(TCppFunction_t op);
+
+  ///\returns list of operator overloads
+  void GetOperator(TCppScope_t scope, Operator op,
+                   std::vector<TCppFunction_t>& operators,
+                   OperatorArity kind = kBoth);
 
   /// Creates an instance of the interpreter we need for the various interop
   /// services.
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -3315,22 +3315,43 @@ namespace Cpp {
     return PI->getNameAsString();
   }
 
-  void GetBinaryOperator(TCppScope_t scope, enum BinaryOperator op,
-                         std::vector<TCppFunction_t>& operators) {
+  OperatorArity GetOperatorArity(TCppFunction_t op) {
+    Decl* D = static_cast<Decl*>(op);
+    if (auto* FD = llvm::dyn_cast<FunctionDecl>(D)) {
+      if (FD->isOverloadedOperator()) {
+        switch (FD->getOverloadedOperator()) {
+#define OVERLOADED_OPERATOR(Name, Spelling, Token, Unary, Binary,            \
+                              MemberOnly)                                      \
+    case OO_##Name:                                                            \
+      if ((Unary) && (Binary))                                                 \
+        return kBoth;                                                          \
+      if (Unary)                                                               \
+        return kUnary;                                                         \
+      if (Binary)                                                              \
+        return kBinary;                                                        \
+      break;
+#include "clang/Basic/OperatorKinds.def"
+        default:
+          break;
+        }
+      }
+    }
+    return (OperatorArity)~0U;
+  }
+
+  void GetOperator(TCppScope_t scope, Operator op,
+                   std::vector<TCppFunction_t>& operators, OperatorArity kind) {
     Decl* D = static_cast<Decl*>(scope);
-    auto* DC = llvm::dyn_cast<DeclContext>(D);
-    Scope* S = getSema().getScopeForContext(DC);
-    if (!S)
-      return;
+    if (auto* DC = llvm::dyn_cast_or_null<DeclContext>(D)) {
+      ASTContext& C = getSema().getASTContext();
+      DeclContextLookupResult Result =
+          DC->lookup(C.DeclarationNames.getCXXOperatorName(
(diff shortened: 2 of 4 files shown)


include/clang/Interpreter/CppInterOp.h
```diff
@@ -512,9 +525,9 @@ namespace Cpp {
   CPPINTEROP_API std::string GetFunctionArgName(TCppFunction_t func,
                                                 TCppIndex_t param_index);
 
-  ///\returns function that performs operation op on lc and rc
-  void GetBinaryOperator(TCppScope_t scope, enum BinaryOperator op,
-                         std::vector<TCppFunction_t>& operators);
+  ///\returns list of operator overloads
+  void GetOperator(TCppScope_t scope, enum Operator op,
```
vgvassilev, line 529:
```suggestion
  void GetOperator(TCppScope_t scope, Operator op,
```

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3262,22 +3262,18 @@ namespace Cpp {
     return PI->getNameAsString();
   }
 
-  void GetBinaryOperator(TCppScope_t scope, enum BinaryOperator op,
-                         std::vector<TCppFunction_t>& operators) {
+  void GetOperator(TCppScope_t scope, enum Operator op,
```
vgvassilev, line 3265:
```suggestion
  void GetOperator(TCppScope_t scope, Operator op,
```

vgvassilev (commented):
Maybe address the clang-tidy issue.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3262,22 +3262,18 @@ namespace Cpp {
...
+      DeclContextLookupResult Result =
+          DC->lookup(C.DeclarationNames.getCXXOperatorName(
+              (clang::OverloadedOperatorKind)op));
 
-    for (NamedDecl* D : lookup) {
-      if (auto* FD = llvm::dyn_cast<Decl>(D))
-        operators.push_back(FD);
+      for (auto* i : Result) {
+        operators.push_back(i);
+      }
```
vgvassilev, line 3276:
```suggestion
      for (auto* i : Result)
        operators.push_back(i);
```

lib/Interpreter/CppInterOp.cpp
vgvassilev: And possibly add a `enum OperatorKind { kUnary, kBinary, kBoth}` as a third argument with possibly a `kBoth` as a default value.

include/clang/Interpreter/CppInterOp.h
```diff
@@ -37,40 +37,59 @@ namespace Cpp {
...
+    OP_Call,
+    OP_Subscript,
+    OP_Conditional,
+    OP_Coawait,
+  };
+
+  enum OperatorKind {
+    kUnary = 0b001,
+    kBinary = 0b010,
+    kBoth = 0b011,
```
vgvassilev, line 92: I guess we won't `or` these, 1,2,3 should be fine...

include/clang/Interpreter/CppInterOp.h
```diff
@@ -37,40 +37,59 @@ namespace Cpp {
...
+    OP_Comma,
+    OP_ArrowStar,
+    OP_Arrow,
+    OP_Call,
+    OP_Subscript,
+    OP_Conditional,
+    OP_Coawait,
+  };
+
+  enum OperatorKind {
```
vgvassilev, line 89:
```suggestion
  enum OperatorArity {
```

include/clang/Interpreter/CppInterOp.h
```diff
@@ -37,42 +37,57 @@ namespace Cpp {
...
+    OP_Comma,
+    OP_ArrowStar,
+    OP_Arrow,
+    OP_Call,
+    OP_Subscript,
+    OP_Conditional,
+    OP_Coawait,
   };
 
+  enum OperatorArity { kUnary = 1, kBinary, kBoth };
```
vgvassilev, line 89:
```suggestion
  enum OperatorArity { kBoth = 0, kUnary = 0, kBinary };
```

include/clang/Interpreter/CppInterOp.h
```diff
@@ -513,9 +528,16 @@ namespace Cpp {
   CPPINTEROP_API std::string GetFunctionArgName(TCppFunction_t func,
                                                 TCppIndex_t param_index);
 
-  ///\returns function that performs operation op on lc and rc
-  void GetBinaryOperator(TCppScope_t scope, enum BinaryOperator op,
-                         std::vector<TCppFunction_t>& operators);
+  ///\returns true if scope is a unary operator
+  bool IsUnaryOperator(TCppScope_t scope);
```
vgvassilev, line 532:
```suggestion
  bool IsUnaryOperator(TCppFunction_t op);
```

include/clang/Interpreter/CppInterOp.h
```diff
@@ -513,9 +528,16 @@ namespace Cpp {
...
+
+  ///\returns true if scope is a binary operator
+  bool IsBinaryOperator(TCppScope_t scope);
```
vgvassilev, line 535:
```suggestion
  bool IsBinaryOperator(TCppFunction_t op);
```

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3262,22 +3262,56 @@ namespace Cpp {
...
+  bool IsUnaryOperator(TCppScope_t scope) {
     Decl* D = static_cast<Decl*>(scope);
-    auto* DC = llvm::dyn_cast<DeclContext>(D);
-    Scope* S = getSema().getScopeForContext(DC);
-    if (!S)
-      return;
+    if (auto* FD = llvm::dyn_cast<FunctionDecl>(D)) {
+      if (FD->isOverloadedOperator()) {
+        switch (FD->getOverloadedOperator()) {
+#define OVERLOADED_OPERATOR(Name, Spelling, Token, Unary, Binary,            \
```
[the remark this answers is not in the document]
vgvassilev: That type of check should be disabled in our clang-tidy config..

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3307,22 +3307,56 @@ namespace Cpp {
...
+                              MemberOnly)                                      \
+    case OO_##Name:                                                            \
+      return Unary;
+#include "clang/Basic/OperatorKinds.def"
+        default:
+          break;
+        }
+      }
+    }
```
vgvassilev, line 3324: We should move that into a utility function, write the switch such that it returns 
```
OperatorArity getOperatorArity(...) {
  if (Unary && Binary) return kBoth; ...
}
```

include/clang/Interpreter/CppInterOp.h
Vipul-Cariappa: I did not understand this.
All Unary operators are not Binary operators. So `kBoth = 0` and `kUnary = 0` is confusing. Example `~` operator.

vgvassilev: Sorry, my bad, fixed the suggestion.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3307,22 +3307,42 @@ namespace Cpp {
...
+        return kUnary;                                                         \
+      if (Binary)                                                              \
+        return kBinary;
+#include "clang/Basic/OperatorKinds.def"
+        default:
+          break;
+        }
+      }
+    }
+    return kNone;
```
vgvassilev, line 3330: That should be unreachable, right? What does it mean `arity::kNone`? An operator without arguments? I do not think that's possible.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3307,22 +3307,42 @@ namespace Cpp {
...
+      ASTContext& C = getSema().getASTContext();
+      DeclContextLookupResult Result =
+          DC->lookup(C.DeclarationNames.getCXXOperatorName(
+              (clang::OverloadedOperatorKind)op));
 
-    for (NamedDecl* D : lookup) {
-      if (auto* FD = llvm::dyn_cast<Decl>(D))
-        operators.push_back(FD);
+      for (auto* i : Result) {
+        if (kind & getOperatorArity(i))
```
vgvassilev, line 3343:
```suggestion
        if (kind == getOperatorArity(i))
```

lib/Interpreter/CppInterOp.cpp
Vipul-Cariappa: Yes, should be unreachable.
It is for cases when the argument is something other than an operator or a function that is not an operator.

lib/Interpreter/CppInterOp.cpp
Vipul-Cariappa: When `kind = kBoth` should we return `Unary` $\cap$ `Binary` or `Unary` $\cup$ `Binary`?
This logic using `&` returns `Unary` $\cup$ `Binary`.
Whereas `==` will return   `Unary` $\cap$ `Binary`.

Operators like `-` is both `Unary` and `Binary`.

I set the `enum OperatorArity` values so that I can perform this bitwise operation.

vgvassilev: When I say kBoth, I expect all unary and binary operators to be present in that list.

lib/Interpreter/CppInterOp.cpp
vgvassilev: Shouldn't that be an assert or an early exit?

lib/Interpreter/CppInterOp.cpp
Vipul-Cariappa: In that case, we should be using `&`. `kUnary = 0b001` and `kBinary = 0b010` and `kBoth = kUnary ^ kBinary = 0b011`.
No changes are required here.

lib/Interpreter/CppInterOp.cpp
Vipul-Cariappa: I will say this is simpler. Assert will crash the program. In this case, the user can handle `kNone` as an error.

Also if I remove the return, clang-tidy will get angry.

Let me know if you still think we should remove `kNone`. I can do that, but not sure how I would satisfy clang-tidy's message of `Non-void function does not return a value in all control paths`.

vgvassilev: You should remove it. You can return `~0U`

Vipul-Cariappa: Compiler error:
```bash
/home/vipul-cariappa/Workspace/cpp-py/compiler-research/CppInterOp/lib/Interpreter/CppInterOp.cpp:3339:12: error: invalid conversion from ‘unsigned int’ to ‘Cpp::OperatorArity’ [-fpermissive]
 3339 |     return ~0U;
      |            ^~~
      |            |
      |            unsigned int
```

Should I rename it to `kInvalid` instead?

vgvassilev:
Do we know why the check if failing?

aaronj0:
> Do we know why the check if failing?

Looks like the API update in cppyy-backend was not pulled, rerunning now. should be green

PR #396, opened 2024-12-17
fix `GetVariableOffset` in cases of multi-level and multiple inheritance

Vipul-Cariappa:
# Description

Fixes the calculation of data member offsets of a class C. Where C inherits from multiple classes and these classes themselves inherit from other classes.

## Fixes # (issue)

I found this issue when tinkering with cross-inheritance in cppyy.
It looks like upstream cppyy also faces a similar problem. Opened a PR stating the issue at https://github.com/wlav/cppyy/issues/280.

> Does not contribute to test cases that pass in cppyy.

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

Added in tests with multi-level and multiple inheritance.

## Checklist

### DIFF
--- a/include/clang/Interpreter/CppInterOp.h
+++ b/include/clang/Interpreter/CppInterOp.h
@@ -446,7 +446,8 @@ namespace Cpp {
 
   /// Gets the address of the variable, you can use it to get the
   /// value stored in the variable.
-  CPPINTEROP_API intptr_t GetVariableOffset(TCppScope_t var);
+  CPPINTEROP_API intptr_t GetVariableOffset(TCppScope_t var,
+                                            TCppScope_t parent = nullptr);
 
   /// Checks if the provided variable is a 'Public' variable.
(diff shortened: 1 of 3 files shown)


vgvassilev:
We should probably turn off clang tidy from the unittests.

Vipul-Cariappa:
Looks like, if we include a virtual destructor in the base classes these changes are not required.

vgvassilev (approved):
LGTM, just make clang-tidy happy about this missing include.

mcbarton:
> We should probably turn off clang tidy from the unittests.

@vgvassilev @Vipul-Cariappa If you want clang tidy to not provide comments on unit tests then you'll want to change https://github.com/compiler-research/CppInterOp/blob/c71069fc87d8a1fc758803d7382b58532c9288ef/.github/workflows/clang-format.yml#L6 to only look in the lib and include folders.

PR #420, opened 2025-01-08
Add correct options to build xeus-cpp-lite deployment correctly

mcbarton:
# Description

This PR should fix the emscripten build in the ci. The new llvm options may not be necessary. They are just used in emscripten-forge so I decided to add here for consistency. The LDFLAGS definitions are needed otherwise the CppInterOp shared library is not built correctly. I have cleared the cache of one build to check I have not broken the build there. Once it passes this PR should be ready for merging.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -92,7 +92,7 @@ jobs:
       shell: bash -l {0}
       run: |
         ./emsdk/emsdk activate ${{matrix.emsdk_ver}}
-         source ./emsdk/emsdk_env.sh
+        source ./emsdk/emsdk_env.sh
         micromamba create -f environment-wasm.yml --platform=emscripten-wasm32
 
         export PREFIX=$MAMBA_ROOT_PREFIX/envs/CppInterOp-wasm
@@ -142,7 +142,7 @@ jobs:
                 ../
         fi
         
-        emmake make -j ${{ env.ncpus }}
+        emmake make -j ${{ env.ncpus }} install
 
         cd ..
         
@@ -181,12 +181,12 @@ jobs:
       shell: bash -l {0}
       run: |
           cd ./xeus-cpp/
-          micromamba create -n xeus-lite-host jupyterlite-core
+          micromamba create -n xeus-lite-host jupyterlite-core -c conda-forge
           micromamba activate xeus-lite-host
           python -m pip install jupyterlite-xeus
           jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }} --output-dir dist
           cp ${{ env.PREFIX }}/bin/xcpp.data dist/extensions/@jupyterlite/xeus/static
-          cp ${{ env.CPPINTEROP_BUILD_DIR }}/lib/libclangCppInterOp.so dist/extensions/@jupyterlite/xeus/static
+          cp ${{ env.PREFIX }}/lib/libclangCppInterOp.so dist/extensions/@jupyterlite/xeus/static
 
     - name: Upload artifact
       uses: actions/upload-pages-artifact@v3
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -541,7 +541,7 @@ jobs:
       shell: bash -l {0}
       run: |
         ./emsdk/emsdk activate ${{matrix.emsdk_ver}}
-         source ./emsdk/emsdk_env.sh
+        source ./emsdk/emsdk_env.sh
         micromamba create -f environment-wasm.yml --platform=emscripten-wasm32
 
         export PREFIX=$MAMBA_ROOT_PREFIX/envs/CppInterOp-wasm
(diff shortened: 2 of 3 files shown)


.github/workflows/deploy-pages.yml
```diff
@@ -92,12 +92,13 @@ jobs:
...
       run: |
         ./emsdk/emsdk activate ${{matrix.emsdk_ver}}
-         source ./emsdk/emsdk_env.sh
+        source ./emsdk/emsdk_env.sh
         micromamba create -f environment-wasm.yml --platform=emscripten-wasm32
 
         export PREFIX=$MAMBA_ROOT_PREFIX/envs/CppInterOp-wasm
         export CMAKE_PREFIX_PATH=$PREFIX
         export CMAKE_SYSTEM_PREFIX_PATH=$PREFIX
+        export LDFLAGS="-s MODULARIZE=1 -s LINKABLE=1 -s EXPORT_ALL=1 -s WASM=1 -std=c++14 -s SIDE_MODULE=1 -sWASM_BIGINT"
```
anutosh491, line 101: Could you maybe quickly point me to the source of this (probably somewhere on emscripten-forge) ? Probably would be nice to know how each flag contributes, hence asking for the source.

Also Not sure I am a big fan of exporting flags in the CI anyways. 
Can we do them here if that's a possiblity
https://github.com/compiler-research/CppInterOp/blob/c446c3e2c43a1d3f374c739295d1484d5349274e/lib/Interpreter/CMakeLists.txt#L1-L20

We can set them as required for the emscripten build.

mcbarton: Here is where I think they come from in emscripten-forge. When you set up emsdk https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes/emscripten_emscripten-wasm32/activate.sh .

mcbarton: Addressed.

lib/Interpreter/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 if(EMSCRIPTEN)
   set_property(GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS TRUE)
+  set(LDFLAGS "-s MODULARIZE=1 -s LINKABLE=1 -s EXPORT_ALL=1 -s WASM=1 -std=c++14 -s SIDE_MODULE=1 -sWASM_BIGINT")
```
anutosh491, line 3: Get rid of `LDFLAGS` here and add the following after the add_llvm_library function. 
```
  target_link_options(clangCppInterOp PRIVATE
    PUBLIC "SHELL: -s WASM_BIGINT"
  )
```
As we are in cmake, we should prefer using `target_link_options` like what xeus-cpp/xeus does to set flags rather than setting them like this (probably this way can be preferred while running a build.sh script or in the CI)

.github/workflows/emscripten.yml
```diff
@@ -302,6 +302,8 @@ jobs:
                         -DLLVM_INCLUDE_EXAMPLES=OFF                     \
                         -DLLVM_INCLUDE_TESTS=OFF                        \
                         -DLLVM_ENABLE_THREADS=OFF                       \
+                        -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON          \
```
anutosh491, line 305: Also I am severely confused with respect to `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE`. As far as my understanding goes there are just 3 values this can take (ONLY, NEVER and BOTH) 

https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_PACKAGE.html

I don't know what is meant by "ON"

Actually I see it being used so many times in our source code (https://github.com/search?q=repo%3Acompiler-research%2FCppInterOp%20CMAKE_FIND_ROOT_PATH_MODE_PACKAGE&type=code)

I am not sure if this makes sense. We could just get rid of this and use BOTH which is the default value for this !

.github/workflows/emscripten.yml
```diff
@@ -302,6 +302,8 @@ jobs:
...
+                        -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=true     \
```
anutosh491, line 306: I don't see a reason why this is needed. We need position independent code when creating a shared object or a module.

In case of llvm we are interested in coming up with static libs (libclangInterpreter.a) and cmake prefers would use false by default for such cases. This would not be needed.

.github/workflows/emscripten.yml
anutosh491: Let us get rid of this for now !

.github/workflows/emscripten.yml
mcbarton: Done.

.github/workflows/emscripten.yml
mcbarton: Done.

lib/Interpreter/CMakeLists.txt
mcbarton: Done.

anutosh491 (approved):
Looks good. Thanks !

PR #423, opened 2025-01-08
Use Ninja to build llvm in the ci (osx and Ubuntu)

mcbarton:
# Description

My local testing on an osx arm Macbook found an 35% reduction in the size of the build folder when using Ninja to build llvm. This should help reduce the pressure on the cache. Supposedly Ninja is faster so we should see a speed in the llvm build when we have to rebuild the cache. If the ci passes then all Ubuntu and MacOS builds should be cleared from the cache before merging.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/MacOS.yml
+++ b/.github/workflows/MacOS.yml
@@ -206,6 +206,7 @@ jobs:
         done
         brew upgrade openssl >/dev/null 2>&1 
         brew upgrade
+        brew install ninja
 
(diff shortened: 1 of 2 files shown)


mcbarton:
@anutosh491 Ninja works now. I have tried it and plan to put in another PR for the emscripten build once the deployment PR is merged.

PR #424, opened 2025-01-09
Replace std::regex in codecomplete to use llvm::regex

aaronj0:
@alexander-penev  We ran into an issue with torchlib that is triggered by the loading of `std::regex`. I have replaced the usage with `llvm::Regex` that fi xes it. Can you verify if this works with your usecases when developing the feature? The unittests for codecomplete pass but I am not sure if it is extensive enough.
.

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/lib/Interpreter/Compatibility.h
+++ b/lib/Interpreter/Compatibility.h
@@ -78,6 +78,9 @@ static inline char* GetEnv(const char* Var_Name) {
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/Path.h"
 
+// std::regex breaks pytorch's jit: pytorch/pytorch#49460
+#include "llvm/Support/Regex.h"
+
 #ifdef USE_CLING
 
 #include "cling/Interpreter/DynamicLibraryManager.h"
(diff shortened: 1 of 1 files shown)


lib/Interpreter/Compatibility.h
```diff
@@ -77,6 +77,7 @@ static inline char* GetEnv(const char* Var_Name) {
 #include "llvm/ExecutionEngine/Orc/LLJIT.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/Path.h"
+#include "llvm/Support/Regex.h"
```
vgvassilev, line 80:
```suggestion
// std::regex breaks pytorch's jit: pytorch/pytorch#49460
#include "llvm/Support/Regex.h"
```

lib/Interpreter/Compatibility.h
```diff
@@ -163,17 +166,45 @@ inline void codeComplete(std::vector<std::string>& Results,
   std::vector<std::string> results;
   size_t column = complete_column;
   I.codeComplete(code, column, results);
+  std::string error;
+  llvm::Error Err;
```
vgvassilev, line 170:
```suggestion
  llvm::Error Err = llvm::Error::success();
```

vgvassilev (approved):
LGTM!

PR #425, opened 2025-01-10
Export __clang_Interpreter_SetValueNoAlloc to allow automatic printf in xeus-cpp-lite

anutosh491:
# Description

Check 
1) https://github.com/emscripten-forge/recipes/pull/1515
2) https://github.com/compiler-research/xeus-cpp/issues/202
3) https://github.com/compiler-research/xeus-cpp/pull/203

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/lib/Interpreter/CMakeLists.txt
+++ b/lib/Interpreter/CMakeLists.txt
@@ -19,6 +19,7 @@ if(EMSCRIPTEN)
   )
   target_link_options(clangCppInterOp PRIVATE
     PUBLIC "SHELL: -s WASM_BIGINT"
+    PUBLIC "SHELL: -Wl,--export=__clang_Interpreter_SetValueNoAlloc"
   )
 else()
   set(LLVM_LINK_COMPONENTS


anutosh491:
Actually this can be seen in the two links 

1) Cppinterop:  https://compiler-research.org/CppInterOp/lab/index.html

2) xeus-cpp: https://compiler-research.github.io/xeus-cpp/lab/index.html

Try automatic printf (maybe type 10 and execute the shell )

it would error out for cppinterop but not for xeus-cpp (now an output is not expected cause its not yet implemented https://github.com/llvm/llvm-project/blob/73dd730fb9403ca648a46b489bf04e27b2a93840/clang/lib/Interpreter/Value.cpp#L258-L266 but whenever it would be we are ready to support it and this would be required.)

vgvassilev (commented):
Can we add a test demonstrating this now works?

anutosh491:
> Can we add a test demonstrating this now works?

Well unfortunately this is one thing I am not well versed with :(

Never written any concrete "tests' as such for an emscripten use case and also I don't think any other xeus-lite kernels support testing right now.

I shall raise this point in front of xeus-lite maintainers today and see if there is something we can do !

vgvassilev:
We could add a unit test that uses this symbol and it will fail as part of the ci, right?

anutosh491:
Sorry for not being able to get back here. Not sure I know how to test this well enough as of now and also focusing more on the implementation part than testing these days. There is some kind of testing framework (as in cell by cell) in jupyterlite-xeus which is already being put to use for xeus-python and we probably need to introduce the same for xeus-cpp.

That being said although I don't know how to test this through the Ci right away, I know for sure that we need it.

If you open cppinterop's lite deployment, you shall find that **we fail to compute the C++ version being supported**

image

This won't be seen in xeus-cpp's lite deployment cause the same change is added through a patch on cppinterop's recipe on emscripten-forge.

anutosh491:
Can be closed as superseded through #447

PR #438, opened 2025-01-12
Remove valgrind install emscripten.yml

mcbarton:
# Description

Please include a summary of changes, motivation and context for this PR.

When doing the emscripten llvm build in the ci we have a random valgrind install as a dependency. This can be removed.

Fixes # (issue)

## Type of change

Please tick all options which are relevant.

- [x] Bug fix
- [ ] New feature
- [ ] Requires documentation updates

## Testing

Please describe the test(s) that you added and ran to verify your changes.

## Checklist

- [x] I have read the contribution guide recently

PR #440, opened 2025-01-12
Fix symbol visibility issues for C API

Gnimuc:
Add the missing `CINDEX_LINKAGE` macros for C APIs

### DIFF
--- a/lib/Interpreter/CMakeLists.txt
+++ b/lib/Interpreter/CMakeLists.txt
@@ -133,6 +133,8 @@ else()
     LINK_LIBS
     ${link_libs}
   )
+
+  target_compile_definitions(clangCppInterOp PUBLIC "_CINDEX_LIB_") # workaround for the use of `CINDEX_LINKAGE`
 endif()
 
 string(REPLACE ";" "\;" _VER CPPINTEROP_VERSION)
--- a/include/clang-c/CXCppInterOp.h
+++ b/include/clang-c/CXCppInterOp.h
@@ -34,7 +34,8 @@ typedef struct CXInterpreterImpl* CXInterpreter;
  *
  * \returns a \c CXInterpreter.
  */
-CXInterpreter clang_createInterpreter(const char* const* argv, int argc);
+CINDEX_LINKAGE CXInterpreter clang_createInterpreter(const char* const* argv,
+                                                     int argc);
 
 typedef void* TInterp_t;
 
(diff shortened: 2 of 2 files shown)


mcbarton:
@Gnimuc Your PR which added the cpi was largely untested (see https://github.com/compiler-research/CppInterOp/pull/438#issuecomment-2439894848). Would you mind putting in a PR which tests your C API sp can get our total test coverage back up?

Gnimuc:
> @Gnimuc Your PR which added the cpi was largely untested (see [#438 (comment)](https://github.com/compiler-research/CppInterOp/pull/438#issuecomment-2439894848)). Would you mind putting in a PR which tests your C API so can get our total test coverage back up?

@mcbarton

Half of the untested lines are copied from libclang:

https://app.codecov.io/gh/compiler-research/CppInterOp/blob/main/lib%2FInterpreter%2FCXCppInterOp.cpp#L156

Since those C interfaces are just thin wrappers, I think a more practical way to improve the coverage of the project is adding `CXCppInterOp.cpp` to the ignore path of CodeCov.

mcbarton:
> > @Gnimuc Your PR which added the cpi was largely untested (see [#438 (comment)](https://github.com/compiler-research/CppInterOp/pull/438#issuecomment-2439894848)). Would you mind putting in a PR which tests your C API so can get our total test coverage back up?
> 
> @mcbarton
> 
> Half of the untested lines are copied from libclang:
> 
> https://app.codecov.io/gh/compiler-research/CppInterOp/blob/main/lib%2FInterpreter%2FCXCppInterOp.cpp#L156
> 
> Since those C interfaces are just thin wrappers, I think a more practical way to improve the coverage of the project is adding `CXCppInterOp.cpp` to the ignore path of CodeCov.

@vgvassilev What do you think of @Gnimuc suggestion?

vgvassilev:
I don't have a strong opinion. Generally it is good to have tests especially for the things that are not in libclang.

mcbarton:
@Gnimuc I gave it some thought and think it would still be better to have the tests. Hopefully given they are just thin wrappers then I believe it shouldn't be too much to get tests written with them too. Then we can guarantee everything is working as expected.

Gnimuc:
@vgvassilev  Is it OK to turn off warning C4273 as error?

vgvassilev (approved):
Lgtm!

vgvassilev:
> @vgvassilev Is it OK to turn off warning C4273 as error?

Do we understand why the warning fires and why it is not relevant?

Gnimuc:
> > @vgvassilev Is it OK to turn off warning C4273 as error?
> 
> Do we understand why the warning fires and why it is not relevant?

It turns out that the `CINDEX_LINKAGE` is not correctly triggered in the CMake configuration. I added a small fix to workaround the problem.

lib/Interpreter/CMakeLists.txt
```diff
@@ -133,6 +133,8 @@ else()
     LINK_LIBS
     ${link_libs}
   )
+
+  target_compile_definitions(clangCppInterOp PUBLIC "_CINDEX_LIB_") # workaround for the use of `CINDEX_LINKAGE`
```
vgvassilev, line 137: How about using `CLANG_ABI` from `clang/Support/Compiler.h`?

PR #455, opened 2025-01-16
[cmake] Perform symbol exports depending on winx86 and winx64

aaronj0:
Fixes issues building libCling and unit tests on ROOT for winx86


vgvassilev (approved):
LGTM!

PR #456, opened 2025-01-16
Upgrade to emsdk to version 3.1.73

mcbarton:
# Description

This is a PR to upgrade the emsdk version we use to version 3.1.73, once emscripten forge is finished with its transition to this version.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -27,7 +27,7 @@ jobs:
             clang-runtime: '19'
             cling: Off
             micromamba_shell_init: bash
-            emsdk_ver: "3.1.45"
+            emsdk_ver: "3.1.73"
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -28,7 +28,7 @@ jobs:
             cling: Off
             llvm_enable_projects: "clang;lld"
             llvm_targets_to_build: "WebAssembly"
-            emsdk_ver: "3.1.45"
+            emsdk_ver: "3.1.73"
           - name: osx15-arm-clang-clang-repl-19-emscripten
             os: macos-15
             compiler: clang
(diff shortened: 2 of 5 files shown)


anutosh491:
I would prefer we use the patched version of emsdk that both pyodide and emscripten-forge use. They use the same patch and to my knowledge it is better for building shared libraries. This will ensure consistency with emscripten-forge too !

See https://github.com/jupyter-xeus/xeus-r/pull/116 on how to do it

anutosh491:
Hi, 

Could you please finish this PR ?
Cause once done we can try moving to static build for cppinterop !

anutosh491:
Was going through the PR. There are some changes needed

1) As decided we should prefer linux on all the wasm fronts (so https://github.com/compiler-research/CppInterOp/pull/456/files#diff-0f132b0962009071404990a6789b3466dbd8723d1f7b744f6aca0312396a7900R25 needs to be changed)
2) Also we would be using the patched emsdk either from pyodide or emscripten-forge hence https://github.com/compiler-research/CppInterOp/pull/456/files#diff-0f132b0962009071404990a6789b3466dbd8723d1f7b744f6aca0312396a7900R76-R81 should be changed (no setup required. Use it right from the env. Check https://github.com/compiler-research/xeus-cpp/blob/main/.github/workflows/deploy-github-page.yml)

.github/workflows/deploy-pages.yml
```diff
@@ -27,7 +27,7 @@ jobs:
             clang-runtime: '19'
             cling: Off
             micromamba_shell_init: bash
-            emsdk_ver: "3.1.45"
+            emsdk_ver: "3.1.73"
```
anutosh491, line 30: This can completely be removed once the above comment is addressed

anutosh491: Similarly for docs and the main.yml. We wouldn't need to store this in a variable !

anutosh491:
@mcbarton why was this merged ?
I don't see any of my comments being addressed!

anutosh491:
?

PR #459, opened 2025-01-18
Minor patches to unittests, improve offset initialization in getExecutionEngine

aaronj0:
# Description

Set of patches required on top of InterOp master for enabling ROOT, which can added upstream.


vgvassilev (approved):
Please make clang-format happy

PR #463, opened 2025-01-18
Add Ubuntu 24.04 arm runners to ci

mcbarton:
# Description

I wanted to do this PR when we added the new Ubuntu 22.04 arm runners to the ci, but we didn't have the cache space. Also CppInterOp doesn't current support Ubuntu 24.04 x86 due to an unfixed warning, so its not guaranteed to pass.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

## Testing

## Checklist


lib/Interpreter/CppInterOp.cpp
```diff
@@ -2828,7 +2828,14 @@ namespace Cpp {
 #define DEBUG_TYPE "exec"
 
     std::array<char, 256> buffer;
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wignored-attributes"
+#endif
     std::unique_ptr<FILE, decltype(&pclose)> pipe(popen(cmd, "r"), pclose);
+#if defined(__GNUC__)
```
vgvassilev, line 2836: What is the diagnostic we get here?

mcbarton: https://github.com/compiler-research/CppInterOp/actions/runs/12851542774/job/35832389048?pr=463#step:10:187 I reverted the change so you can see the warning here. I've tried in the past to fix this warning without success. Happy to take suggestions if you have ideas.

vgvassilev: Does that help:

https://stackoverflow.com/questions/76867698/what-does-ignoring-attributes-on-template-argument-mean-in-this-context

mcbarton: @vgvassilev No it doesn't, as that is the reference I used when trying to fix it previously without success.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -2828,7 +2828,11 @@ namespace Cpp {
 #define DEBUG_TYPE "exec"
 
     std::array<char, 256> buffer;
-    std::unique_ptr<FILE, decltype(&pclose)> pipe(popen(cmd, "r"), pclose);
+    struct file_deleter {
+      void operator()(FILE* fp) { pclose(fp); }
+    };
+    using file_pointer = std::unique_ptr<FILE, file_deleter>;
+    file_pointer pipe{popen(cmd, "r")};
```
vgvassilev, line 2835:
```suggestion
    std::unique_ptr<FILE, file_deleter> pipe{popen(cmd, "r")};
```

lib/Interpreter/CppInterOp.cpp
```diff
@@ -3437,26 +3441,28 @@ namespace Cpp {
   }
 
   class StreamCaptureInfo {
-    std::unique_ptr<FILE, decltype(std::fclose)*> m_TempFile;
+    struct file_deleter {
+      void operator()(FILE* fp) { pclose(fp); }
+    };
+    using file_pointer = std::unique_ptr<FILE, file_deleter>;
+    file_pointer m_TempFile;
```
vgvassilev, line 3448:
```suggestion
    std::unique_ptr<FILE, file_deleter> m_TempFile;
```

PR #483, opened 2025-01-24
Add emscripten automated tests

mcbarton:
# Description

My first PR to add automated emscripten tests became broken for unknown reasons (see https://github.com/compiler-research/CppInterOp/pull/448). This PR is another attempt to added emscripten tests to CppInterOp

Fixes # (issue)

## Type of change

- [x] New feature

## Testing

## Checklist


unittests/CppInterOp/main.cpp
```diff
@@ -0,0 +1,6 @@
+#include <gtest/gtest.h>
```
vgvassilev, line 1: Why do we need this file?

vgvassilev (commented):
I am adding a few comments from my partial review.

unittests/CppInterOp/main.cpp
mcbarton: Basically for gtest to work for emscripten builds, you need a main defined througha script. It doesn't work if you simply link gtest_main like you do with native builds.

lib/Interpreter/CMakeLists.txt
```diff
@@ -133,6 +113,28 @@ else()
     LINK_LIBS
     ${link_libs}
   )
+
+if(EMSCRIPTEN)
+  # Read the undefined symbols from the text file
+  file(READ "${CMAKE_SOURCE_DIR}/lib/Interpreter/undefined_symbols.txt" SYMBOLS_LIST)
```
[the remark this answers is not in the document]
mcbarton: I'm not aware of linker scripts. Can you point me a resource about them? I'll remame the file as suggested.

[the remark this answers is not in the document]
vgvassilev: https://users.informatik.haw-hamburg.de/~krabat/FH-Labor/gnupro/5_GNUPro_Utilities/c_Using_LD/ldLinker_scripts.html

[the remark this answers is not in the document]
mcbarton: Addressed point 2 of renaming the file. Point 1 will take some more investigation on my part. My initial attempts to use a linker script ended in failure. Apart from this point I believe this PR is ready for another review when your available.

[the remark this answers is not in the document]
mcbarton: @vgvassilev I got confirmation from the Emscripten maintainers here that linker scripts are not supported by either Emscripten or wasm-ld https://github.com/emscripten-core/emscripten/issues/23637 . He did suggest if possible to use EXPORTED_FUNCTIONS Emscripten flag instead of -Wl,export , but in our case because we are using side_module=1 to get our shared library this option is not valid (you literally get a warning that this flag is not valid if you try and include it). Can I mark this as resolved, and stick with the current working solution?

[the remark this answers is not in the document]
vgvassilev: Sure.

cmake/modules/GoogleTest.cmake
```diff
@@ -13,41 +13,49 @@ if(WIN32)
...
+    CONFIGURE_COMMAND ${CONFIGURE_COMMAND}
+    BUILD_COMMAND ${BUILD_COMMAND}
+    # Disable install step
+    INSTALL_COMMAND ""
+    BUILD_BYPRODUCTS ${_gtest_byproducts}
+    # Wrap download, configure and build steps in a script to log output
+    LOG_DOWNLOAD ON
+    LOG_CONFIGURE ON
+    LOG_BUILD ON
+    TIMEOUT 600
```
vgvassilev, line 58: Here the diff should be minimal. We should only change the values of the commands that are different into separate variables outside. Eg.
```cmake
if (emscripten)
  set(config_cmd "CONFIGURE_COMMAND emcmake cmake") # if more than one could be accumulated in `extra_cmds`
endif(emscripten)
```
Then we just add this variable to the overall `ExternalProject_Add`...

cmake/modules/GoogleTest.cmake
```diff
@@ -13,41 +13,47 @@ if(WIN32)
...
-  TIMEOUT 600
+if (EMSCRIPTEN)
+  set(config_cmd emcmake cmake)
+  set(build_cmd emmake make)
+else()
+  set(config_cmd ${CMAKE_COMMAND})
+  set(build_cmd ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR}/unittests/googletest-prefix/src/googletest-build/ --config $<CONFIG>)
+endif()
+
+  ExternalProject_Add(
```
vgvassilev, line 30: Please fix the indentation as it were before because it confuses the diff...

cmake/modules/GoogleTest.cmake
mcbarton: Done.

cmake/modules/GoogleTest.cmake
mcbarton: Done.

unittests/CppInterOp/TestSharedLib/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+if(EMSCRIPTEN)
+# Without this it tries compiling to static library and get a warning
+set_property(GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS TRUE)
+set(CMAKE_STRIP FALSE)
+endif()
```
vgvassilev, line 5:
```suggestion
if (EMSCRIPTEN)
  # Without this it tries compiling to static library and get a warning
  set_property(GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS TRUE)
  set(CMAKE_STRIP FALSE)
endif()
```

unittests/CppInterOp/TestSharedLib/CMakeLists.txt
```diff
@@ -8,4 +14,15 @@ set_output_directory(TestSharedLib
   BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/unittests/bin/$<CONFIG>/
   LIBRARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/unittests/bin/$<CONFIG>/
   )
+
+
+if(EMSCRIPTEN)
```
vgvassilev, line 19:
```suggestion
if (EMSCRIPTEN)
```

vgvassilev (commented):
@anutosh491, can you take a look?

anutosh491:
> @anutosh491, can you take a look?

Hi, 

Thanks for the work here. Just finishing working a couple other things. So shall review this by today or tomorrow at max !

lib/Interpreter/CMakeLists.txt
```diff
@@ -1,33 +1,7 @@
...
-  # is patched out of emsdk, as --soname is not recognised by emscripten. A PR to do this has been done here https://github.com/emscripten-core/emscripten/pull/23453 
-  #FIXME: A patch is needed to llvm to remove -Wl,-z,defs since it is now recognised on emscripten. What needs to be removed is here
-  # https://github.com/llvm/llvm-project/blob/128e2e446e90c3b1827cfc7d4d19e3c0976beff3/llvm/cmake/modules/HandleLLVMOptions.cmake#L318 . The PR to do try to do this is here 
-  # https://github.com/llvm/llvm-project/pull/123396
-  set_target_properties(clangCppInterOp
-  PROPERTIES NO_SONAME 1
-)
-  target_link_options(clangCppInterOp PRIVATE
-    PUBLIC "SHELL: -s WASM_BIGINT"
-  )
```
anutosh491, line 30: CONCERN 1:

I got a copy of these changes and was trying them locally .... only to realize we don't generate a shared library no more. This was alarming as I think we for sure need a shared/static build for Jupyterlite. The whole point why this exists in xeus-cpp https://github.com/compiler-research/xeus-cpp/blob/f05495180c191947e6c4675d5ce1881a2ab8425c/share/jupyter/kernels/xcpp20/wasm_kernel.json.in#L12-L13

So I quickly hit up a Jupyterlite maintainer on Zulip to confirm this and this is his reply !

"
@Anutosh (anutosh491) so your question is if we can use a wasm binary instead of a shared/static library in xeus-lite?

No we cannot
"

So although I haven't gotten to a point where I can review code changes to get the tests working .... I shall first look at it from a point where we are not spoiling anything we already have. 

And even though we may or may need a wasm binary (`libclangCppInterOp.wasm` in this case) for running the tests .... I don't think we should remove our shared build.

mcbarton: @anutosh491 We are still generating a shared library. I just changed the suffix to .wasm, which is considered an acceptable extension for shared libraries in Emscripten (see the Emscripten documentation here https://emscripten.org/docs/compiling/Dynamic-Linking.html#load-time-dynamic-linking). The change in the suffix was to stop having situation like https://github.com/emscripten-core/emscripten/issues/16592 where someone build the Emscripten shared library make confuse it with native shared library.

anutosh491: 1) So it is just an extension change ? I guess it might be cause the file type looks similar .
```
(xeus-cpp-wasm-build) anutosh491@Anutoshs-MacBook-Air build % file lib/libclangCppInterOp.so 
lib/libclangCppInterOp.so: WebAssembly (wasm) binary module version 0x1 (MVP)

(xeus-cpp-wasm-build) anutosh491@Anutoshs-MacBook-Air build % file lib/libclangCppInterOp.wasm 
lib/libclangCppInterOp.wasm: WebAssembly (wasm) binary module version 0x1 (MVP)
```

2) Hmmm, that being I don't think we should get rid of the `so` extension anyways cause to my understanding working with jupyterlite-xeus that's what it understands. Not sure if this calls for duplication and if that's an issue

3) On top of this I have only come across shared libraries for emscripten being built with this construct.

https://github.com/emscripten-core/emscripten/issues/15276#issuecomment-1039349267

All shared libs are pyodide/emscripten-forge are built using this !

So yeah not yet saying anything is right/wrong here but I'll have to do more research if we change stuff in this regard.

mcbarton: @anutosh491 I'm fine with reverting this change about the suffix, as I'm not too bothered about keeping it. Having the compile and link flags set under set_target_properties is better though than our current approach, since the current approach scatters everything over multiple cmake functions, and this new approachmore easily allowed me to modify just the link flags for the export.

anutosh491: > Having the compile and link flags set under set_target_properties is better though than our current approach, since the current approach scatters everything over multiple cmake functions, and this new approach more easily allowed me to modify just the link flags for the export.

Okay let's do something in this case.

If you think you have a method to improve our current emscripten build ....
1) Raise a PR for that and let's have it reviewed
2) And then we can rebase this PR on top of that.

This being something new I would absolutely like to keep implementation and refactoring/improvements separate.
Being the initial pr that enables testing, the sole goal of this PR should probably be to have tests working without changing (or refactoring/improving) what we already have.

Just not healthy to review it this way. 

If you don't think you want to create a new PR .... what I would recommend is lets stick to what we have and then improve stuff separately (basically let's remove all the new improvements from this and only have enough to get the tests working)

I would go by atomic commits rather than large changes for something like this.

lib/Interpreter/CMakeLists.txt
```diff
@@ -73,6 +53,7 @@ else()
     clangLex
     clangSema
     )
+endif()
 
   if(NOT WIN32)
     list(APPEND link_libs dl)
```
anutosh491, line 59: dl doesn't look necessary for the emscripten case either.

lib/Interpreter/CMakeLists.txt
```diff
@@ -133,6 +113,31 @@ else()
...
+  string(REPLACE "\n" " " SYMBOLS_LIST "${SYMBOLS_LIST}")
+  
+  #FIXME: Setting no_soname=1 is needed until https://github.com/emscripten-core/emscripten/blob/ac676d5e437525d15df5fd46bc2c208ec6d376a3/cmake/Modules/Platform/Emscripten.cmake#L36
+  # is patched out of emsdk, as --soname is not recognised by emscripten. A PR to do this has been done here https://github.com/emscripten-core/emscripten/pull/23453 
+  #FIXME: A patch is needed to llvm to remove -Wl,-z,defs since it is now recognised on emscripten. What needs to be removed is here
+  # https://github.com/llvm/llvm-project/blob/128e2e446e90c3b1827cfc7d4d19e3c0976beff3/llvm/cmake/modules/HandleLLVMOptions.cmake#L318 . The PR to do try to do this is here 
+  # https://github.com/llvm/llvm-project/pull/123396
+  set_target_properties(clangCppInterOp PROPERTIES 
+    NO_SONAME 1
+    COMPILE_FLAGS "-s SIDE_MODULE=1"
```
anutosh491, line 134: Why do we need this compile_flag here ? 

I know SIDE_MODULE=1 can be used as both a compile time and a link time flag but its use case is not too extensive as a compile time flag as far as I know.

CMakeLists.txt
```diff
@@ -361,11 +361,15 @@ endif()
 
 # Add appropriate flags for GCC
 if (LLVM_COMPILER_IS_GCC_COMPATIBLE)
-  if (APPLE)
+  if (APPLE OR EMSCRIPTEN)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-common -Woverloaded-virtual -Wcast-qual -fno-strict-aliasing -Wno-long-long -Wall -W -Wno-unused-parameter -Wwrite-strings")
   else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-common -Woverloaded-virtual -Wcast-qual -fno-strict-aliasing -pedantic -Wno-long-long -Wall -W -Wno-unused-parameter -Wwrite-strings")
```
anutosh491, line 367: Among both of these I see that the pedantic flag is whats different between both the "set...." options we have 

I don't think pedantic has a huge role here but do you have any specific reason to remove it ?

mcbarton: The reason for removing it is because it is passed through to the emscripten build of gtest, creating a warning, and as we treat all warnings as errors, then the gtest build errors out. Same reason it was remove during the apple build.

lib/Interpreter/CMakeLists.txt
mcbarton: My understanding is that is what the CMAKE_SHARED_LIBRARY_CREATE_CXX/C_FLAGS did previously. If I am wrong, then I am happy to remove it.

CMakeLists.txt
anutosh491: Okay if that's the case I am not sure of this change. 

This is the global cmake file and making changes here affect not just the build for unittests but also the library.

I see the following CXX_FLAGS are being passed after your change while building `libclangCppInterOp.so`

```
CXX_FLAGS =  -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough  -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -ffunction-sections -fdata-sections -fno-common -Woverloaded-virtual -Wcast-qual -fno-strict-aliasing -Wno-long-long -Wall -W -Wno-unused-parameter -Wwrite-strings  -Wno-sign-compare -O3 -DNDEBUG -std=c++17 -fPIC -s SIDE_MODULE=1 -UNDEBUG
```
And this doesn't have `pedantic` which it used to have originally.

I haven't gotten to the gtest thing yet. My first aim is to ensure the library that we were building already is built without any unncessary changes. 

So yeah this might have to play a role for gtest (but if that's the case you try supressing it there to avoid duplication) . Supressing it here also affects how we build the shared object.

lib/Interpreter/CMakeLists.txt
```diff
@@ -41,6 +15,7 @@ else()
   if ("LLVMFrontendDriver" IN_LIST LLVM_AVAILABLE_LIBS)
     list(APPEND LLVM_LINK_COMPONENTS  FrontendDriver)
   endif()
+endif()
   if ("LLVMOrcDebugging" IN_LIST LLVM_AVAILABLE_LIBS)
     list(APPEND LLVM_LINK_COMPONENTS OrcDebugging)
   endif()
```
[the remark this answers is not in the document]
mcbarton: The endif should have been after the OrcDebugging. Its to separate out what we define for LLVM_LINK_COMPONENTS in both the emscripten and non emscripten cases. Will change.

lib/Interpreter/CMakeLists.txt
mcbarton: Will change.

lib/Interpreter/CMakeLists.txt
anutosh491: Ahh okay but no `CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS` deals with link time flags !!

So if you see now we have these CXX_FLAGS
```
CXX_FLAGS =  -fPIC ....... lot of flags .......-s SIDE_MODULE=1 -UNDEBUG
```

Earlier we had 

```
CXX_FLAGS =  -fPIC .....lot of flags ... -fno-rtti -UNDEBUG
```

Basically our use case never involved using `SIDE_MODULE=1` as a compile time flag .

lib/Interpreter/CMakeLists.txt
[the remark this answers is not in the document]
anutosh491: thanks

lib/Interpreter/CMakeLists.txt
anutosh491: thanks .

lib/Interpreter/CMakeLists.txt
mcbarton: I will remove this compile flag, and assuming no tests start failing that did pass, then I will leave it removed.

CMakeLists.txt
anutosh491: > This fix is what we settled on when we encountered the same issue upgrading to the latest gtest for MacOS builds

Not sure if this was discussed before. Don't have enough context. 

Neither do I think pedantic is a game changer here ! But commenting as a reviewer is tough unless I don't know of the error (looks just randomly removed to me)

Maybe paste the error here ?

lib/Interpreter/CMakeLists.txt
mcbarton: Done.

lib/Interpreter/CMakeLists.txt
mcbarton: Done.

lib/Interpreter/CMakeLists.txt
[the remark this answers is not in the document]
mcbarton: Done.

CMakeLists.txt
vgvassilev: > @vgvassilev On this point I don't agree. This change is necessary. This fix is what we settled on when we encountered the same issue upgrading to the latest gtest for MacOS builds. No idea why its acceptable in the case of native builds, but no Emscripten builds.

Can we get more context on the error?

lib/Interpreter/exports.ld
```diff
@@ -0,0 +1,62 @@
...
+-Wl,--export=_ZNK5clang11DeclContext11decls_beginEv
+-Wl,--export=_ZNK5clang11DeclContext6lookupENS_15DeclarationNameE
+-Wl,--export=_ZNK5clang12FunctionDecl29getTemplateSpecializationArgsEv
+-Wl,--export=_ZNK5clang12FunctionDecl31getTemplateInstantiationPatternEb
+-Wl,--export=_ZNK5clang17ClassTemplateDecl18getSpecializationsEv
+-Wl,--export=_ZNK5clang29VarTemplateSpecializationDecl22getSpecializedTemplateEv
+-Wl,--export=_ZNK5clang4Decl13getASTContextEv
+-Wl,--export=_ZNK5clang4Decl15hasDefiningAttrEv
+-Wl,--export=_ZNK5clang4Decl8getAttrsEv
+-Wl,--export=_ZN4llvm3sys2fs6accessERKNS_5TwineENS1_10AccessModeE
```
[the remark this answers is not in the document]
anutosh491: 1) We probably need to take a step back and see what symbols come out of which transitive lib
2) Then look into if these are propagated correctly into clangInterpreter.
3) Then look into if they are propagated into clangCppInterOp

Otherwise this can be so tedious and shabby to maintain. Any change in llvm or here and there involves changing this I would guess.

CMakeLists.txt
mcbarton: @anutosh491 @vgvassilev Here is what happens if you keep the pedantic flag ttps://github.com/compiler-research/CppInterOp/actions/runs/13412927760/job/37466965423?pr=483#step:9:250 .

lib/Interpreter/exports.ld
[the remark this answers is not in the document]
mcbarton: @anutosh491 @vgvassilev These come out as undefined symbols before the refactoring, so the refactoring is not the issue. I can show they come out as undefined symbols if this in any way help. 

On a side note (which may or may not be related to this issue), how does xeus-cpp-lite even dynamically link to CppInterop? According to one of the maintainers of Emscripten you need to use main_module to link to a shared library with Emscripten (see here https://github.com/emscripten-core/emscripten/issues/23543#issuecomment-2625334414), and i see no reference to it in xeus-cpp. Its not entirely clear that xeus-cpp-lite is even treating CppInterOp as a shared library based on this warning https://github.com/compiler-research/xeus-cpp/actions/runs/13399607465/job/37427009943#step:5:97 .

mcbarton:
@anutosh491 Please do not mark the PRs I make as draft, when I have set them as ready for review. If I have put it as 'ready for review', then it is ready for review, and am asking for feedback. This also counts for xeus-cpp PRs. I find it quite annoying that you do this. You can say that a PR does not have your approval, but that is not the same as saying it is not ready for review.

cc @vgvassilev

lib/Interpreter/exports.ld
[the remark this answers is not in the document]
anutosh491: > you need to use main_module to link to a shared library with Emscripten

I mean we for sure need to have main_module=1 to generate the xcpp.wasm executable 😅 
You might have missed that but it comes from here (xeus)
https://github.com/jupyter-xeus/xeus/blob/a752d1432fcd5e5d2e8b45180be4cd528d864862/cmake/WasmBuildOptions.cmake#L40

Our current approach regarding running clang-repl in the browser involves loading side modules built at runtime (one for each cell) on top of the main module xcpp.wasm one by one (cell by cell).

So that's far from my concern here.

anutosh491:
> Please do not mark the PRs I make as draft, when I have set them as ready for review. If I have put it as 'ready for review', then it is ready for review, and am asking for feedback. This also counts for xeus-cpp PRs. I find it quite annoying that you do this. You can say that a PR does not have your approval, but that is not the same as saying it is not ready for review.


I mean isn't that how it is supposed to go ?

1) I wasn't taking part in the PR and was working on different tasks
2) I get "invited" for a review after an initial set of reviews here.
3) My understanding tells me this is far from going in.
4) I as a reviewer convert it to draft (only so that I can convert it back to ready once I am satisfied)
5) I don't do this if we think we are close to ready.

This is also partly done so that no other maintainer who is satisfied maybe goes ahead and merges it .... Rather the other guy should see that Anutosh is not yet happy and has converted it to draft. This is what I saw maintainers do in SymPy (13.3k stars) and LPython (1.5k stars) when I used to contribute daily to these as an open source newbie.

CMakeLists.txt
vgvassilev: Okay. In that case can we limit flag just for gtest?

mcbarton: This is the initial approach we took in the gtest PR (here https://github.com/compiler-research/CppInterOp/pull/344#discussion_r1825484822), but we ended up at the solution we have here (I think to still have consistency between the flags applied to gtest and CppInterOp).

lib/Interpreter/exports.ld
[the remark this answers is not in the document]
mcbarton: @vgvassilev If we drop these exports then the main_modules (the tests) won't even build, let alone be able to run the tests. I prefer to keep the solution we have as is. If a better solution can be found in the future then this export solution can be replaced.

vgvassilev:
> > Please do not mark the PRs I make as draft, when I have set them as ready for review. If I have put it as 'ready for review', then it is ready for review, and am asking for feedback. This also counts for xeus-cpp PRs. I find it quite annoying that you do this. You can say that a PR does not have your approval, but that is not the same as saying it is not ready for review.
> 
> I mean isn't that how it is supposed to go ?
> 
> 1. I wasn't taking part in the PR and was working on different tasks
> 2. I get "invited" for a review after an initial set of reviews here.
> 3. My understanding tells me this is far from going in.
> 4. I as a reviewer convert it to draft (only so that I can convert it back to ready once I am satisfied)
> 5. I don't do this if we think we are close to ready.
> 
> This is also partly done so that no other maintainer who is satisfied maybe goes ahead and merges it .... Rather the other guy should see that Anutosh is not yet happy and has converted it to draft. This is what I saw maintainers do in SymPy (13.3k stars) and LPython (1.5k stars) when I used to contribute daily to these as an open source newbie 3 years back

Can you briefly summarize why this is not ready to go in? Is there a design problem?

lib/Interpreter/exports.ld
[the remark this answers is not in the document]
anutosh491: I was just discussing this with @aaronj0 over video a couple days back and what we discussed is this
Okay it looks like a quick fix but can easily end up being a shabby solution. 

1) Each time the code changes or new symbols are added, the exports.ld file needs to be updated manually.
2) This creates a brittle dependency that can easily fall out of sync, leading to missing symbols and runtime errors.

We were exactly standing at the same step 3-4 months back, when I was facing this with xeus-cpp-lite (through xcpp.wasm and xcpp.js) . Pasting an image from the [missing symbols issue](https://github.com/compiler-research/CppInterOp/issues/334) I had raised 

image

**I can see about 85-90% symbols present in this file above (I see 63 symbols) are also present in my issue. So if we really wanted to go this way I could have cherry picked symbols this way and had gotten xeus-cpp-lite way back.**
But we moved from a static approach to a shared approach only to tackle the failing symbol resolution and now linking and generating xcpp.wasm has no issues. 

For eg I see the symbol `_ZN5clang26IncrementalCompilerBuilder9CreateCppEv` being exported. This is literally the symbol that is reponsible for creating the interpreter and xeus-cpp-lite would not even work if it couldn't access this symbol. 

And on top of it, emscripten doesn't really encourage exporting unncessary symbols anyways, which means we kinda need to go through these cherry picked symbols and make sure we need each one of them.

I would take a step back and use exactly what we did for getting xeus-cpp-lite 

https://github.com/compiler-research/CppInterOp/pull/483#discussion_r1961598252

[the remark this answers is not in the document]
anutosh491: Answering to the comment above.

> Also there is a reason why I have labelled is a fixme

There is nothing wrong with adding a fixme. But who takes the responsibility for fixing it ? 
I would take it up but only after I have done my research and can promise a better approach.
I'm not even concerned about the fix for now .... but the approach. All of us collectively don't even have an concrete approach in mind to fix the fixme.

> I have no idea why you need to export these symbols to be able to link the test Javascripts to the shared library.I am open to any suggestion to avoid this export, but I don't have any idea how to fix it.

Exactly, we need to figure out fundamentally why this is needed. We need to think of ideas if we don't have any.

> If you cannot think of how to avoid this export then my solution is to keep as is, but open up an issue in CppInterOp outlining the issue, leaving the fixme in place.

I don't side by that. According to my experience of working with building stuff with emscripten for the past 1.5 years  on a daily basis, I see undefined symbols everyday.

 But that doesn't mean, I have seen any of the good libraries come up with such a workaround. There are around 600 libraries on pyodide + emscripten-forge and **a lot of these can only do a shared build** (I'm talking about pretty fundamental libraries like Blas and Lapack here) just like we are doing ... but I haven't come across any exporting a file having ginormous symbols listed !
 
 Do you know of any other libraries (the good ones) doing this ? Please let me know !

anutosh491:
> Can you briefly summarize why this is not ready to go in? Is there a design problem?

This is what Aaron and I had discussed 2 days back 
1) https://github.com/compiler-research/CppInterOp/pull/483#discussion_r1965304382
2) https://github.com/compiler-research/CppInterOp/pull/483#discussion_r1965304564

I would be putting in some time here (hopefully soon in the coming week) and try to see how we can get around any flakiness here and having stuff polished.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -12,16 +21,32 @@ add_cppinterop_unittest(CppInterOpTests
...
+  # Explaination of compile and link flags
+  # To dynamically link to a shared library for Emscripten builds you must use the MAIN_MODULE flag (see both https://github.com/emscripten-core/emscripten/issues/23543#issuecomment-2625334414
+  # and https://emscripten.org/docs/compiling/Dynamic-Linking.html)
+  # Without WASM_BIGINT flag then you get fatal errors when trying to run compiled Javascript about trying to convert between BigInt and non BigInt types
+  # EXPORTED_RUNTIME_METHODS='[\"FS\",\"PATH\",\"LDSO\",\"loadDynamicLibrary\",\"ERRNO_CODES\"]' and --preload-file ${SYSROOT_PATH}/include@/include are not allow the Javascript that is 
+  # compiled to have access to the standard library headers (approach taken from xeus-cpp)
+  # Without ALLOW_MEMORY_GROWTH=1 tests will fail with aborted(OOM). Approach to fix taken from answers to
+  # https://stackoverflow.com/questions/67222200/runtimeerror-abortoom-build-with-s-assertions-1-for-more-info
+  set_target_properties(CppInterOpTests PROPERTIES
+    COMPILE_FLAGS "-s MAIN_MODULE=1"
```
anutosh491, line 37: Shouldn't be needed as a compile flag

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -37,16 +62,38 @@ export_executable_symbols(CppInterOpTests)
...
+  # Explaination of compile and link flags
+  # To dynamically link to a shared library for Emscripten builds you must use the MAIN_MODULE flag (see both https://github.com/emscripten-core/emscripten/issues/23543#issuecomment-2625334414
+  # and https://emscripten.org/docs/compiling/Dynamic-Linking.html)
+  # Without WASM_BIGINT flag then you get fatal errors when trying to run compiled Javascript about trying to convert between BigInt and non BigInt types
+  # EXPORTED_RUNTIME_METHODS='[\"FS\",\"PATH\",\"LDSO\",\"loadDynamicLibrary\",\"ERRNO_CODES\"]' and --preload-file ${SYSROOT_PATH}/include@/include are not allow the Javascript that is 
+  # compiled to have access to the standard library headers (approach taken from xeus-cpp)
+  # Without ALLOW_MEMORY_GROWTH=1 tests will fail with aborted(OOM). Approach to fix taken from answers to
+  # https://stackoverflow.com/questions/67222200/runtimeerror-abortoom-build-with-s-assertions-1-for-more-info
+  set_target_properties(DynamicLibraryManagerTests PROPERTIES
+    COMPILE_FLAGS "-s MAIN_MODULE=1"
```
anutosh491, line 82: Shouldn't be needed as a compile flag

anutosh491:
Installation needs to be improved I think. When I build on this branch I see 
```
Install the project...
-- Install configuration: "Release"
-- Up-to-date: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/lib/cmake/CppInterOp
-- Installing: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/lib/cmake/CppInterOp/CppInterOpConfig.cmake
-- Installing: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/lib/cmake/CppInterOp/CppInterOpConfigVersion.cmake
-- Up-to-date: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/include
-- Up-to-date: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/include/clang
-- Up-to-date: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/include/clang/Interpreter
-- Installing: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/include/clang/Interpreter/CppInterOp.h
-- Up-to-date: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/include/clang-c
-- Installing: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/include/clang-c/CXCppInterOp.h
-- Installing: /Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/lib/libclangCppInterOp.so
```
Which is what we already used to install isn't it ?
This PR adds quite some builds (not sure what is important or can be skipped. In any case shouldn't we  provide CppInterOpTests.js and CppInterOpTests.wasm if we enable testing ? )

lib/Interpreter/CMakeLists.txt
```diff
@@ -124,7 +104,6 @@ else()
       clangStaticAnalyzerCore
       )
   endif(LLVM_LINK_LLVM_DYLIB)
-
   add_llvm_library(clangCppInterOp
     DISABLE_LLVM_LINK_LLVM_DYLIB
```
anutosh491, line 108: Not sure what role this plays for plays for the emscripten build. I remember as we were forcing a shared build ... we were using 
```
  add_llvm_library(clangCppInterOp
    SHARED
    ......

```

lib/Interpreter/CMakeLists.txt
```diff
@@ -133,6 +112,30 @@ else()
...
+  #FIXME: Setting no_soname=1 is needed until https://github.com/emscripten-core/emscripten/blob/ac676d5e437525d15df5fd46bc2c208ec6d376a3/cmake/Modules/Platform/Emscripten.cmake#L36
+  # is patched out of emsdk, as --soname is not recognised by emscripten. A PR to do this has been done here https://github.com/emscripten-core/emscripten/pull/23453 
+  #FIXME: A patch is needed to llvm to remove -Wl,-z,defs since it is now recognised on emscripten. What needs to be removed is here
+  # https://github.com/llvm/llvm-project/blob/128e2e446e90c3b1827cfc7d4d19e3c0976beff3/llvm/cmake/modules/HandleLLVMOptions.cmake#L318 . The PR to do try to do this is here 
+  # https://github.com/llvm/llvm-project/pull/123396
+  set_target_properties(clangCppInterOp PROPERTIES 
+    NO_SONAME 1
+    LINK_FLAGS "-s WASM_BIGINT  -s SIDE_MODULE=1 ${SYMBOLS_LIST}"
+  )
+  # When compiling Emscripten tests the shared library it links to is expected to be in the same folder as the compiled Javascript
```
anutosh491, line 135: I guess this is only needed if tests are enabled (default is on but the user can switch it off)

anutosh491:
> I also just want to make it clear just like I did in our private Discord chat that the goal of this PR is not to have all tests passing.

Hi,

just curious here. Do we atleast know why some work and some don't ?

lib/Interpreter/CMakeLists.txt
anutosh491: Just wanted to make sure `DISABLE_LLVM_LINK_LLVM_DYLIB` plays a part for the emscripten build as we weren't using it earlier. If you're confident we need it, i'm happy

lib/Interpreter/exports.ld
```diff
@@ -0,0 +1,62 @@
+-Wl,--export=_ZN4llvm3sys2fs17getMainExecutableEPKcPv
```
anutosh491, line 1: I actually found a couple of ways to get rid of all the symbols coming out of llvm and clang. So almost 50+ from these 62 symbols can be taken care off easily. Though I've landed on the rest right now 
```
error: undefined symbol: _ZN3Cpp11GetOperatorEPvNS_8OperatorERNSt3__26vectorIS0_NS2_9allocatorIS0_EEEENS_13OperatorArityE (referenced by root reference (e.g. compiled C/C++ code))
warning: To disable errors for undefined symbols use `-sERROR_ON_UNDEFINED_SYMBOLS=0`
warning: __ZN3Cpp11GetOperatorEPvNS_8OperatorERNSt3__26vectorIS0_NS2_9allocatorIS0_EEEENS_13OperatorArityE may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: clang_Interpreter_dispose (referenced by root reference (e.g. compiled C/C++ code))
warning: _clang_Interpreter_dispose may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: clang_Interpreter_getClangInterpreter (referenced by root reference (e.g. compiled C/C++ code))
warning: _clang_Interpreter_getClangInterpreter may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: clang_Interpreter_takeInterpreterAsPtr (referenced by root reference (e.g. compiled C/C++ code))
warning: _clang_Interpreter_takeInterpreterAsPtr may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: clang_createInterpreterFromRawPtr (referenced by root reference (e.g. compiled C/C++ code))
warning: _clang_createInterpreterFromRawPtr may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: clang_getComplexType (referenced by root reference (e.g. compiled C/C++ code))
warning: _clang_getComplexType may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: clang_getTypeAsString (referenced by root reference (e.g. compiled C/C++ code))
warning: _clang_getTypeAsString may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: clang_instantiateTemplate (referenced by root reference (e.g. compiled C/C++ code))
warning: _clang_instantiateTemplate may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
```

1) all but one of these symbols are coming out of https://github.com/compiler-research/CppInterOp/blob/7caf10b1ccae74e62600b1b5fe05dc4e71db93bb/lib/Interpreter/CXCppInterOp.cpp

2) `__ZN3Cpp11GetOperatorEPvNS_8OperatorERNSt3__26vectorIS0_NS2_9allocatorIS0_EEEENS_13OperatorArityE` comes from here https://github.com/compiler-research/CppInterOp/blob/7caf10b1ccae74e62600b1b5fe05dc4e71db93bb/lib/Interpreter/CppInterOp.cpp#L3448

Let's take out some time and think why these symbols are not being exported.

For 1) We do have this 

https://github.com/compiler-research/CppInterOp/blob/7caf10b1ccae74e62600b1b5fe05dc4e71db93bb/lib/Interpreter/CMakeLists.txt#L11

So technically we are compiling it as a part of libclangCppInterOp.so 
So not sure what hinders the export.

For 2) Not sure why that's the only symbol coming out of `CppInterOp.cpp` that is not exported. Maybe something with visibility or so (I don't know)

anutosh491: Actually I understood what's happening here. Shall explain offline

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -58,6 +58,9 @@ TEST(InterpreterTest, DebugFlag) {
 }
 
 TEST(InterpreterTest, Evaluate) {
+#ifdef EMSCRIPTEN
+  GTEST_SKIP() << "Test fails for Emscipten builds";
+#endif
```
anutosh491, line 63: Hmmm technically we should be able to test this in the browser. 

What error are you encountering when you do this.

I see this through xeus-cpp-lite. We use c++20 hence, we have `202002` and I tried the other Evaluate thingy too

image

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -145,6 +151,9 @@ TEST(InterpreterTest, DISABLED_DetectResourceDir) {
 }
 
 TEST(InterpreterTest, DetectSystemCompilerIncludePaths) {
+#ifdef EMSCRIPTEN
+  GTEST_SKIP() << "Test fails for Emscipten builds";
+#endif
```
anutosh491, line 156: Hmm, this is a weird spot actually. We don't expect `DetectSystemCompilerIncludePaths` to really work when we work in the browser correct ? Cause CppInterOp's `DetectSystemCompilerIncludePaths` uses `exec`

https://github.com/compiler-research/CppInterOp/blob/7caf10b1ccae74e62600b1b5fe05dc4e71db93bb/lib/Interpreter/CppInterOp.cpp#L2947-L2954

and `exec` has been marked as "won't-fix" by emscripten. So technically this is not expected to work (unless some change I guess is introduced in DetectSystemCompilerIncludePaths)

So we can prove it should be empty for this use case.

image

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -532,6 +532,9 @@ TEST(FunctionReflectionTest, ExistsFunctionTemplate) {
 }
 
 TEST(FunctionReflectionTest, InstantiateTemplateFunctionFromString) {
+#ifdef EMSCRIPTEN
+  GTEST_SKIP() << "Test crashes gtest on Emscipten";
+#endif
```
anutosh491, line 537: I see this also returns a non nullptr so technically EXPECT_TRUE should work here 

image

anutosh491:
Regarding my last 3 reviews. **Obviously as you said the goal might not be to get all tests runnings here, so issues.**

I am just trying to get an idea on why something would not work. 

For eg 

1) regarding the [first review](https://github.com/compiler-research/CppInterOp/pull/483#discussion_r1978737953) ... we've marked an evaluate call as "fails for emscripten" but evaluate technically just calls process and we haven't had any problems with process. So we kinda like to track down what might be going wrong here !

Maybe we might be overlooking something obvious !

anutosh491:
Hey @mcbarton 

I just realized https://github.com/compiler-research/CppInterOp/pull/440 went in. This should also do the job for us and expose all symbols we need. 

So we no longer need #518 (and I am closing that issue)

Dumping all the symbols should be enough to see if they are being exposed correctly
```
anutosh491@Anutoshs-MacBook-Air lib % emnm --debug-syms libclangCppInterOp.so > test.txt

004b3c48 T _Z14linkComponentsv
.........
0048640d T clang_Interpreter_addIncludePath
0048636a T clang_Interpreter_addSearchPath
004864a4 T clang_Interpreter_declare
00486308 T clang_Interpreter_dispose
004867a4 T clang_Interpreter_evaluate
004861f7 T clang_Interpreter_getClangInterpreter
00486a6f T clang_Interpreter_loadLibrary
004868c7 T clang_Interpreter_lookupLibrary
00486625 T clang_Interpreter_process
00486234 T clang_Interpreter_takeInterpreterAsPtr
00486249 T clang_Interpreter_undo
00486b8f T clang_Interpreter_unloadLibrary
00486789 T clang_Value_dispose
0048786f T clang_allocate
00487885 T clang_construct
0048609d T clang_createInterpreter
004861e1 T clang_createInterpreterFromRawPtr
0048674c T clang_createValue
0048787a T clang_deallocate
00487a63 T clang_destruct
004874f2 T clang_existsFunctionTemplate
00486e15 T clang_getComplexType
00486f70 T clang_getDefaultConstructor
00487103 T clang_getDestructor
00487241 T clang_getFunctionSignature
00486c2e T clang_getTypeAsString
00486f29 T clang_hasDefaultConstructor
0048769e T clang_instantiateTemplate
004878d3 T clang_invoke
0048749f T clang_isTemplatedFunction
00486f1b T clang_scope_dump
............
0401e06d T llvm_strlcpy
```

If you rebase your PR on latest main and remove all symbols relevant to the c API (https://github.com/compiler-research/CppInterOp/pull/483/files#diff-054935aee7c0887220251f89b2179dbc4a0c418a69bca9f74854305ae7c0f967R31)
we should be good here.

lib/Interpreter/CMakeLists.txt
```diff
@@ -134,9 +113,37 @@ else()
...
+  string(REPLACE "\n" " " SYMBOLS_LIST "${SYMBOLS_LIST}")
+  
+  #FIXME: Setting no_soname=1 is needed until https://github.com/emscripten-core/emscripten/blob/ac676d5e437525d15df5fd46bc2c208ec6d376a3/cmake/Modules/Platform/Emscripten.cmake#L36
+  # is patched out of emsdk, as --soname is not recognised by emscripten. A PR to do this has been done here https://github.com/emscripten-core/emscripten/pull/23453 
+  #FIXME: A patch is needed to llvm to remove -Wl,-z,defs since it is now recognised on emscripten. What needs to be removed is here
+  # https://github.com/llvm/llvm-project/blob/128e2e446e90c3b1827cfc7d4d19e3c0976beff3/llvm/cmake/modules/HandleLLVMOptions.cmake#L318 . The PR to do try to do this is here 
+  # https://github.com/llvm/llvm-project/pull/123396
+  set_target_properties(clangCppInterOp PROPERTIES 
+    NO_SONAME 1
+    LINK_FLAGS "-s WASM_BIGINT  -s SIDE_MODULE=1 ${SYMBOLS_LIST}"
```
anutosh491, line 134: Hi, 

I see set_target_properties being put to use in this PR.

If possible, can we stick to 
i) target_link_libraries when linking with libraries
ii) target_link_options when setting link options.

Those are the 2 most fundamental functions cmake has to offer when it comes to linking. So we should put them to use wherever possible.

unittests/CMakeLists.txt
```diff
@@ -7,10 +7,15 @@ if (NOT TARGET gtest)
   include(GoogleTest)
 endif()
 
-set(gtest_libs gtest gtest_main)
-# Clang prior than clang13 (I think) merges both gmock into gtest.
-if (TARGET gmock)
-  list(APPEND gtest_libs gmock gmock_main)
+if(EMSCRIPTEN)
+  set(gtest_libs gtest gmock)
```
anutosh491, line 11: Hi, 

can someone please explain why we only need `gtest` and `gmock` when it comes to emscripten (and possibly gtest_main and gmock_main too for other platforms) ? 

Sorry I might be missing something because of my limited experience with googletest as a whole.

anutosh491: Ahhh is this the reasoning behind this 

https://github.com/compiler-research/CppInterOp/pull/483/files#r1948150573

If yes, I think this is correct. I see why an emscripten build would need a main.cpp !

But again what happens if we link against gtest_main ? We eventually need to call main() isn't it ? 
The flag which ensures main() runs even when using Emscripten’s default JS glue code is `-s INVOKE_RUN=1` so if simply linking with gtest_main doesn't run the test and we need an extra file, we can try building it with the above flag. 

Maybe something for the future if we really don't want the extra file.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -12,16 +21,31 @@ add_cppinterop_unittest(CppInterOpTests
...
+  # To dynamically link to a shared library for Emscripten builds you must use the MAIN_MODULE flag (see both https://github.com/emscripten-core/emscripten/issues/23543#issuecomment-2625334414
+  # and https://emscripten.org/docs/compiling/Dynamic-Linking.html)
+  # Without WASM_BIGINT flag then you get fatal errors when trying to run compiled Javascript about trying to convert between BigInt and non BigInt types
+  # EXPORTED_RUNTIME_METHODS='[\"FS\",\"PATH\",\"LDSO\",\"loadDynamicLibrary\",\"ERRNO_CODES\"]' and --preload-file ${SYSROOT_PATH}/include@/include are not allow the Javascript that is 
+  # compiled to have access to the standard library headers (approach taken from xeus-cpp)
+  # Without ALLOW_MEMORY_GROWTH=1 tests will fail with aborted(OOM). Approach to fix taken from answers to
+  # https://stackoverflow.com/questions/67222200/runtimeerror-abortoom-build-with-s-assertions-1-for-more-info
+  set_target_properties(CppInterOpTests PROPERTIES
+    LINK_FLAGS "-s MAIN_MODULE=1 -s WASM_BIGINT -s EXPORTED_RUNTIME_METHODS='[\"FS\",\"PATH\",\"LDSO\",\"loadDynamicLibrary\",\"ERRNO_CODES\"]' --preload-file ${SYSROOT_PATH}/include@/include -s ALLOW_MEMORY_GROWTH=1"
+  )
```
anutosh491, line 38: Hey @mcbarton 

I think I was able to get tests to atleast run (if they run they either pass or fail but not crash. Crash here means the build is not correct and that's not what we want for now. I think failure is fine because we atleast know the reason why it fails too)

For starters I removed the "crash" message from all tests from the ScopeReflectionTest.cpp and some from the FunctionReflectionTests.cpp and all pass. 

For that we need to address some flags missing here. (most of it are needed while we run lite, some like Wasm=1 work well when running Main_Module=1 and I'm sure the EXPORTED_RUNTIME_METHODS don't play a role here. The purpose why they are present in xeus-cpp is completely different and don't assist tests in any way)

```
  target_compile_options(CppInterOpTests // may not be required but good to have
    PUBLIC "SHELL: -fexceptions"
  )

  target_link_options(CppInterOpTests
    PUBLIC "SHELL: -fexceptions"
    PUBLIC "SHELL: -s MAIN_MODULE=1"
    PUBLIC "SHELL: -s WASM=1"
    PUBLIC "SHELL: -s WASM_BIGINT"
    PUBLIC "SHELL: -s ASSERTIONS=0"
    PUBLIC "SHELL: -s ALLOW_MEMORY_GROWTH=1"
    PUBLIC "SHELL: -s EXIT_RUNTIME=1"
    PUBLIC "SHELL: -s STACK_SIZE=32mb"
    PUBLIC "SHELL: -s INITIAL_MEMORY=128mb"
    PUBLIC "SHELL: --preload-file ${SYSROOT_PATH}/include@/include"
  )
endif()
```

anutosh491: After using this I think tests would either pass or fail (not crash) which is good.

As I wrote above I tried some tests where I could see the crash message. I would like if you try all of them. And if anything fails we mark it as "expected fail" rather than "crash" for now.

anutosh491: And the best part is we try and literally replicate it just to confirm it.

for eg I see 6 "crash" messages in `FunctionReflectionTests.cpp`

If I uncomment the last 2, they still build and run but they fail saying
```
1: /Users/anutosh491/work/CppInterOp/unittests/CppInterOp/FunctionReflectionTest.cpp:1467: Failure
1: Expected equality of these values:
1:   output
1:     Which is: "/Users/anutosh491/work/CppInterOp/unittests/CppInterOp/FunctionReflectionTest.cpp:1465: Failure\nValue of: object != nullptr\n  Actual: false\nExpected: true\n\n"
1:   "Constructor Executed"
1: With diff:
1: @@ -1,5 +1,1 @@
1: -/Users/anutosh491/work/CppInterOp/unittests/CppInterOp/FunctionReflectionTest.cpp:1465: Failure
1: -Value of: object != nullptr
1: -  Actual: false
1: -Expected: true
1: -\n
1: +Constructor Executed
1: 
```


So basically an object ends up being a nullptr. I can confirm this failure through lite too


image

**So after this change, I don't see anything crashing but only failing and we have an error message to confirm if it is really the reason !!!!!**

anutosh491: Same change for `DynamicLibraryManagerTests` too

anutosh491: I ran all tests and this is what I see from the tests you've marked as "crash"

1) ScopeReflectionTest everything passes
2) FunctionReflectionTest - 6 were marked as crashing (1st one passes, the rest fail)
3) TypeReflectionTest - everything passes
4) JitTest - 1 marked as crashing , now fails
5) VariableReflectionTest - 2 were marked as crashing, VariableOffsetsWithInheritance now passes, GetVariableOffset now fails

anutosh491 (approved):
Looks okay to me

PR #504, opened 2025-02-17
Disable JitCall test that fails on windows + cling

aaronj0:


unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -868,6 +868,9 @@ TEST(FunctionReflectionTest, JitCallAdvanced) {
 TEST(FunctionReflectionTest, GetFunctionCallWrapper) {
   if (llvm::sys::RunningOnValgrind())
     GTEST_SKIP() << "XFAIL due to Valgrind report";
+#if defined(USE_CLING) && defined(_WIN32)
+  GTEST_SKIP() << "Disabled, invoking functions containing printf does not work with Cling";
```
mcbarton, line 872: Shouldn't this specifically mention not working on Windows, instead saying it doesn't work with Cling (since it works with Cling on Unix systems)?

PR #505, opened 2025-02-18
Prefix the interpreter cmake options

aaronj0:
Prefix `USE_REPL` and `USE_CLING`

### DIFF
--- a/.github/workflows/MacOS-arm.yml
+++ b/.github/workflows/MacOS-arm.yml
@@ -396,8 +396,8 @@ jobs:
         cling_on=$(echo "${{ matrix.cling }}" | tr '[:lower:]' '[:upper:]')
         if [[ "${cling_on}" == "ON" ]]; then
           cmake -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}        \
-                -DUSE_CLING=ON                                  \
-                -DUSE_REPL=OFF                                  \
+                -DCPPINTEROP_USE_CLING=ON                                  \
+                -DCPPINTEROP_USE_REPL=OFF                                  \
                 -DCPPINTEROP_INCLUDE_DOCS=${{ matrix.documentation }} \
                 -DCling_DIR=$LLVM_BUILD_DIR/tools/cling         \
                 -DLLVM_DIR=$LLVM_BUILD_DIR/lib/cmake/llvm       \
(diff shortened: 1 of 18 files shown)


mcbarton:
@aaronj0 I don't understand the motivation for this change. Seems random especially given the PR description doesn't say why it needs to be done, just what is being done.

cc @vgvassilev

aaronj0:
> @aaronj0 I don't understand the motivation for this change. Seems random. The PR description doesn't say why it needs to be done, just what is being done.
> 
> cc @vgvassilev
> 
> Also the workflows (and probably the code too even though I haven't checked) in cppyy and related repos all assume the old options, so will break a lot of things. If any other repo which builds on CppInterOp we don't work on is using the old options they will break too.

This is nor random. This is a requirement for ROOT and has been discussed with @vgvassilev
`USE_CLING `does not specify which library this options is for and is not good practice for Cmake, when integrating with projects with a larger build tree. This is what LLVM does too, as you can see all their options are preefixed with `LLVM_`

aaronj0:
@mcbarton As for your concerns about the workflows, it's something that can be just updated. Other than the cppyy repos, xeus does not use this option anywhere since InterOp is not configured in any way and uses clang-repl by default.

mcbarton:
> > @aaronj0 I don't understand the motivation for this change. Seems random. The PR description doesn't say why it needs to be done, just what is being done.
> > cc @vgvassilev
> > Also the workflows (and probably the code too even though I haven't checked) in cppyy and related repos all assume the old options, so will break a lot of things. If any other repo which builds on CppInterOp we don't work on is using the old options they will break too.
> 
> This is not random. This is a requirement for ROOT and has been discussed with @vgvassilev `USE_CLING `does not specify which library this options is for and is not good practice for Cmake, when integrating with projects with a larger build tree. This is what LLVM does too, as you can see all their options are prefixed with `LLVM_`

Even so, its best to have this reason in the PR description, because without it, anybody stumbling across the PR who was not party to the conversation between you and @vgvassilev would have no idea why this change was made. Can you make the required changes in the cppyy related repos, so they don't break with this change?

vgvassilev (approved):
LGTM. It's quite annoying that so many yaml files had to be updated.

vgvassilev:
> > > @aaronj0 I don't understand the motivation for this change. Seems random. The PR description doesn't say why it needs to be done, just what is being done.
> > > cc @vgvassilev
> > > Also the workflows (and probably the code too even though I haven't checked) in cppyy and related repos all assume the old options, so will break a lot of things. If any other repo which builds on CppInterOp we don't work on is using the old options they will break too.
> > 
> > 
> > This is not random. This is a requirement for ROOT and has been discussed with @vgvassilev `USE_CLING `does not specify which library this options is for and is not good practice for Cmake, when integrating with projects with a larger build tree. This is what LLVM does too, as you can see all their options are prefixed with `LLVM_`
> 
> Even so, its best to have this reason in the PR description, because without it, anybody stumbling across the PR who was not party to the conversation between you and @vgvassilev would have no idea why this change was made. Can you make the required changes in the cppyy related repos, so they don't break with this change?

That was an urgent request from our ROOT client and we need to act quickly...

PR #511, opened 2025-02-26
Improve template instantiate

Vipul-Cariappa:
# Description

Match parameter types to argument types in cases of when the parameter type itself is templated.

### Fixes # (issue)

Helps fix up to 4 tests in cppyy.

Also fixes the https://github.com/compiler-research/cppyy-backend/pull/129 's cling failure.

Please tick all options which are relevant.

- [x] Bug fix
- [ ] New feature
- [ ] Requires documentation updates

## Testing

Included tests. I feel codecov will still not be happy. Will add more according to its suggestions.

## Checklist

- [x] I have read the contribution guide recently

PR #514, opened 2025-03-05
Add `BestOverloadFunctionMatch` & `IsFunction` Remove `BestTemplateFunctionMatch`

Vipul-Cariappa:
# Description

Resolve and instantiate (if templated) overload from vector of overloads for the given template parameter types and argument types.
With these changes, we replicate clang's overload resolution and template instantiation.

## Fixes # (issue)

Fixes 10 test cases in cppyy.
Can close https://github.com/compiler-research/CppInterOp/pull/352 and https://github.com/compiler-research/CppInterOp/pull/511.

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

Included tests

## Checklist

### DIFF
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -1075,7 +1075,7 @@ namespace Cpp {
     struct WrapperExpr : public OpaqueValueExpr {
       WrapperExpr() : OpaqueValueExpr(clang::Stmt::EmptyShell()) {}
     };
-    WrapperExpr Exprs[arg_types.size()];
+    auto* Exprs = new WrapperExpr[arg_types.size()];
     llvm::SmallVector<Expr*> Args;
     Args.reserve(arg_types.size());
     size_t idx = 0;
@@ -1135,7 +1135,9 @@ namespace Cpp {
     OverloadCandidateSet::iterator Best;
     Overloads.BestViableFunction(S, SourceLocation(), Best);
 
-    return Best != Overloads.end() ? Best->Function : nullptr;
+    FunctionDecl* Result = Best != Overloads.end() ? Best->Function : nullptr;
+    delete[] Exprs;
+    return Result;
   }
 
   // Gets the AccessSpecifier of the function and checks if it is equal to
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -900,7 +900,7 @@ namespace Cpp {
       if (Type->isUndeducedAutoType() && IsTemplatedFunction(FD) &&
           !FD->isDefined()) {
 #ifdef CPPINTEROP_USE_CLING
-        cling::Interpreter::PushTransactionRAII RAII(&I);
+        cling::Interpreter::PushTransactionRAII RAII(&getInterp());
 #endif
         getSema().InstantiateFunctionDefinition(SourceLocation(), FD, true,
                                                 true);
@@ -1061,6 +1061,10 @@ namespace Cpp {
                             const std::vector<TemplateArgInfo>& arg_types) {
     auto& S = getSema();
     auto& C = S.getASTContext();
+
+#ifdef CPPINTEROP_USE_CLING
+    cling::Interpreter::PushTransactionRAII RAII(&getInterp());
+#endif
 
     llvm::SmallVector<Expr*> Args;
     for (auto i : arg_types) {
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -1055,6 +1055,7 @@ namespace Cpp {
         funcs.push_back(Found);
   }
 
+  // Adapted from inner workings of Sema::BuildCallExpr
   TCppFunction_t
   BestOverloadFunctionMatch(const std::vector<TCppFunction_t>& candidates,
                             const std::vector<TemplateArgInfo>& explicit_types,
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -1072,16 +1072,23 @@ namespace Cpp {
     // Here we synthesize a placeholder expression to be able to use
     // Sema::AddOverloadCandidate. Made up expressions are fine because the
     // interface uses the list size and the expression types.
+    struct WrapperExpr : public OpaqueValueExpr {
+      WrapperExpr() : OpaqueValueExpr(clang::Stmt::EmptyShell()) {}
+    };
+    WrapperExpr Exprs[arg_types.size()];
     llvm::SmallVector<Expr*> Args;
     Args.reserve(arg_types.size());
+    size_t idx = 0;
     for (auto i : arg_types) {
       QualType Type = QualType::getFromOpaquePtr(i.m_Type);
       ExprValueKind ExprKind = ExprValueKind::VK_PRValue;
       if (Type->isReferenceType())
         ExprKind = ExprValueKind::VK_LValue;
-      Args.push_back(new (C)
-                         OpaqueValueExpr(SourceLocation::getFromRawEncoding(1),
-                                         Type.getNonReferenceType(), ExprKind));
+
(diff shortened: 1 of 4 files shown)


lib/Interpreter/CppInterOp.cpp
```diff
@@ -1027,61 +1056,78 @@ namespace Cpp {
...
-      total_arg_set.insert(total_arg_set.end(), explicit_types.begin(),
-                           explicit_types.end());
-      total_arg_set.insert(total_arg_set.end(), arg_types.begin(),
-                           arg_types.end());
-
-      instantiated = InstantiateTemplate(candidate, total_arg_set.data(),
-                                         total_arg_set.size());
-      if (instantiated)
-        return instantiated;
+    llvm::SmallVector<Expr*> Args;
```
vgvassilev, line 1069: Here we should write a comment of the sort:

"The overload resolution interfaces in Sema require a list of expressions. However, unlike handwritten C++, we do not always have a expression. Here we synthesize a placeholder expression to be able to use Sema::AddOverloadCandidate. Made up expressions are fine because the interface uses the list size and the expression types."

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1027,61 +1056,78 @@ namespace Cpp {
...
+    for (auto i : arg_types) {
+      QualType Type = QualType::getFromOpaquePtr(i.m_Type);
+      ExprValueKind ExprKind = ExprValueKind::VK_PRValue;
+      if (Type->isReferenceType())
+        ExprKind = ExprValueKind::VK_LValue;
+      Args.push_back(new (getSema().getASTContext())
```
vgvassilev, line 1075: Construct the Expr on the stack and pass it as a reference. This way if we start using something outside of the types and expression values valgrind will be able to tell us (upon future changes of the implementation in clang). To do this you will need an `std::vector<Expr> Exprs` with the same lifetime as `Args` and `Args` taking `Exprs` addresses elementwise.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1027,61 +1056,78 @@ namespace Cpp {
...
+    OverloadCandidateSet Overloads(
+        SourceLocation(), OverloadCandidateSet::CandidateSetKind::CSK_Normal);
+    for (void* i : candidates) {
+      Decl* D = static_cast<Decl*>(i);
+      if (auto* MD = dyn_cast<CXXMethodDecl>(D)) {
+        S.AddMethodCandidate(
+            MD, DeclAccessPair::make(MD, MD->getAccess()), MD->getParent(),
+            C.getTypeDeclType(MD->getParent()),
+            Expr::Classification::makeSimpleLValue(), Args, Overloads);
+      } else if (auto* FD = dyn_cast<FunctionDecl>(D)) {
```
vgvassilev, line 1110: [AddOverloadCandidate](https://clang.llvm.org/doxygen/classclang_1_1Sema.html#affc2ef4040a1ebccc5e3a59d68807f85) already calls `AddMethodCandidate` if we got a `CXXMethodDecl`...

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1027,61 +1056,78 @@ namespace Cpp {
...
+    }
+
+    TemplateArgumentListInfo TLI{};
+    for (auto TA : TemplateArgs)
+      TLI.addArgument(
+          S.getTrivialTemplateArgumentLoc(TA, QualType(), SourceLocation()));
+
+    OverloadCandidateSet Overloads(
+        SourceLocation(), OverloadCandidateSet::CandidateSetKind::CSK_Normal);
+    for (void* i : candidates) {
```
vgvassilev, line 1103: Why not using [Sema::AddFunctionCandidates](https://clang.llvm.org/doxygen/classclang_1_1Sema.html#abb87d8a7ae2840db51030e162f3018dc) instead of this loop?

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1027,61 +1056,78 @@ namespace Cpp {
   }
 
   TCppFunction_t
-  BestTemplateFunctionMatch(const std::vector<TCppFunction_t>& candidates,
+  BestOverloadFunctionMatch(const std::vector<TCppFunction_t>& candidates,
```
vgvassilev, line 1059: This function feels very similar to `Sema::BuildCallExpr` can we use that instead and unwrap the result from the call expression? I think the cost would be an AST node allocation per call but we currently do that anyway...

Vipul-Cariappa (commented):
I will make the necessary changes in my next commit.
For the comments where I have not replied.

lib/Interpreter/CppInterOp.cpp
vgvassilev: Maybe that assertion is coming from an interface on our end that needs fixing?

lib/Interpreter/CppInterOp.cpp
vgvassilev: Yes, it will but we do that anyway. If you prefer not to do it, then write a comment that this is copied from there. FWIW, that did not seem the case when I looked...

lib/Interpreter/CppInterOp.cpp
vgvassilev: I seem to remember that we could create an expression on the stack but apparently I am confusing myself. We should probably put it in some scratch area, call placement new there and then destroy it...

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1027,61 +1056,73 @@ namespace Cpp {
...
```
vgvassilev, line 1059: Another note is that we should check the memory growth when we call this function N times. Ideally it should be 0 for failed overload resolution candidate set. In reality it should be close to 0...

lib/Interpreter/CppInterOp.cpp
vgvassilev: > We will still require a loop to populate UnresolvedSetImpl from candidates.

It seems to have an `append` operator taking a range...

lib/Interpreter/CppInterOp.cpp
Vipul-Cariappa: Not sure how I would go about destroying it. The default `delete` operator is marked unreachable (reference: https://github.com/llvm/llvm-project/blob/release/19.x/clang/include/clang/AST/Stmt.h#L106-L109). And the `delete` operators defined are empty (reference: https://github.com/llvm/llvm-project/blob/release/19.x/clang/include/clang/AST/Stmt.h#L1289-L1293).
I can call into one of these `delete` operators if required?

lib/Interpreter/CppInterOp.cpp
Vipul-Cariappa: > It seems to have an append operator taking a range...

The iterator is a [`UnresolvedSetIterator`](https://clang.llvm.org/doxygen/classclang_1_1UnresolvedSetIterator.html)(reference: https://clang.llvm.org/doxygen/classclang_1_1UnresolvedSetImpl.html#ad1e4dde3cd343da5f9918eb0caedd0a7)

---

> Maybe that assertion is coming from an interface on our end that needs fixing?

I have been debugging this. I am not able to detect the cause. Our interfaces are being used correctly, as far as I can see.

vgvassilev (approved):
Nicely done. Please run clang-format and squash on merge.

Vipul-Cariappa:
Hi @mcbarton, the Windows CI is failing in the build step of CppInterOp. I am not sure what is causing this. Can you please look into this. This PR is good to go otherwise. I expect it to fail while building cppyy, with an error message similar to the Ubuntu/OSX CI run (but I guess we don't test cppyy on Windows).

Vipul-Cariappa:
@mcbarton, I have fixed the CI problem. Now it looks like a code problem. Sorry, I should have read the error message properly before pinging you.

---

@vgvassilev, it looks like MSVC does not support Variable Length Array, reference: https://stackoverflow.com/a/33423538.
How to proceed?

PR #516, opened 2025-03-07
Add explicit extern "C"

### LINKED ISSUE #515: [Bug]: extern "C" missing?
### Description of bug

There is [this](https://github.com/compiler-research/CppInterOp/blob/7caf10b1ccae74e62600b1b5fe05dc4e71db93bb/include/clang-c/CXCppInterOp.h#L15C1-L15C28), but `LLVM_CLANG_C_EXTERN_C_BEGIN` does not seem to be defined. 

Am I missing something obvious? The binaries (at least on win) from conda-forge do not export anything like e.g. `clang_Interpreter_evaluate` using the c linkage.

If you need why: I'd like to call this lib via ctypes in python.

### What operating system was you using when the bug occured?

Windows

### What is the architechture of the cpu on your system?

x64

### What did you build CppInterOp against?

Not sure, I used the conda-forge build

adam-urbanczyk:
# Description

I added explicit `extern "C"`

Fixes #515

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/include/clang-c/CXCppInterOp.h
+++ b/include/clang-c/CXCppInterOp.h
@@ -12,7 +12,7 @@
 #include <stddef.h>
 #include <stdint.h>
 
-LLVM_CLANG_C_EXTERN_C_BEGIN
+extern "C" {
 
 /**
  * \defgroup CPPINTEROP_INTERPRETER_MANIP Interpreter manipulations
@@ -360,7 +360,7 @@ void clang_destruct(CXObject This, CXScope S, bool withFree);
  * @}
  */
 
-LLVM_CLANG_C_EXTERN_C_END
+}
 
 #endif // LLVM_CLANG_C_CXCPPINTEROP_H
        // NOLINTEND()


include/clang-c/CXCppInterOp.h
```diff
@@ -12,7 +12,7 @@
 #include <stddef.h>
 #include <stdint.h>
 
-LLVM_CLANG_C_EXTERN_C_BEGIN
+extern "C" {
```
vgvassilev, line 15: Can we include `clang-c/ExternC.h` which defines the macro instead?

Gnimuc:
maybe just merge #440?

vgvassilev:
> maybe just merge #440?

Good point. @adam-urbanczyk, can you still reproduce the issue with the master?

anutosh491:
> can you still reproduce the issue with the master?

Hi, 

I think the problem was relevant to the win use-case 
but I had seen similar issues related to symbol visibility for C API when doing an emscripten build.

Those problems are resolved now once #440 went in (check https://github.com/compiler-research/CppInterOp/pull/483#issuecomment-2712960406)
and I think it would also be addressed with other platform builds.

include/clang-c/CXCppInterOp.h
adam-urbanczyk: clang-c/ExternC.h is already included, it looks like this PR does not make sense.

vgvassilev:
Thank you!

PR #524, opened 2025-03-17
Use target_link_options instead of set_target_properties for linking

anutosh491:
# Description

CppInterOp used to use `target_link_options`  before #483 went in. Not sure why this change was made. set_target_properties comes under the old style cmake and `target_link_options` and `target_compile_options` falls under mordern cmake which cmake promotes to be used with `target_link_libraries`.

This PR reverts back to what we were originally using. `add_llvm_library` internally make use of target_link_libraries and we should use `target_link_options` that goes with it .

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist


anutosh491:
Was requested here https://github.com/compiler-research/CppInterOp/pull/483#discussion_r1992791439 originally.

mcbarton:
You should probably make the change here too https://github.com/anutosh491/CppInterOp/blob/03fe9d122a8f756befaf77d699ad720cf10af6f3/unittests/CppInterOp/TestSharedLib/CMakeLists.txt#L14 as part of this PR.

PR #529, opened 2025-03-17
Add doctest to Emscripten conda environment

mcbarton:
# Description

Once this PR in xeus-cpp https://github.com/compiler-research/xeus-cpp/pull/277 is ready and merged, doctest will need to be added Emscripten conda environment in CppInterOp. 

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/environment-wasm.yml
+++ b/environment-wasm.yml
@@ -8,3 +8,4 @@ dependencies:
   - xeus
   - cpp-argparse
   - pugixml
+  - doctest


vgvassilev:
CppInterOp does not depend on doctest. Why do we need this?

mcbarton:
> CppInterOp does not depend on doctest. Why do we need this?

@vgvassilev Because xeus-cpp does, and once the automated Emscripten tests PR reference in the PR description is in, xeus-cpp will try and find doctest to link against for the tests (tests are on and build by default). We could in theory set the xeus-cpp tests to OFF by a cmake option, but by keeping them on we will be able to have another check that no changes to CppInterOp break the Emscripten build of xeus-cpp.

vgvassilev:
I am not sure I completely understand that dependency. The CppInterOp tests and the xeus-cpp tests should be separate and independent from each other. If there is some required dependency then we have a problem.

mcbarton:
> I am not sure I completely understand that dependency. The CppInterOp tests and the xeus-cpp tests should be separate and independent from each other. If there is some required dependency then we have a problem.

I can change this PR to turn off xeus-cpp tests regardless of what the default value in xeus-cpp is, but something will need to be done in preparation for the PR mentioned in the PR description. 

Just one thing to note the environment-wasm.yml file is only in this repo because of our xeus-cpp-lite build. Nothing in that file is a dependency of CppInterOp.

vgvassilev:
Now I think I understand. That’s a ci dependency. Okay I think the pr is good to go from my side.

mcbarton:
> Now I think I understand. That’s a ci dependency. Okay I think the pr is good to go from my side.

Would you like to wait for the automated wasm tests PR in xeus-cpp to go in first, or can this go on now?

vgvassilev:
This can go.

vgvassilev (approved):
Lgtm.

PR #531, opened 2025-03-17
Reduce the Emscripten build size and reduce build time by building libclang target instead of the clang target

mcbarton:
# Description

Clang.js takes a while to build. We don't use clang.js, we just use the library. In theory we should be able to pass all Emscripten tests by building the libclang target instead of clang target. This will reduce the Emscripten cache (freeing up some very much needed space in the Github cache with the llvm 20 upgrade PR to come), and will speed up build times for developers. If the ci passes, then I will update the documentation accordingly. 

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -327,7 +327,7 @@ jobs:
                         -DLLVM_ENABLE_THREADS=OFF                       \
                         -G Ninja                                         \
                         ../llvm
-          emmake ninja clang clangInterpreter clangStaticAnalyzerCore lld -j ${{ env.ncpus }}
+          emmake ninja libclang clangInterpreter clangStaticAnalyzerCore lld -j ${{ env.ncpus }}
         fi
         cd ../
         rm -rf $(find . -maxdepth 1 ! -name "build" ! -name "llvm" ! -name "clang" ! -name ".")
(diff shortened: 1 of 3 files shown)


anutosh491:
Technically we don't need any tool (neither the js nor the wasm). For example currently how we build llvm, we build so many tools along the way like clang-repl.js/ clang-repl.wasm and none of this is put to use. 

I even tried putting these to use actually. For example we build llvm-nm.js/wasm ; llvm-c++filt.js/wasm (so technically there is nothing stopping us from using these with jupyterlite's terminal). I tried using these tools there but they are far from working :|

So its not that these tools are even being built correctly, even if they are. Would be great if we could only build libclangInterpreter.a (and its static transitive libs)

mcbarton:
> 
> So its not that these tools are even being built correctly, even if they are. Would be great if we could only build libclangInterpreter.a (and its static transitive libs)

@anutosh491 That is already done in the ci. You can build just them by swaping the clang-repl target for clangInterpreter clangStaticAnalyzerCore targets.

I thought the liblld target would exist without checking (I already knew the libclang target existed). I will investigate swapping out the lld target for a future PR

PR #533, opened 2025-03-18
General improvements to Emscripten build and ci

mcbarton:
# Description

This PR makes general improvements to Emscripten build and ci
 
- Reduces the size of the Emscripten cache by removing NATIVE tools folder
- If developer decided to build all targets, then they get a reduced number of targets that are not relevant (new options inspired by https://github.com/emscripten-core/emsdk/blob/b665079cb9ad9eb1704652f962281a7fa1633e2d/emsdk.py#L1124C42-L1124C66 in emsdk and https://github.com/WebAssembly/wasi-sdk/blob/d94a133384eb7a805ad7185195db96435d2ecdac/cmake/wasi-sdk-toolchain.cmake#L86 in wasi sdk)
- Ninja defaults to a larger number of parallel processes than the number of cpu cores. This means we currently build slower using Ninja than it is capable of
- We do not need to set the compiler version Emscripten jobs, or the Python version.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -21,25 +21,22 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: ubu24-arm-gcc12-clang-repl-19-emscripten
+          - name: ubu24-arm-clang-repl-19-emscripten
             os: ubuntu-24.04-arm
-            compiler: gcc-12
             clang-runtime: '19'
             cling: Off
             llvm_enable_projects: "clang;lld"
             llvm_targets_to_build: "WebAssembly"
             emsdk_ver: "3.1.73"
-          - name: osx15-arm-clang-clang-repl-19-emscripten
+          - name: osx15-arm-clang-repl-19-emscripten
             os: macos-15
-            compiler: clang
             clang-runtime: '19'
             cling: Off
             llvm_enable_projects: "clang;lld"
             llvm_targets_to_build: "WebAssembly"            
             emsdk_ver: "3.1.73"
-          - name: ubu24-x86-gcc12-clang-repl-19-emscripten
+          - name: ubu24-x86-clang-repl-19-emscripten
             os: ubuntu-24.04
-            compiler: gcc-12
             clang-runtime: '19'
             cling: Off
             llvm_enable_projects: "clang;lld"
@@ -112,7 +109,7 @@ jobs:
         path: |
           llvm-project
           ${{ matrix.cling=='On' && 'cling' || '' }}
-        key: ${{ env.CLING_HASH }}-${{ runner.os }}-${{ matrix.os }}-${{ matrix.compiler }}-clang-${{ matrix.clang-runtime }}.x-emscripten
+        key: ${{ env.CLING_HASH }}-${{ runner.os }}-${{ matrix.os }}-clang-${{ matrix.clang-runtime }}.x-emscripten
         lookup-only: true
 
     - name: Setup emsdk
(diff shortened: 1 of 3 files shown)


mcbarton:
> These changes are super small. Let's try to put up bigger pull requests that change more things in a more uniform way. Eg. if we have 10 times to improve in the CI let's do them at once instead of generating 10 PRs.



> These changes are super small. Let's try to put up bigger pull requests that change more things in a more uniform way. Eg. if we have 10 times to improve in the CI let's do them at once instead of generating 10 PRs.

I will do this from now on (I have the very big change in the works to have reusable actions in the workflow to stop repetition). In most cases i'll have to just put the PR title and description as general ci improvements. The reason I put this llvm cache PR as small incremental changes, was so that if anything went wrong it was obvious. as to what target change or in this case remove of the folder was responsible in case any test broke. 

image

The reason I am trying to get small amounts from the cache size is that we are close to the 10GB limit, and currently we won't be able to get the llvm 20 PR in and stay under this limit (its going to be very close). If I cannot get us under this 10GB limit, do you have any preference as to what jobs get dropped from the ci?

mcbarton:
@vgvassilev can you please approve this PR? I've added a few improvements to the Emscripten ci as part this PR (detailed in the PR description). If you approve I will merge when activity is quiet in the repo, since it requires me to clear some builds in the Github cache.

anutosh491:
Hi,

Mostly looks good. Just a question here and there (afk currently, shall type them out soon )

.github/workflows/emscripten.yml
```diff
@@ -325,10 +232,13 @@ jobs:
                         -DLLVM_INCLUDE_EXAMPLES=OFF                     \
                         -DLLVM_INCLUDE_TESTS=OFF                        \
                         -DLLVM_ENABLE_THREADS=OFF                       \
+                        -DLLVM_BUILD_TOOLS=OFF                          \
+                        -DLLVM_ENABLE_LIBPFM=OFF                        \
```
anutosh491, line 236: Could you educate me on what this last flag does and why do we need it ?

mcbarton: @anutosh491 going by the LLVM documentation 
```
LLVM_ENABLE_LIBPFM:BOOL

    Enable building with libpfm to support hardware counter measurements in LLVM tools. Defaults to ON.
```
I got the idea for the flag emsdk build of llvm here https://github.com/emscripten-core/emsdk/blob/b665079cb9ad9eb1704652f962281a7fa1633e2d/emsdk.py#L1124C42-L1124C66 . I thought if its good enough for the developers of Emscripten to think it can be off, then its good enough for CppInterOp.

mcbarton:
@anutosh491 if you approve, please leave it to me to merge this PR, since I need to clear some stuff from the cache first before merging.

anutosh491:
Actually I just realized that if we build with -DLLVM_BUILD_TOOLS=OFF, it only makes sense to also use `CLANG_BUILD_TOOLS=OFF` too isn't it ?

mcbarton:
> Actually I just realized that if we build with -DLLVM_BUILD_TOOLS=OFF, it only makes sense to also use `CLANG_BUILD_TOOLS=OFF` too isn't it ?

@anutosh491 I can do this additional change to the PR. Do you approve the PR subject to this change?

anutosh491 (approved):
Looks good. 

Maybe add the clang-enable-tools=off config in the docs too ?

PR #534, opened 2025-03-20
Add type info interfaces motivated by numba

aaronj0:
- Add `IsSameType` used for matching arg types between numba inferred signatures and CppOverloads in cppyy.
- Add type info reflection interfaces for integer, void pointer, signed and unsigned types.
- Also added interfaces `IsIntegral` and `IsFloating` that check if types have the respective representations. Used in the case of Numba where scoring based on reflection is required.

will add tests

### DIFF
--- a/include/CppInterOp/CppInterOp.h
+++ b/include/CppInterOp/CppInterOp.h
@@ -93,6 +93,7 @@ enum Operator : unsigned char {
 };
 
 enum OperatorArity : unsigned char { kUnary = 1, kBinary, kBoth };
+enum Signedness : unsigned char { kSigned = 1, kUnsigned, kAny };
 
 /// A class modeling function calls for functions produced by the interpreter
 /// in compiled code. It provides an information if we are calling a standard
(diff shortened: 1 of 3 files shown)


vgvassilev:
We should add tests.

jeaye:
Would love to have this. What's needed before we can merge?

aaronj0:
> Would love to have this. What's needed before we can merge?

Hi! I have some unittests that I haven't pushed yet, and this PR should be ready to go. Will do in the next few hours once I have access to a laptop

include/clang/Interpreter/CppInterOp.h
```diff
@@ -509,15 +509,40 @@ namespace Cpp {
   /// Checks if the provided parameter is a Plain Old Data Type (POD).
   CPPINTEROP_API bool IsPODType(TCppType_t type);
 
+  /// Checks if type has an integer representation
+  CPPINTEROP_API bool IsIntegerType(TCppType_t type);
```
vgvassilev, line 513: Probably we should have an enum instead of many interfaces..

include/CppInterOp/CppInterOp.h
```diff
@@ -578,6 +579,22 @@ CPPINTEROP_API bool IsRecordType(TCppType_t type);
 /// Checks if the provided parameter is a Plain Old Data Type (POD).
 CPPINTEROP_API bool IsPODType(TCppType_t type);
 
+/// Checks if type has an integer representation
+CPPINTEROP_API bool IsIntegerType(TCppType_t type,
+                                  Signedness s = Signedness::kAny);
```
vgvassilev, line 584:
```suggestion
                                  Signedness *s = nullptr);
```
We should probably take that as an output argument because you can have an integer type that's of different signedness and is still an int type.

aaronj0: Yes, what you describe will happen if the caller specifies signed or unsigned. I was of the opinion that if you call this by default (for kAny) it will return true if the qualtype has an integer representation which should cover the above scenario

vgvassilev: Yes, but if you don't know the underlying signedness you will get a `false` telling you that this is not an integral type. This assumes that you will call the function couple of times before you figure everything out. My proposal is to enable this in a single call.

aaronj0: Done, I've opened a fresh PR instead since this was hard to rebase: https://github.com/compiler-research/CppInterOp/pull/913

aaronj0:
Closing in favour of https://github.com/compiler-research/CppInterOp/pull/913

PR #535, opened 2025-03-21
Add `LookupConstructors` and fix logic in `IsConstructor` for templates

aaronj0:
Add `LookupConstructors` and fixes logic in `IsConstructor` for templates. Accumulate constructors with this new lookup in `GetClassTemplatedMethods` enabling 8 tests. Also splits IsTemplatedFunction into `IsTemplatedFunction` for FunctionTemplateDecl, and `IsTemplateInstantiationOrSpecialization` when we have an instantiation or a specialization.

### DIFF
--- a/include/clang/Interpreter/CppInterOp.h
+++ b/include/clang/Interpreter/CppInterOp.h
@@ -416,13 +416,27 @@ namespace Cpp {
   CPPINTEROP_API bool ExistsFunctionTemplate(const std::string& name,
                                              TCppScope_t parent = nullptr);
 
+  /// Sets a list of all the constructor for a scope/class that is
+  /// supplied as a parameter.
+  ///\param[in] name - This string is used as a constraint, that clients can use
+  ///           to ensure the constructors match the name that they provide
+  ///\param[in] parent - Pointer to the scope/class for which the constructors
+  ///           are being looked up
+  ///           to be retrieved
+  ///\param[out] funcs - vector of handles to all constructors found under the
+  ///            given scope
+  CPPINTEROP_API void LookupConstructors(const std::string& name,
+                                         TCppScope_t parent,
+                                         std::vector<TCppFunction_t>& funcs);
+
   /// Sets a list of all the Templated Methods that are in the Class that is
   /// supplied as a parameter.
+  ///\returns true if the lookup succeeded, and false if there are no candidates
   ///\param[in] name - method name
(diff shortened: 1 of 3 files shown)


include/clang/Interpreter/CppInterOp.h
```diff
@@ -422,7 +422,7 @@ namespace Cpp {
   ///\param[in] parent - Pointer to the scope/class under which the methods have
   ///           to be retrieved
   ///\param[out] funcs - vector of function pointers matching the name
-  CPPINTEROP_API void
+  CPPINTEROP_API bool
```
vgvassilev, line 425: We need to specify what that return result means.

aaronj0: Added

aaronj0:
The added tests here depend on #536 to obtain the function signatures correctly for `FunctionTemplateDecls`. Without this fix, templated constructors don't return their actual signatures.

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1031,38 +1035,69 @@ namespace Cpp {
...
 
-    if (!parent || name.empty())
-      return;
+    if (auto* CXXRD = llvm::dyn_cast_or_null<CXXRecordDecl>(D)) {
+      getSema().ForceDeclarationOfImplicitMembers(CXXRD);
+      DeclContextLookupResult Result = getSema().LookupConstructors(CXXRD);
+      // Obtaining all constructors when we intend to lookup a method under a
+      // scope can lead to crashes. We avoid that by accumulating constructors
+      // only if the Decl matches the lookup name.
+      if (!Result.empty())
```
vgvassilev, line 1056:
```suggestion
```

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1031,38 +1035,68 @@ namespace Cpp {
...
+      for (auto* Found : R)
+        if (IsTemplatedFunction(Found))
+          funcs.push_back(Found);
+    }
 
-    R.resolveKind();
+    // Ambiguously found
+    // FIXME : if result not empty, add them anyway?
+    else {
+      if (!R.empty()) {
```
vgvassilev, line 1093:
```suggestion
```

vgvassilev (approved):
LGTM!

PR #546, opened 2025-04-01
Support NixOS resource dirs

jeaye:
Previously, CppInterOp assumed that the resource dir would end in the clang major version. This is common, but it's not always the case. On NixOS, for example, we have something like this:

```
❯ clang++ -print-resource-dir
/nix/store/xl0vlc2wdchfbq8536zs19pj2r3xdmma-clang-wrapper-19.1.5/resource-root
```

I have updated the function to check that the path contains the major version, rather than ending in it. This is still a useful version check, though it's not as precise as it was before.

### DIFF
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -3042,7 +3042,7 @@ namespace Cpp {
         CLANG_VERSION_MAJOR_STRING;
 #endif
     // We need to check if the detected resource directory is compatible.
-    if (llvm::sys::path::filename(detected_resource_dir) != version)
+    if (detected_resource_dir.find(version) == std::string::npos)
       return "";
 
     return detected_resource_dir;


vgvassilev:
Can we somehow add a test for this?

jeaye:
> Can we somehow add a test for this?

Seems possible, if we were to:

1. Create a mock executable script which prints a resource dir for normal situations
2. Create a mock executable script which prints a resource dir for NixOS situations
3. Create a unit test for that function which passes in the path of the custom scripts

I didn't bother, since that didn't seem worth the effort. If this is required, please let me know where the scripts should live.

vgvassilev:
Perhaps we can move the compatibility check logic into a new function that takes a string and then implement a test based on the expected strings?

mcbarton:
Although it wouldn't cover NixOS, if you can suggest another OS which is not following the typical pattern, but for which a docker image exists on Docker hub (I couldn't find an image for NixOS), I should be able to construct a workflow for this operating system, and therefore provide a way to test this change.

PR #547, opened 2025-04-01
Return pointer to non-const for getExecutionEngine

jeaye:
The `clang::Interpreter` wrapper previously returned a pointer to const, but the actual `clang::Interpreter` did not. This breaks code for anyone porting from `clang::Interpreter` to `Cpp::Interpreter`, especially since most operations on the execution engine are mutations of some sort.

Yes, the user could static cast the wrapper to the inner interpreter, but, at that point, why both exposing the `getExecutionEngine` function at all?


vgvassilev (approved):
Lgtm!

PR #550, opened 2025-04-02
Reduce cling llvm cache (makes room for Emscripten llvm 20 builds)

mcbarton:
# Description

Currently we are have just enough room in the cache to include the native builds of llvm 20 (see image below). My plan was to remove the llvm 19 Windows job to make room for the Emscripten llvm 20, but that won't be enough given we are so close to capacity. This PR should reduce the impact of the cling llvm builds enough that we have enough room for the Emscripten builds. I'll be able to quantify the impact of this PR once the workflows finish running.

image

## Checklist

### DIFF
--- a/.github/workflows/MacOS-arm.yml
+++ b/.github/workflows/MacOS-arm.yml
@@ -225,7 +225,8 @@ jobs:
           rm -rf $(find . -maxdepth 1 ! -name "include" ! -name "lib" ! -name "cmake" ! -name "utils" ! -name ".")
           cd ../clang/
           rm -rf $(find . -maxdepth 1 ! -name "include" ! -name "lib" ! -name "cmake" ! -name "utils" ! -name ".")
-          cd ../..
+          cd ../../cling/
+          rm -rf $(find . -maxdepth 1 ! -name "include" ! -name "lib" ! -name "cmake" ! -name ".")
         else # repl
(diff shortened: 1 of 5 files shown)


mcbarton:
This PR would save 140mb of cache space, which although not much, when you add on the space saved by removing the llvm 19 Windows job (450mb), then you have enough space for all Emscripten llvm 20 jobs.

anutosh491:
Went through this! 
To be fair my knowledge with respect to how the cache works with respect to the CI is very limited 😅 
Can someone help us with the reviews here ?

mcbarton:
@anutosh491 To see the space savings go here https://github.com/compiler-research/CppInterOp/actions/caches . Here is the example of one such space saving. Search 'bbbc20b4c96929d3b7d8f9a83c9987f9a154a569-refs/tags/v1.0-macOS-macos-13-clang-clang-13.x-patch-none' and you'll see it has the cache size 190mb on main . If you then go to the second page you'll see that the size of that cache for this PR (refs/pull/550/merge) is 170mb. If you did this one by one you'll see each of the 6 jobs gets a cache 20mb smaller, totaling the 120mb I state above. Hope this helps.

anutosh491 (approved):
Sure, I think the above explanation makes sense ! 
I'll let you merge this when you're happy.

PR #551, opened 2025-04-03
Improve formatting, readability and add newlines

aaronj0:


vgvassilev:
We generally don’t add white space change PRs. Is there a particular reason to add this one?

aaronj0:
> We generally don’t add white space change PRs. Is there a particular reason to add this one?

While a few changes are whitespaces (mostly deletions that won't show up on a blame), a lot of files are not newline terminated, which I think should be addressed. They make reading diffs harder, and is not good practice in terms of POSIX standards.

vgvassilev (approved):
LGTM, please remove the remaining new line changes.

Issue #541, opened 2025-03-26
[tests] : Improve test coverage


mcbarton:
### Description of bug

Over the last 12 months the test coverage has dropped from approximately 87% to approximately 74% .  The objective of the PR(s) which would solve this issue would be to increase the test coverage so that it is comparable to the level achieved previously, e.g. 85-90 %.

mcbarton:
Closing as I opened this and its too generic an issue to ever get resolved.

PR #513, opened 2025-03-03
Added undo command for CppInterOp

kr-2003:
# Description

Added ```Undo``` command for Clang-REPL. Reverts the last N operations performed by the interpreter.

## Type of change

- [x] New feature

## Testing
1. Tests in ```unittests/CppInterOp/FunctionReflectionTest.cpp```.
2. Demo - https://res.cloudinary.com/abhi9av/video/upload/v1740997292/wqgfbgqefrwpx2o7ko95.mp4

## Checklist

- [x] I have read the contribution guide recently

PR #553, opened 2025-04-04
Add error propagation to interpreter creation logic

anutosh491:
# Description

@Vipul-Cariappa and I were discussing a use case yesterday (especially while building with repl) and this is what we realized

1) The function responsible for creating `clang::Interpreter` is `createClangInterpreter` which can return a nullptr, for eg 
https://github.com/compiler-research/CppInterOp/blob/c3384fb9d784a3f1606b6fc20c283c1e48d9b14b/lib/Interpreter/Compatibility.h#L277

2) This is assigned to `inner` while the constructor for `compat::Interpreter` is called
https://github.com/compiler-research/CppInterOp/blob/c3384fb9d784a3f1606b6fc20c283c1e48d9b14b/lib/Interpreter/CppInterOpInterpreter.h#L153

which means `inner` can be `nullptr`

3) And then if we look at `CreateInterpreter` from the top, we're creating the `compat::Interpreter` here 

https://github.com/compiler-research/CppInterOp/blob/c3384fb9d784a3f1606b6fc20c283c1e48d9b14b/lib/Interpreter/CppInterOp.cpp#L2960

Which means `compat::Interpreter` can exist even when the underlying `clang::Interpreter` is not existing. So we can say that the Interpreter from the top is half broken

4) So if we try something like 
```
Cpp::CreateInterpreter(...)
Cpp::Parse(..)
```
We would hit the following finally
https://github.com/compiler-research/CppInterOp/blob/c3384fb9d784a3f1606b6fc20c283c1e48d9b14b/lib/Interpreter/CppInterOpInterpreter.h#L174-L176

And although the call is made to `compat::Interpreter`, we end up using the underlying `clang::Interpreter` **without checking if it is valid.**

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/lib/Interpreter/Compatibility.h
+++ b/lib/Interpreter/Compatibility.h
@@ -294,9 +294,11 @@ createClangInterpreter(std::vector<const char*>& args) {
     return nullptr;
   }
   if (CudaEnabled) {
-    if (auto Err = (*innerOrErr)->LoadDynamicLibrary("libcudart.so"))
+    if (auto Err = (*innerOrErr)->LoadDynamicLibrary("libcudart.so")) {
       llvm::logAllUnhandledErrors(std::move(Err), llvm::errs(),
                                   "Failed load libcudart.so runtime:");
+      return nullptr;
+    }
   }
 
   return std::move(*innerOrErr);
--- a/unittests/CppInterOp/CUDATest.cpp
+++ b/unittests/CppInterOp/CUDATest.cpp
@@ -13,7 +13,7 @@ static bool HasCudaSDK() {
     // FIXME: Enable this for cling.
     return false;
 #endif // CLANG_VERSION_MAJOR < 16
-    Cpp::CreateInterpreter({}, {"--cuda"});
+    if(!Cpp::CreateInterpreter({}, {"--cuda"})) return false;
     return Cpp::Declare("__global__ void test_func() {}"
                         "test_func<<<1,1>>>();") == 0;
   };
@@ -30,7 +30,7 @@ static bool HasCudaRuntime() {
     if (!HasCudaSDK())
       return false;
 
-    Cpp::CreateInterpreter({}, {"--cuda"});
+    if(!Cpp::CreateInterpreter({}, {"--cuda"})) return false;
     if (Cpp::Declare("__global__ void test_func() {}"
                      "test_func<<<1,1>>>();"))
       return false;
--- a/unittests/CppInterOp/CUDATest.cpp
+++ b/unittests/CppInterOp/CUDATest.cpp
@@ -13,7 +13,8 @@ static bool HasCudaSDK() {
     // FIXME: Enable this for cling.
     return false;
 #endif // CLANG_VERSION_MAJOR < 16
-    if(!Cpp::CreateInterpreter({}, {"--cuda"})) return false;
+    if (!Cpp::CreateInterpreter({}, {"--cuda"}))
+      return false;
     return Cpp::Declare("__global__ void test_func() {}"
                         "test_func<<<1,1>>>();") == 0;
   };
@@ -30,7 +31,8 @@ static bool HasCudaRuntime() {
     if (!HasCudaSDK())
       return false;
 
-    if(!Cpp::CreateInterpreter({}, {"--cuda"})) return false;
+    if (!Cpp::CreateInterpreter({}, {"--cuda"}))
+      return false;
     if (Cpp::Declare("__global__ void test_func() {}"
                      "test_func<<<1,1>>>();"))
       return false;
--- a/unittests/CppInterOp/InterpreterTest.cpp
+++ b/unittests/CppInterOp/InterpreterTest.cpp
@@ -167,6 +167,22 @@ TEST(InterpreterTest, CreateInterpreter) {
 #endif
 }
 
+#ifndef CPPINTEROP_USE_CLING
+TEST(InterpreterTest, CreateInterpreterCAPI) {
+  const char* argv[] = {"-std=c++17"};
+  auto *CXI = clang_createInterpreter(argv, 1);
+  auto CLI = clang_Interpreter_getClangInterpreter(CXI);
+  EXPECT_TRUE(CLI);
+  clang_Interpreter_dispose(CXI);
+}
+
+TEST(InterpreterTest, CreateInterpreterCAPIFailure) {
+  const char* argv[] = {"-fsyntax-only", "-Xclang", "-invalid-plugin"};
+  auto *CXI = clang_createInterpreter(argv, 3);
+  EXPECT_EQ(CXI, nullptr);
+}
+#endif
+
 #ifdef LLVM_BINARY_DIR
 TEST(InterpreterTest, DetectResourceDir) {
 #ifdef EMSCRIPTEN
--- a/unittests/CppInterOp/InterpreterTest.cpp
+++ b/unittests/CppInterOp/InterpreterTest.cpp
@@ -177,6 +177,9 @@ TEST(InterpreterTest, CreateInterpreterCAPI) {
 }
 
 TEST(InterpreterTest, CreateInterpreterCAPIFailure) {
+#ifdef _WIN32
+  GTEST_SKIP() << "Disabled on Windows. Needs fixing.";
+#endif
   const char* argv[] = {"-fsyntax-only", "-Xclang", "-invalid-plugin"};
   auto *CXI = clang_createInterpreter(argv, 3);
   EXPECT_EQ(CXI, nullptr);
--- a/lib/Interpreter/CXCppInterOp.cpp
+++ b/lib/Interpreter/CXCppInterOp.cpp
@@ -271,8 +271,15 @@ static inline compat::Interpreter* getInterpreter(const CXInterpreterImpl* I) {
 }
 
 CXInterpreter clang_createInterpreter(const char* const* argv, int argc) {
-  auto* I = new CXInterpreterImpl(); // NOLINT(*-owning-memory)
+  auto I = std::make_unique<CXInterpreterImpl>();
+#ifdef CPPINTEROP_USE_CLING
   I->Interp = std::make_unique<compat::Interpreter>(argc, argv);
+#else
+  I->Interp = compat::Interpreter::create(argc, argv);
+  if (!I->Interp) {
+    return nullptr;
+  }
+#endif
   // create a bridge between CXTranslationUnit and clang::Interpreter
   // auto AU = std::make_unique<ASTUnit>(false);
   // AU->FileMgr = I->Interp->getCompilerInstance().getFileManager();
@@ -281,7 +288,7 @@ CXInterpreter clang_createInterpreter(const char* const* argv, int argc) {
   // AU->Ctx = &I->Interp->getSema().getASTContext();
   // I->TU.reset(MakeCXTranslationUnit(static_cast<CIndexer*>(clang_createIndex(0,
   // 0)), AU));
-  return I;
+  return I.release();
 }
 
 CXInterpreter clang_createInterpreterFromRawPtr(TInterp_t I) {
--- a/lib/Interpreter/CppInterOp.cpp
+++ b/lib/Interpreter/CppInterOp.cpp
@@ -2957,7 +2957,15 @@ namespace Cpp {
                    std::back_inserter(ClingArgv),
                    [&](const std::string& str) { return str.c_str(); });
 
+#ifdef CPPINTEROP_USE_CLING
     auto I = new compat::Interpreter(ClingArgv.size(), &ClingArgv[0]);
+#else
+    auto Interp = compat::Interpreter::create(
+        static_cast<int>(ClingArgv.size()), ClingArgv.data());
+    if (!Interp)
+      return nullptr;
+    auto* I = Interp.release();
+#endif
 
     // Honor -mllvm.
     //
--- a/lib/Interpreter/CppInterOpInterpreter.h
+++ b/lib/Interpreter/CppInterOpInterpreter.h
@@ -25,6 +25,7 @@
 #if CLANG_VERSION_MAJOR >= 19
 #include "clang/Sema/Redeclaration.h"
 #endif
+#include "clang/Serialization/ModuleFileExtension.h"
 
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/SmallSet.h"
@@ -33,7 +34,11 @@
 #include "llvm/ExecutionEngine/Orc/LLJIT.h"
 #include "llvm/Support/Compiler.h"
 #include "llvm/Support/Error.h"
+#include "llvm/Support/raw_ostream.h"
 #include "llvm/Support/TargetSelect.h"
+
+#include <utility>
+#include <vector>
 
 namespace clang {
 class CompilerInstance;
--- a/lib/Interpreter/CppInterOpInterpreter.h
+++ b/lib/Interpreter/CppInterOpInterpreter.h
@@ -34,8 +34,8 @@
 #include "llvm/ExecutionEngine/Orc/LLJIT.h"
 #include "llvm/Support/Compiler.h"
 #include "llvm/Support/Error.h"
-#include "llvm/Support/raw_ostream.h"
 #include "llvm/Support/TargetSelect.h"
+#include "llvm/Support/raw_ostream.h"
 
 #include <utility>
 #include <vector>
--- a/lib/Interpreter/CppInterOpInterpreter.h
+++ b/lib/Interpreter/CppInterOpInterpreter.h
@@ -136,11 +136,14 @@ class Interpreter {
 private:
   std::unique_ptr<clang::Interpreter> inner;
 
+  Interpreter(std::unique_ptr<clang::Interpreter> CI) : inner(std::move(CI)) {}
+
 public:
-  Interpreter(int argc, const char* const* argv, const char* llvmdir = 0,
-              const std::vector<std::shared_ptr<clang::ModuleFileExtension>>&
-                  moduleExtensions = {},
-              void* extraLibHandle = 0, bool noRuntime = true) {
+  static std::unique_ptr<Interpreter>
+  create(int argc, const char* const* argv, const char* llvmdir = nullptr,
+         const std::vector<std::shared_ptr<clang::ModuleFileExtension>>&
+             moduleExtensions = {},
+         void* extraLibHandle = nullptr, bool noRuntime = true) {
     // Initialize all targets (required for device offloading)
     llvm::InitializeAllTargetInfos();
     llvm::InitializeAllTargets();
@@ -150,7 +153,13 @@ public:
     std::vector<const char*> vargs(argv + 1, argv + argc);
     vargs.push_back("-include");
     vargs.push_back("new");
-    inner = compat::createClangInterpreter(vargs);
+    auto CI = compat::createClangInterpreter(vargs);
+    if (!CI) {
+      llvm::errs() << "Interpreter creation failed\n";
+      return nullptr;
+    }
+
+    return std::unique_ptr<Interpreter>(new Interpreter(std::move(CI)));
   }
 
   ~Interpreter() {}


anutosh491:
Hence, we had a couple approaches to go ahead here

i) Introduce a isValid for clang::Interpreter and anytime we have a inner->func_call we use that (sounds a bit cumbersome checking for validity everytime)

ii) What Vipul and I liked was propagating the error all the way to that top, so that a half broken interpreter can't be put to use and a utility function like parse/declare would only work if CreateInterpreter doesn't return null along with an error message saying Interpreter creation failed.

lib/Interpreter/Compatibility.h
```diff
@@ -294,9 +294,11 @@ createClangInterpreter(std::vector<const char*>& args) {
     return nullptr;
   }
   if (CudaEnabled) {
-    if (auto Err = (*innerOrErr)->LoadDynamicLibrary("libcudart.so"))
+    if (auto Err = (*innerOrErr)->LoadDynamicLibrary("libcudart.so")) {
       llvm::logAllUnhandledErrors(std::move(Err), llvm::errs(),
                                   "Failed load libcudart.so runtime:");
+      return nullptr;
```
anutosh491, line 300: This was some inconsistency that I saw along the way. We return a nullptr whenever clang-repl uses ExitOnErr except this case (https://github.com/llvm/llvm-project/blob/main/clang/tools/clang-repl/ClangRepl.cpp#L212)

So clang-repl would just exit out ... through something like
```
(clang-repl-env-19) anutosh491@Anutoshs-MacBook-Air bin % ./clang-repl --cuda
clang-repl: dlopen(libcudart.so, 0x0009): tried: 'libcudart.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibcudart.so' (no such file), '/Users/anutosh491/micromamba/envs/clang-repl-env-19/lib/../lib/libcudart.so' (no such file), '/Users/anutosh491/micromamba/envs/clang-repl-env-19/lib/libcudart.so' (no such file), '/Users/anutosh491/micromamba/envs/clang-repl-env-19/bin/../lib/libcudart.so' (no such file), '/Users/anutosh491/micromamba/envs/clang-repl-env-19/bin/../lib/libcudart.so' (no such file), '/usr/lib/libcudart.so' (no such file, not in dyld cache), 'libcudart.so' (no such file), '/usr/local/lib/libcudart.so' (no such file), '/usr/lib/libcudart.so' (no such file, not in dyld cache)
```

But in our case we are continuing to use a half broken interpreter without the above working and we run declare on top of this 

https://github.com/compiler-research/CppInterOp/blob/c3384fb9d784a3f1606b6fc20c283c1e48d9b14b/unittests/CppInterOp/CUDATest.cpp#L17-L18

which is bound to not work as can be seen [here](https://github.com/compiler-research/CppInterOp/actions/runs/14242331470/job/39914920362#step:10:1001)

vgvassilev (approved):
Can we figure out how to write tests?

anutosh491:
> Can we figure out how to write tests?

Okay I think the codecov says that 3 patches are not being tested. All of them are directly based out of the changes in `CXCppInterOp.cpp`

I tried adding tests but was quick to realize that I am not really doing just justice as I might not be the best one to goto for adding tests for the C API (hence I reverted them and squashed the commits 😅)

cc @Gnimuc 

Would be great if you could help me out with the tests here. The PR description (https://github.com/compiler-research/CppInterOp/pull/553#issue-2971536179) should be good to get you started on the changed made in the C-API (the changes are not too concerning, simple ones !)

1) Maybe for the creation of the Intepreter we could start with something simple like 
```
#ifndef CPPINTEROP_USE_CLING
TEST(InterpreterTest, CreateInterpreterCAPI) {
 const char* argv[] = {"-std=c++17"};
 CXInterpreter CXI = clang_createInterpreter(argv, 1);
 EXPECT_TRUE(CXI);


}
#endif
```
2) For testing the failure while creating the interpreter (the case where we return `nullptr` from `clang_createInterpreter`), not sure what we can try. Maybe creating an non existent target or so I guess 
```
#ifndef CPPINTEROP_USE_CLING
TEST(InterpreterTest, CreateInterpreterCAPIFailure) {
  // This should return nullptr due to bad target triple
  const char* argv[] = {"-target", "nonexistent-unknown-target"};
  CXInterpreter CXI = clang_createInterpreter(argv, 2);
  EXPECT_EQ(CXI, nullptr);
}
#endif
```

anutosh491:
> I'll give it a look this weekend.

Thanks a lot for the help. This change ends up being critical now that we'll soon be moving to llvm 20 !

Gnimuc:
The C API `clang_createInterpreter` aims to provide a way to customize compiler arguments that are used to create the compiler instance, so it doesn't adopt the following setups provided in the C++ API. 

https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/lib/Interpreter/CppInterOp.cpp#L2924-L2958

It's a bit tricky to set up a successful test environment because we need to know the path of system headers for each test platform. (Although most of the arguments are excluded in the C API,  [`-include new`](https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/lib/Interpreter/CppInterOpInterpreter.h#L151-L152) is mandatory, which needs the correct path of system headers)

Gnimuc:
Given `auto* I = Cpp::CreateInterpreter();`, if we can find a way to retrieve the arguments used to create the interpreter, then  those arguments can be used to test the C API.

anutosh491:
Thanks for the reply @Gnimuc 

> BTW, I think the changes in the PR look good.

Also thanks for going through the changes ! Much appreciated :)

Hey @vgvassilev , do you think it's okay if we move ahead with this ?

As said above there's some discussion that needs to be done on testing the C API. And although some simple dummy tests can possibly be added as I wrote above, those won't probably be meaningful enough. So yeah the changes introduce **4 lines** of untested code in the C API but we can probably get to it in the future.

The main reason for this Pr is to help @mcbarton 's PR moving to llvm 20 (as his Pr should benefit with these changes after a rebase)

vgvassilev:
> The C API `clang_createInterpreter` aims to provide a way to customize compiler arguments that are used to create the interpreter, so it doesn't adopt the following setups provided in the C++ API.
> 
> https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/lib/Interpreter/CppInterOp.cpp#L2924-L2958
> 
> It's a bit tricky to set up a successful test environment because we need to know the path of system headers for each test platform. (Although most of the arguments are excluded in the C API, [`-include new`](https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/lib/Interpreter/CppInterOpInterpreter.h#L151-L152) is mandatory, which needs the correct path of system headers)

Can we make the new header optional?

anutosh491:
> Can we make the new header optional?

I think this should be possible isn't it !

The include new header would be required for tests like these 
https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/unittests/CppInterOp/FunctionReflectionTest.cpp#L1361-L1375

Now on line `1375` we have the `GetAllTopLevelDecls` which ends up calling our `CreateInterpreter`

https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/unittests/CppInterOp/Utils.cpp#L21-L23

So technically if we can improve `GetAllTopLevelDecls` to take care of custom args-> we can pass the same args to `CreateInterpreter` from there and whichever test needs whatever header include can use it. That way it is not mandatory to get that include for tests who wouldn't want to use it.

But yeah that would mean changing the function definition of the utility function `GetAllTopLevelDecls` overall :\ 
So probably should be done in subsequent PR !!!

mcbarton:
> Ahh the 3 failing checks are due to this
> 
> [compiler-research/xeus-cpp#286](https://github.com/compiler-research/xeus-cpp/pull/286)
> 
> Making the same change here too ! I think we should be good after that !

I think this change should be made as a separate PR, and this PR rebased on it, given this pinning change has nothing to do with the goal of this PR. I will approve and merge this pinning PR, as soon as I see its been put in.

> > Can we make the new header optional?
> 
> I think this should be possible isn't it !
> 
> The include new header would be required for tests like these
> 
> https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/unittests/CppInterOp/FunctionReflectionTest.cpp#L1361-L1375
> 
> Now on line `1375` we have the `GetAllTopLevelDecls` which ends up calling our `CreateInterpreter`
> 
> https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/unittests/CppInterOp/Utils.cpp#L21-L23
> 
> So technically if we can improve `GetAllTopLevelDecls` to take care of custom args-> we can pass the same args to `CreateInterpreter` from there and whichever test needs whatever header include can use it. That way it is not mandatory to get that include for tests who wouldn't want to use it.
> 
> But yeah that would mean changing the function definition of the utility function `GetAllTopLevelDecls` overall :\ So probably should be done in subsequent PR !!! (which I can try tomorrow)

If making this header option would allow this PR to be tested, then should probably be done first, and then this PR should be rebased this on it. Otherwise we risk tests never been made to cover these lines. I know its just a few lines, but its because due to not insisting all lines be covered by tests, we have ended up in a situation where CppInterOps coverage has dropped by approximately 13% in the last 12 months.

anutosh491:
Hi,

> I think this change should be made as a separate PR

I've taken care of this here https://github.com/compiler-research/CppInterOp/pull/559

> If making this header option would allow this PR to be tested, then should probably be done first, and then this PR should be rebased this on it. Otherwise we risk tests never been made to cover these lines. I know its just a few lines, but its because due to not insisting all lines be covered by tests, we have ended up in a situation where CppInterOps coverage has dropped by approximately 13% in the last 12 months (I made an issue about this here https://github.com/compiler-research/CppInterOp/issues/541).

Hmmm, I get it but shouldn't higher  priority be moving to LLVM 20 now (it was out on FEB 2 and its high time we move) . The purpose of this PR was to help with that. 

Also I am not a huge fan of nobody officially taking responsibility for covering the C API. It shouldn't have been added in the first place if that was the case !

Most of my focus since the past 2-3 weeks is around getting a good build (native/emscripten) for cppinterop around llvm 20 and technically this PR is also tested well on our C++ API. But for the C-API
i) I am not educated enough on the use case.
ii) I would really fall short on time to handle development/coverage there.

But this shouldn't slow any development on C++ API side is what I'm trying to convey.

I'm just going by Gnimuc 's word who added the C-API (he says "BTW, I think the changes in the PR look good." https://github.com/compiler-research/CppInterOp/pull/553#issuecomment-2799855918) so I am just hoping we would be in a good enough state to test it in the near future. But if something as simple as a "create interpreter" function is not tested in the C API do we really stop other work due to that ? I don't know.

Gnimuc:
> #ifndef CPPINTEROP_USE_CLING
> TEST(InterpreterTest, CreateInterpreterCAPIFailure) {
>   // This should return nullptr due to bad target triple
>   const char* argv[] = {"-target", "nonexistent-unknown-target"};
>   CXInterpreter CXI = clang_createInterpreter(argv, 2);
>   EXPECT_EQ(CXI, nullptr);
> }
> #endif

@anutosh491 Maybe it's a good idea to add this failure test for now. It will improve the coverage a bit.

Gnimuc:
> > The C API `clang_createInterpreter` aims to provide a way to customize compiler arguments that are used to create the interpreter, so it doesn't adopt the following setups provided in the C++ API.
> > https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/lib/Interpreter/CppInterOp.cpp#L2924-L2958
> > 
> > It's a bit tricky to set up a successful test environment because we need to know the path of system headers for each test platform. (Although most of the arguments are excluded in the C API, [`-include new`](https://github.com/compiler-research/CppInterOp/blob/98a57b4d31e487fb64bce223431d5249a369ff86/lib/Interpreter/CppInterOpInterpreter.h#L151-L152) is mandatory, which needs the correct path of system headers)
> 
> Can we make the new header optional?

I'm wondering what's the original purpose of adding this header. Is it related to the hard-coded code in `clang::Interpreter::create`?  An operator new trampoline function is declared there.

anutosh491:
Hey @Gnimuc, we just dropped the header through #https://github.com/compiler-research/CppInterOp/pull/561

Does this put you in a better position to add some basic tests for `clang_createInterpreter` or other related functions ?

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -166,6 +166,25 @@
 #endif
 }
 
+#ifndef CPPINTEROP_USE_CLING
+TEST(InterpreterTest, CreateInterpreterCAPI) {
+  const char* argv[] = {"-std=c++17"};
+  CXInterpreter CXI = clang_createInterpreter(argv, 1);
+  EXPECT_TRUE(CXI);
+
+  auto CLI = clang_Interpreter_getClangInterpreter(CXI);
```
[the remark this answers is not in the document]
anutosh491: I see "auto CLI" being put to use through while getting access to the Interpreter. So hope that's fine.

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -167,6 +167,22 @@ TEST(InterpreterTest, CreateInterpreter) {
...
+TEST(InterpreterTest, CreateInterpreterCAPI) {
+  const char* argv[] = {"-std=c++17"};
+  auto *CXI = clang_createInterpreter(argv, 1);
+  auto CLI = clang_Interpreter_getClangInterpreter(CXI);
+  EXPECT_TRUE(CLI);
+  clang_Interpreter_dispose(CXI);
+}
+
+TEST(InterpreterTest, CreateInterpreterCAPIFailure) {
+  const char* argv[] = {"-fsyntax-only", "-Xclang", "-invalid-plugin"};
```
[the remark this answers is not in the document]
anutosh491: https://github.com/compiler-research/CppInterOp/pull/553/files#r2043868646

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -167,6 +167,22 @@ TEST(InterpreterTest, CreateInterpreter) {
 #endif
 }
 
+#ifndef CPPINTEROP_USE_CLING
```
mcbarton, line 170: Is there a reason why we are not doing this test for Cling?

anutosh491: No clue,

I see no C-API tests done through cling and that's what I am trying to replicate. Maybe @Gnimuc can educate us here.

kr-2003:
> Hey @Gnimuc @vgvassilev
> 
> I have added some simple tests and ensured codecov doesn't complain about changes atleast with regard to this PR.
> 
> All tests pass no clue why the windows CI fail (don't look like a direct outcome of my tests)
> 
> image
> 
> Shall try looking into it for a bit (or shall skip If I am helpless cause I don't have a windows machine :\ )

Tests which write something on screen fails on Windows. Here, in this comment @mcbarton explained this:
https://github.com/compiler-research/CppInterOp/pull/513#issuecomment-2735698456

mcbarton:
> > Hey @Gnimuc @vgvassilev
> > I have added some simple tests and ensured codecov doesn't complain about changes atleast with regard to this PR.
> > All tests pass no clue why the windows CI fail (don't look like a direct outcome of my tests)
> > image
> > Shall try looking into it for a bit (or shall skip If I am helpless cause I don't have a windows machine :\ )
> 
> Tests which write something on screen fails on Windows. Here, in this comment @mcbarton explained this: [#513 (comment)](https://github.com/compiler-research/CppInterOp/pull/513#issuecomment-2735698456)

To refine this comment from before. I think anything which writes an error message in Windows, will be picked up as an error in the overall running of the tests. Tests which seem to write a warning message are ok. I believe it is ok for you skip this test on Windows for this reason.

kr-2003:
> To refine this comment from before. I think anything which makes the interpreter write an error message in Windows, will be picked up as an error in the overall running of the tests. Tests which seem to write a warning message are ok. I believe it is ok for you skip this test on Windows for this reason.

Alternatively, we can suppress the error message and retain this test on windows too.
```

TEST(InterpreterTest, CreateInterpreterCAPIFailure) {
#ifdef CPPINTEROP_USE_CLING
  GTEST_SKIP() << "C API is not available in this build";
#endif
  const char* argv[] = {"-fsyntax-only", "-Xclang", "-invalid-plugin"};
  testing::internal::CaptureStderr(); // Suppress error messages
  auto *CXI = clang_createInterpreter(argv, 3);
  std::string capturedStderr = testing::internal::GetCapturedStderr();
  EXPECT_EQ(CXI, nullptr);
}
```

mcbarton:
> > To refine this comment from before. I think anything which makes the interpreter write an error message in Windows, will be picked up as an error in the overall running of the tests. Tests which seem to write a warning message are ok. I believe it is ok for you skip this test on Windows for this reason.
> 
> Alternatively, we can suppress the error message and retain this test on windows too. It should pass now.
> 
> ```
> 
> TEST(InterpreterTest, CreateInterpreterCAPIFailure) {
> #ifdef CPPINTEROP_USE_CLING
>   GTEST_SKIP() << "C API is not available in this build";
> #endif
>   const char* argv[] = {"-fsyntax-only", "-Xclang", "-invalid-plugin"};
>   testing::internal::CaptureStderr(); // Suppress error messages
>   auto *CXI = clang_createInterpreter(argv, 3);
>   std::string capturedStderr = testing::internal::GetCapturedStderr();
>   EXPECT_EQ(CXI, nullptr);
> }
> ```

Not entirely sure testing::internal::CaptureStderr() will work to resolve this issue. Its probably better to skip for now on Windows. We do however want to use GTEST_SKIP(), instead of the way its done currently in the PR.

anutosh491:
> Its probably better to skip for now on Windows

Yeah I think we can just skip it for now.

anutosh491:
I think this is ready from my side now that all the checks pass.

unittests/CppInterOp/InterpreterTest.cpp
Gnimuc: The original purpose of adding the C API is to support multiple instances. REPL and Cling are two backends for creating compiler instances. REPL is based on `clang::Interpreter` from the upstream llvm-project. 

I think there is no plan to upstream Cling to the llvm-project, and I didn't check Cling's compatibility when I added those C-API tests.

vgvassilev: You can build, CppInterOp by providing cling and it will embed cling in it. If that’s not too much of work maybe we can support that, shouldn’t hurt.

anutosh491: Can we have a dedicated PR for that ? 

I'll say the purpose of this PR is complete now.

>  If that’s not too much of work maybe we can support that, shouldn’t hurt.

@Gnimuc would you like to give this an attempt ?

unittests/CppInterOp/CUDATest.cpp
```diff
@@ -30,7 +30,7 @@ static bool HasCudaRuntime() {
     if (!HasCudaSDK())
       return false;
 
-    Cpp::CreateInterpreter({}, {"--cuda"});
+    if(!Cpp::CreateInterpreter({}, {"--cuda"})) return false;
```
vgvassilev, line 33:
```suggestion
    if (!Cpp::CreateInterpreter({}, {"--cuda"})) 
      return false;
```

why clang-format was happy with this? Please fix the other occurrences too.

vgvassilev (approved):
Please also address the missing includes suggested by clang-tidy before merging.

unittests/CppInterOp/CUDATest.cpp
anutosh491: Oops completely missed this (not sure I saw a complaint regarding this !

Shall address it, thanks

anutosh491:
Addressed all reviews. I think this should be ready now. Thanks everyone for the reviews. Squashing and Merging!

PR #558, opened 2025-04-09
Add tests for nested constructor calls via `Cpp::Construct`

aaronj0:
Adds tests for nested constructor calls through `Cpp::Construct`


vgvassilev (approved):
LGTM!

Issue #500, opened 2025-02-11
[Feature Request]:  Add automated testing of CppInterOp in a browser/c++ tutorial demo of CppInterOp


mcbarton:
### Description of new feature

Now that it has been determined how to do automated emscripten tests of CppInterOp through the use of node, the next step will be to do automated tests within a browser environment. My recommendation is the creation of a c++ tutorial which can be run as part of the ci, which utilises CppInterOp. This will allow us to know exactly what c++ can be executed within a wasm environment by CppInterOp. This c++ tutorial can also be of us to others, wishing to learn c++.

PR #569, opened 2025-05-01
Run Emscripten tests in a browser

mcbarton:
# Description

This PR adds the ability to run the Emscripten tests in a browser. This partially contributes to the goal of https://github.com/compiler-research/CppInterOp/issues/500 , which is to run a c++ tutorial html file utilising CppInterOp, as part of automated testing.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -599,6 +599,93 @@ jobs:
         fi
 
         emmake make -j ${{ env.ncpus }} check-cppinterop
+        cd ./unittests/CppInterOp/
+        # Fresh install browsers, and run Emscripten tests in them
+        # This is to match the Emscripten build instructions, where
+        # we run in a fresh browser, to stop any extra installed
+        # stuff interferring with the running of the tests
+        # Explaination of options for emrun
+        # --browser (name of browser on path)
+        # --kill_exit makes it so that when emrun finishes,
+        # that the headless browser we create is killed along with it
+        # --timeout 60 is such that emrun is killed after 60 seconds if
+        # still running. emrun should have finished long before then,
+        # so if it is still running, something went wrong (such as a test
+        # which crashed the html file). This will cause the ci to fail,
+        # as a non 0 value of will be returned.
+        # In the case of Chrome we have the extra --no-sandbox flag, as on
+        # Ubuntu Chrome will refuse to run otherwise, as it expects to have
+        # been installed with admin privileges. This flag allows it to run
+        # in userspace.
+        os="${{ matrix.os }}"
+        if [[ "${os}" == "macos"* ]]; then
+          # Install Firefox
+          wget "https://download.mozilla.org/?product=firefox-latest&os=osx&lang=en-US" -O Firefox-latest.dmg
+          hdiutil attach Firefox-latest.dmg
+          cp -r /Volumes/Firefox/Firefox.app $PWD
+          hdiutil detach /Volumes/Firefox
+          cd ./Firefox.app/Contents/MacOS/
+          export PATH="$PWD:$PATH"
+          cd -
+
+          # Install Google Chrome
+          wget https://dl.google.com/chrome/mac/stable/accept_tos%3Dhttps%253A%252F%252Fwww.google.com%252Fintl%252Fen_ph%252Fchrome%252Fterms%252F%26_and_accept_tos%3Dhttps%253A%252F%252Fpolicies.google.com%252Fterms/googlechrome.pkg
+          pkgutil --expand-full googlechrome.pkg google-chrome
+          cd ./google-chrome/GoogleChrome.pkg/Payload/Google\ Chrome.app/Contents/MacOS/
+          export PATH="$PWD:$PATH"
+          cd -
+
+          # Run tests in browsers
+          echo "Running CppInterOpTests in Firefox"
+          emrun --browser="firefox" --kill_exit --timeout 60 --browser-args="--headless"  CppInterOpTests.html
+          echo "Running DynamicLibraryManagerTests in Firefox"
+          emrun --browser="firefox" --kill_exit --timeout 60 --browser-args="--headless"  DynamicLibraryManagerTests.html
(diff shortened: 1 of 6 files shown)


mcbarton:
Running the tests makes use of emrun which comes with Emscripten. The documentation for Emscripten can be found here https://emscripten.org/docs/compiling/Running-html-files-with-emrun.html

.github/workflows/emscripten.yml
```diff
@@ -542,6 +542,77 @@ jobs:
         fi
 
         emmake make -j ${{ env.ncpus }} check-cppinterop
+        cd ./unittests/CppInterOp/
+        # Install browsers, and run Emscripten tests in them
+        os="${{ matrix.os }}"
+        if [[ "${os}" == "macos"* ]]; then
```
mcbarton, line 548: The reason we install a fresh install of Firefox and Chome is to follow what is done in the Emscripten repo (for example here with Firefox https://github.com/emscripten-core/emscripten/blob/38e16464cffc8f886364fe4919a712131a4c9456/.circleci/config.yml#L363 ). It has the added advantage by not using the users installed versions of these browsers (if they have them installed), that there are no add ons, etc that could be causing issues with the tests running. I have made the instructions, such that a developer without admin privileges on there machine can install them.

.github/workflows/emscripten.yml
```diff
@@ -542,6 +542,77 @@ jobs:
...
+          # Install Google Chrome
+          wget https://dl.google.com/chrome/mac/stable/accept_tos%3Dhttps%253A%252F%252Fwww.google.com%252Fintl%252Fen_ph%252Fchrome%252Fterms%252F%26_and_accept_tos%3Dhttps%253A%252F%252Fpolicies.google.com%252Fterms/googlechrome.pkg
+          pkgutil --expand-full googlechrome.pkg google-chrome
+          cd ./google-chrome/GoogleChrome.pkg/Payload/Google\ Chrome.app/Contents/MacOS/
+          export PATH="$PWD:$PATH"
+          cd -
+
+          # Run tests in browsers
+          echo "Running CppInterOpTests in Firefox"
+          emrun --browser="firefox" --kill_exit --timeout 60 --browser-args="--headless"  CppInterOpTests.html
```
mcbarton, line 567: emrun allows us to run the created html file

`--kill_exit` makes it so that when emrun finishes, that the headless browser we create is killed along with it
`--timeout 60` is such that emrun is killed after 60 seconds if still running. emrun should have finished long before then, so if it is still running, something went wrong (such as a test which crashed the html file). This will cause the ci to fail, as a non 0 value of will be returned.

.github/workflows/emscripten.yml
```diff
@@ -542,6 +542,77 @@ jobs:
...
+          echo "Running DynamicLibraryManagerTests in Firefox"
+          emrun --browser="firefox" --kill_exit --timeout 60 --browser-args="--headless"  DynamicLibraryManagerTests.html
+          echo "Running CppInterOpTests in Google Chrome"
+          emrun --browser="Google Chrome" --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  CppInterOpTests.html
```
mcbarton, line 571: In the case of Chrome we have the extra `--no-sandbox` flag, as on Ubuntu Chrome will refuse to run otherwise, as it expects to have been installed with admin privileges. This flag allows it to run in userspace.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -50,6 +51,7 @@ if(EMSCRIPTEN)
     PUBLIC "SHELL: -s STACK_SIZE=32mb"
     PUBLIC "SHELL: -s INITIAL_MEMORY=128mb"
     PUBLIC "SHELL: --preload-file ${SYSROOT_PATH}/include@/include"
+    PUBLIC "SHELL: --emrun"
```
mcbarton, line 54:
```
Rebuild your Emscripten application and add the --emrun [linker flag](https://emscripten.org/docs/tools_reference/emcc.html#emcc-emrun).

This flag injects code into the generated [Module object](https://emscripten.org/docs/api_reference/module.html#module) to enable capture of stdout, stderr and exit().

Note

If you skip this step, you can still run any .html file with emrun, but the capture will not work.
```

The above is taken from emruns documentation https://emscripten.org/docs/compiling/Running-html-files-with-emrun.html

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 if (EMSCRIPTEN)
+  set(CMAKE_EXECUTABLE_SUFFIX ".html")
```
mcbarton, line 2: So that we create a html file, as well as the javascript file.

unittests/CMakeLists.txt
```diff
@@ -34,7 +34,11 @@ if(WIN32)
 else()
   target_link_libraries(${name} PRIVATE ${ARG_LIBRARIES} ${gtest_libs} ${link_pthreads_lib})
 endif()
+if(EMSCRIPTEN)
+  add_test(NAME cppinterop-${name} COMMAND $ENV{EMSDK_NODE} ${name}.js)
```
mcbarton, line 38: Without this cmake will try and get node to run the html file. This guarantees that it runs the js file, and uses emsdks node.

.github/workflows/emscripten.yml
vgvassilev: Can you capture these comments as comments in the file?

vgvassilev (approved):
LGTM. Perhaps worth waiting for @anutosh491 to share the blame ;)

vgvassilev:
Can you squash into a single commit?

mcbarton:
Will squash into a single commit as soon as I get to work in about 2 hours

mcbarton:
> Can you squash into a single commit?

All commits have now been squashed into one.

PR #571, opened 2025-05-05
fix codegen for operator wrapper

Vipul-Cariappa:
# Description

Generator operations instead of function calls similar to upstream

## Fixes # (issue)

fixes `test_cpp11features.py::TestCPP11FEATURES::test11_chrono` in cppyy

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist


aaronj0 (approved):
LGTM! This fixes the `std::chrono` test case and functionally brings the wrapper closer to upstream cppyy. The expression `!FD->isOverloadedOperator() || (FD->getOverloadedOperator() == clang::OO_Call)` can be stored in a flag to improve readability, feel free to merge once that is addressed (and the bots pass)

lib/Interpreter/CppInterOp.cpp
```diff
@@ -1963,8 +1963,12 @@ namespace Cpp {
       // Same applies with member methods which seem to cause parse failures
       // even when we supply the object parameter. Therefore we only use it in
       // cases where we know it works and set this variable to true when we do.
+
+      bool op_flag = !FD->isOverloadedOperator() ||
```
vgvassilev, line 1967: Can you add a comment what is the intent with the `op_flag`. To filter out overloaded operators that are not call operators?

vgvassilev (commented):
We should add a relevant test.

aaronj0 (approved):
LGTM! I believe all comments are addressed

aaronj0:
A reminder to squash into a single commit before merging :)

PR #572, opened 2025-05-11
Fix a typo in the `TOperator` test

Gnimuc:
We may need to find a way to catch this.


Vipul-Cariappa:
I cannot see a codecov bot comment. Is it not running?

Vipul-Cariappa (commented):
The change being done here is the `self` argument. Using both `object` and `toperator` is acceptable. And in both cases we are reusing the variable declared before.
I would say `object` is better as it conveys the correct usage.

Gnimuc:
> I would say `object` is better as it conveys the correct usage.

Could you elaborate on the correctness of using `object`? `object` has been destructed in another test case above.

mcbarton:
> I cannot see a codecov bot comment. Is it not running?

@Vipul-Cariappa This is due to the accidentally removal of the coverage job from the ci. This PR fixes it https://github.com/compiler-research/CppInterOp/pull/577 . Once this PR is in, rebase this PR and you should see a coverage comment. Feel free to review and approval the PR, if it hasn't already been done, when you see this message.

PR #581, opened 2025-05-19
Remove cling 1.0 support + Make llvm 16 minimum version

mcbarton:
# Description

Please include a summary of changes, motivation and context for this PR.

This PR will remove support for cling 1.0 from CppInterOp, and make the minimum version of llvm supported as 16. 

Fixes # (issue)

## Type of change

Please tick all options which are relevant.

- [ ] Bug fix
- [ ] New feature
- [ ] Requires documentation updates

## Testing

Please describe the test(s) that you added and ran to verify your changes.

## Checklist

- [x] I have read the contribution guide recently

PR #589, opened 2025-05-22
[interpreter] Drop IncrementalExecutor offsets to obtain LLJIT for Cling

aaronj0:
This patch drops pointer arithmetic used to obtain the `llvm::orc::LLJIT` from older Cling versions. This would make for safer usage of the `cling::Interpreter` and good to have for the 1.7.0 stable release that ROOT will move to.

### DIFF
--- a/lib/Interpreter/Compatibility.h
+++ b/lib/Interpreter/Compatibility.h
@@ -104,43 +104,13 @@ inline void maybeMangleDeclName(const clang::GlobalDecl& GD,
   cling::utils::Analyze::maybeMangleDeclName(GD, mangledName);
 }
 
-/// For Cling <= LLVM 16, this is a horrible hack obtaining the private
-/// llvm::orc::LLJIT by computing the object offsets in the cling::Interpreter
-/// instance(IncrementalExecutor): sizeof (m_Opts) + sizeof(m_LLVMContext). The
-/// IncrementalJIT and JIT itself have an offset of 0 as the first datamember.
-///
-/// The getExecutionEngine() interface has been added for Cling based on LLVM
-/// >=18 and should be used in future releases.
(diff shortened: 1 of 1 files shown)


mcbarton:
@aaronj0 When I was looking at this in my PR to drop cling 1.0 support I was planning to replace every occurance of compat::getExecutionEngine in the code. I just never got around to doing it. Maybe you can do it here?

aaronj0:
> @aaronj0 Given v1.7.0 just got released without this, do you mind me just doing this patch as part of my PR as part of my effort to drop cling 1.0 support here #581 ?

I prefer to land this here as an atomic change...


> @aaronj0 When I was looking at this in my PR to drop cling 1.0 support I was planning to replace every occurance of compat::getExecutionEngine in the code (given its not just a return statement). I just never got around to doing it. Maybe you can do it here?

hmm that doesn't make sense to me:

In a function that accepts a `compat::Interpreter` : https://github.com/compiler-research/CppInterOp/blob/392b754c566f96fb2df9ae270ecdfd2e0454961f/lib/Interpreter/CppInterOp.cpp#L3160-L3162

That uses this API:

https://github.com/compiler-research/CppInterOp/blob/392b754c566f96fb2df9ae270ecdfd2e0454961f/lib/Interpreter/CppInterOp.cpp#L3212-L3217

we need the abstraction anyway since clang-repl requires extra handling:

https://github.com/compiler-research/CppInterOp/blob/392b754c566f96fb2df9ae270ecdfd2e0454961f/lib/Interpreter/Compatibility.h#L342-L345

Another motivation is also to keep the `compat::` abstraction layer consistent, and perform operations defined under `compat` to `compat::Interpreter`

Vipul-Cariappa (approved):

PR #591, opened 2025-05-25
[WIP] Move CppInterOp.h into its own location; issue a warning to help users.

vgvassilev:
Currently we have the `CppInterOp.h` in `clang/Interpreter/` which makes people think that it is a clang header file. This patch moves the header into its own folder to avoid confusion.

This patch also adds a warning for people to update their header files. That's technically not a breaking change and can land any time. The next patch will rename `libclangCppInterOp` into `libCppInterOp` which would be technically a breaking change and will be done in v2.0.

This would be a good opportunity to fix stylistic changes without bloating the history such as the header preamble, indentation and such.


anutosh491:
Thanks a lot for this PR.

vgvassilev:
I cannot explain the windows failures. It might be that it has to do with some leftovers from the rename locally after the PR gets rebased on the bots... I'd prefer to merge this and debug later if necessary.

PR #600, opened 2025-05-28
Fix Windows x86 badge

mcbarton:
# Description

Currently for MacOS and Ubuntu we have separate badges labelled x86 and arm. For Windows x86 we do not have x86 in the badge name, but we do have arm for the Windows arm badge. This PR fixes that.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/Windows.yml
+++ b/.github/workflows/Windows.yml
@@ -1,4 +1,4 @@
-name: Windows
+name: Windows-x86
 on:
   pull_request:
     branches: [main]


vgvassilev:
> > > > These should be merged into a single workflow that shows up in the CI like in Clad?
> > > 
> > > 
> > > @vgvassilev With a single workflow you cannot have the badges for the individual target systems, and with them separate you can manually reactivate a smaller portion of your workflows if something fails.
> > 
> > 
> > I find this very confusing because first we build some stuff which appears green; then all steps are waiting for the building steps to finish; and then we go on with more. I never really know what's going on. I understand that from integration perspective that's not an issue but from developers' and reviewers' perspective this is quite bad.
> > The fine-grained badge level should not be relevant generally because the bots should be always green ;)
> > > For example you don't want to rerun all the Windows runners if your investigating a failing Ubuntu job in a tmate session.
> > 
> > 
> > We don't need to rerun all bots to get a tmate session for clad, too.
> 
> If you really want this to happen, then I can look into it, once all the reusable actions work is finished, but its unlikely I can integrate the Emscripten and native build cis together without some very hacky and hard to maintain code (the reason the Emscripten workflow was separated out to begin with was because it led to some confusing ci stuff).

I am fine to keep emscripten in a separate mode -- it is still a tricky platform. For the rest if we can do it in a single pass like in clad would be great. That is, I mean to have a "Main" action which renders in the CI as a single end-to-end step where if say something is wrong with `osx15-arm-clang-clang-repl-17` we immediately see it. If we can do a better code reuse that the Clad work -- it would be great...

PR #607, opened 2025-05-29
Stack the created interpreters.

### LINKED ISSUE #302: Support multiple interpreter instance?
https://github.com/compiler-research/CppInterOp/blob/ecbffafe0016022e40b7dea6e7197ee062dde38f/lib/Interpreter/CppInterOp.cpp#L58-L73

The interpreter is global and unique at the moment. However, clients do not always agree on the same compiler configuration. I think we need to add more APIs that take `compat::Interpreter` as argument. For example, 

```
const char* GetResourceDirFromInterpreter(const compat::Interpreter* interp) {
    return interp->getCI()->getHeaderSearchOpts().ResourceDir.c_str();
}
```

vgvassilev:
This patch offers an API to switch back and forth different interpreters which can come at handy in several occasions including research into implementing threading.

Fixes #302.

### DIFF
--- a/include/CppInterOp/CppInterOp.h
+++ b/include/CppInterOp/CppInterOp.h
@@ -595,14 +595,25 @@ CPPINTEROP_API void GetOperator(TCppScope_t scope, Operator op,
                                 std::vector<TCppFunction_t>& operators,
                                 OperatorArity kind = kBoth);
 
-/// Creates an instance of the interpreter we need for the various interop
-/// services.
+/// Creates an owned instance of the interpreter we need for the various interop
+/// services and pushes it onto a stack.
 ///\param[in] Args - the list of arguments for interpreter constructor.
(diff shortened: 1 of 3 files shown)


vgvassilev:
@aaronj0 and @Vipul-Cariappa, do you understand the cppyy assertion for this PR?

Vipul-Cariappa:
> @aaronj0 and @Vipul-Cariappa, do you understand the cppyy assertion for this PR?

@vgvassilev, you want to look at https://github.com/compiler-research/cppyy-backend/blob/e1068109110344b63b1546c853862db9632ec737/clingwrapper/src/clingwrapper.cxx#L182C1-L212C10

We first check if an interpreter already exists; if so, we just continue using it. Otherwise, create a new interpreter.
This is required for Xeus-CPP Cppyy integration. Because xeus-cpp would create the interpreter at the start of the kernel, then when a cell with `%%python` is executed, cppyy is loaded. It can not create a new interpreter at that point in time, because doing so will lose all the definitions declared so far.

PR #612, opened 2025-06-02
fix: codegen for nested namespace with using

Vipul-Cariappa:
Enables 3 tests in cppyy.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -518,7 +518,13 @@ std::string GetQualifiedCompleteName(TCppType_t klass) {
     if (auto* TD = llvm::dyn_cast<TagDecl>(ND)) {
       std::string type_name;
       QualType QT = C.getTagDeclType(TD);
-      QT.getAsStringInternal(type_name, C.getPrintingPolicy());
+      PrintingPolicy PP = C.getPrintingPolicy();
+      PP.FullyQualifiedName = true;
+      PP.SuppressUnwrittenScope = true;
+#if CLANG_VERSION_MAJOR > 16
+      PP.SuppressElaboration = true;
+#endif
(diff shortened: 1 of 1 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -518,7 +518,13 @@ std::string GetQualifiedCompleteName(TCppType_t klass) {
     if (auto* TD = llvm::dyn_cast<TagDecl>(ND)) {
       std::string type_name;
       QualType QT = C.getTagDeclType(TD);
-      QT.getAsStringInternal(type_name, C.getPrintingPolicy());
+      PrintingPolicy PP = C.getPrintingPolicy();
+      PP.FullyQualifiedName = true;
```
vgvassilev, line 522: Can we make sure codecov is covered?

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -1733,6 +1733,53 @@ TEST(FunctionReflectionTest, GetFunctionCallWrapper) {
       Cpp::BestOverloadFunctionMatch(operators, {}, {K1, K2});
   auto chrono_op_fn_callable = Cpp::MakeFunctionCallable(kop);
   EXPECT_EQ(chrono_op_fn_callable.getKind(), Cpp::JitCall::kGenericCall);
+
+  Interp->process(R"(
+    #include <tuple>
```
vgvassilev: Can we avoid the inclusion of the header and craft our own make_tuple-like routine? That should avoid future problems when CppInterOp tests are ran without installed stl.

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: This might be a problem with codecov, because the line above and below is covered, and there is no conditional either.

PR #620, opened 2025-06-09
[ci] fix workflow names

aaronj0:
This PR should fix the issue of actions appearing as paths instead of names on the Github UI
The interpreter mode is now dynamically inserted (Cling/Clang-REPL)  based on the config

before:
Screenshot from 2025-06-09 22-24-12

after:

Screenshot from 2025-06-09 22-22-47

### DIFF
--- a/.github/workflows/MacOS-arm.yml
+++ b/.github/workflows/MacOS-arm.yml
@@ -164,8 +164,8 @@ jobs:
         brew install gnu-sed
         pip install distro pytest
 
-
-    - uses: ./.github/actions/Build_LLVM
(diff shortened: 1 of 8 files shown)


mcbarton:
@aaronj0 This is quite a nice change. Could you make it add the cling version or clang repl/llvm version to the action name?

I know the purpose of this PR is to do with naming the reusable actions, but it might be nice to dynamically name the restore and save cache action name, based on whether its cling or clang-repl too.

mcbarton (approved):
LGTM! Thanks @aaronj0 . Merge once the ci passes.

PR #623, opened 2025-06-10
fix ambiguity in ScopeReflection `GetAllCppNames` unittest

aaronj0:
fixes the following warnings:

```
1: In file included from <<< inputs >>>:1:
1: input_line_1:5:15: warning: direct base 'A' is inaccessible due to ambiguity:
1:     class D -> A
1:     class D -> C -> A [-Winaccessible-base]
1:     5 |     class D : public A, public B, public C { int d; };
1:       |               ^~~~~~~~
1: input_line_1:5:25: warning: direct base 'B' is inaccessible due to ambiguity:
1:     class D -> B
1:     class D -> C -> B [-Winaccessible-base]
1:     5 |     class D : public A, public B, public C { int d; };
1:       |                         ^~~~~~~~
1: input_line_1:10:17: warning: direct base 'A' is inaccessible due to ambiguity:
1:     class N::D -> A
1:     class N::D -> C -> A [-Winaccessible-base]
1:    10 |       class D : public A, public B, public C { int d; };
1:       |                 ^~~~~~~~
1: input_line_1:10:27: warning: direct base 'B' is inaccessible due to ambiguity:
1:     class N::D -> B
1:     class N::D -> C -> B [-Winaccessible-base]
1:    10 |       class D : public A, public B, public C { int d; };
1:       |                           ^~~~~~~~
```

### DIFF
--- a/unittests/CppInterOp/ScopeReflectionTest.cpp
+++ b/unittests/CppInterOp/ScopeReflectionTest.cpp
@@ -789,7 +789,10 @@ TEST(ScopeReflectionTest, GetAllCppNames) {
     }
   )";
 
-  GetAllTopLevelDecls(code, Decls);
+  std::vector<const char*> interpreter_args = {"-Wno-inaccessible-base"};
+
+  GetAllTopLevelDecls(code, Decls, /*filter_implicitGenerated=*/false,
+                      interpreter_args);
 
   auto test_get_all_cpp_names =
       [](Decl* D, const std::vector<std::string>& truth_names) {


unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -777,12 +777,12 @@ TEST(ScopeReflectionTest, GetAllCppNames) {
...
     class B { int b; };
     class C : public A, public B { int c; };
-    class D : public A, public B, public C { int d; };
+    class D : public C { int d; };
     namespace N {
       class A { int a; };
       class B { int b; };
       class C : public A, public B { int c; };
-      class D : public A, public B, public C { int d; };
+      class D : public C { int d; };
```
vgvassilev: This is on purpose to make sure what we find. Can you suppress the diagnostic by calling `CreateInterpreter` with the appropriate flags?

vgvassilev:
> > More broadly, we can add to the default interpreter that's being created a special diagnostic consumer that complains if there are diagnostics that are not handled... We try to do something similar in https://github.com/llvm/llvm-project/blob/937be177528de156922c1b5f6cab08ba3009dbf2/clang/unittests/Interpreter/InterpreterTest.cpp#L125
> > LGTM!
> 
> Makes sense, will follow up with that at some point.

Maybe open an issue to track this.

PR #630, opened 2025-06-13
improve the CppInterOp.h include directory warning

aaronj0:
We missed closing angular brackets in the warning for the new header location.


Vipul-Cariappa:
Why is the test failing? I don't think it is due to the changes in this PR.

PR #635, opened 2025-06-17
Adds Out-Of-Process JIT execution for CppInterOp

kr-2003:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

## Testing

## Checklist


.github/actions/Build_LLVM/action.yml
```diff
@@ -49,6 +49,10 @@ runs:
             git apply -v ../patches/llvm/clang${{ matrix.clang-runtime }}-*.patch
             echo "Apply clang${{ matrix.clang-runtime }}-*.patch patches:"
           fi
+          if [[ "${{ matrix.oop-jit }}" == "On" ]]; then
+            git apply -v ../patches/llvm/out-of-process-jit-execution.patch
```
Vipul-Cariappa: Prefix/Suffix the LLVM version.

.github/workflows/main.yml
```diff
@@ -29,17 +29,19 @@ jobs:
             clang-runtime: '20'
             cling: Off
             cppyy: Off
-            llvm_enable_projects: "clang"
+            llvm_enable_projects: "clang;compiler-rt"
             llvm_targets_to_build: "host;NVPTX"
             coverage: true
+            oop-jit: "On"
```
Vipul-Cariappa: You should replicate the matrix and not modify it.
We want to test both with out-of-process and in-process JIT.
You can add new names like `ubu22-arm-gcc12-clang-repl-20-oop`.
Also, we do not need to test where `converage: true`.

CMakeLists.txt
```diff
@@ -305,6 +305,13 @@ else()
   message(FATAL_ERROR "We need either CPPINTEROP_USE_CLING or CPPINTEROP_USE_REPL")
 endif()
 
+if(CPPINTEROP_WITH_OOP_JIT)
+  add_definitions(-DCPPINTEROP_WITH_OOP_JIT)
+endif()
+
+string(REGEX REPLACE "/build/lib/cmake/llvm$" "" LLVM_SOURCE_DIR "${LLVM_DIR}")
+add_definitions(-DLLVM_SOURCE_DIR="${LLVM_SOURCE_DIR}")
```
Vipul-Cariappa, line 321:
```suggestion
string(REGEX REPLACE "/build/lib/cmake/llvm$" "" LLVM_BINARY_DIR "${LLVM_DIR}")
add_definitions(-DLLVM_SOURCE_DIR="${LLVM_BINARY_DIR}")
```

I guess is a better name.

lib/CppInterOp/Compatibility.h
```diff
@@ -246,16 +258,53 @@ createClangInterpreter(std::vector<const char*>& args) {
...
+      }
+  }
+
+  auto innerOrErr =
+      CudaEnabled ? clang::Interpreter::createWithCUDA(std::move(*ciOrErr),
+                                                       std::move(DeviceCI))
+                  : clang::Interpreter::create(std::move(*ciOrErr), std::move(JB));
+  #else
+  if(outOfProcess) {
+    llvm::errs() << "[CreateClangInterpreter]: No compatibility with out-of-process JIT"
```
Vipul-Cariappa:
```suggestion
  if(outOfProcess) {
    llvm::errs() << "[CreateClangInterpreter]: No compatibility with out-of-process JIT (To enable recompile CppInterOp with `-DCPPINTEROP_WITH_OOP_JIT=ON`)"
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3031,11 +3031,11 @@ static std::string MakeResourcesPath() {
 } // namespace
 
 TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
-                            const std::vector<const char*>& GpuArgs /*={}*/) {
+                            const std::vector<const char*>& GpuArgs /*={}*/, bool is_out_of_process) {
   std::string MainExecutableName = sys::fs::getMainExecutable(nullptr, nullptr);
   std::string ResourceDir = MakeResourcesPath();
   std::vector<const char*> ClingArgv = {"-resource-dir", ResourceDir.c_str(),
-                                        "-std=c++14"};
+                                        "-std=c++14", "-gdwarf-4", "-O0"};
```
```suggestion
                                        "-std=c++14"};
```
You should not hard code it here.

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -146,6 +146,23 @@ TEST(InterpreterTest, Process) {
...
+  std::vector<const char*> interpreter_args = { "-include", "new" };
+  auto* I = Cpp::CreateInterpreter(interpreter_args, {}, true);
+  EXPECT_TRUE(Cpp::Process("") == 0);
+  EXPECT_TRUE(Cpp::Process("int a = 12;") == 0);
+  EXPECT_FALSE(Cpp::Process("error_here;") == 0);
+  // Linker/JIT error.
+  EXPECT_FALSE(Cpp::Process("int f(); int res = f();") == 0);
+}
+#endif
+
```
Vipul-Cariappa, line 165: According to this, we only run one test with out-of-process, but we should run all the tests with out-of-process. This should be done through the CI matrix.

lib/CppInterOp/Compatibility.h
```diff
@@ -246,16 +258,53 @@ createClangInterpreter(std::vector<const char*>& args) {
   (*ciOrErr)->LoadRequestedPlugins();
   if (CudaEnabled)
     DeviceCI->LoadRequestedPlugins();
+
+  #ifdef CPPINTEROP_WITH_OOP_JIT
+  std::unique_ptr<llvm::orc::LLJITBuilder> JB;
+
+  if(outOfProcess) {
+      std::string OOPExecutor = std::string(LLVM_SOURCE_DIR) + "/build/bin/llvm-jitlink-executor";
```
Vipul-Cariappa: Will this work if LLVM is installed with conda or is installed with the system's package installer.
Could you make sure if working in the other cases too.
You can first check if the file exists; if not, try other paths like the `CONDA_PREFIX` or `/usr/bin`.

Vipul-Cariappa:
Hi @mcbarton, do we have cache to add another version of LLVM? For the `out-of-process.patch`. We need the normal `LLVM-20` and `LLVM-20-OOP` where `OOP` is for `out-of-process`?
Could you please help @kr-2003 achieve this?

CMakeLists.txt
kr-2003: LLVM_BINARY_DIR is already an env variable

Vipul-Cariappa: Can we reuse `LLVM_BINARY_DIR`?

.github/actions/Build_LLVM/action.yml
```diff
@@ -64,6 +68,13 @@ runs:
                 -DLLVM_INCLUDE_TESTS=OFF                        \
                 ../llvm
           ninja clang clangInterpreter clangStaticAnalyzerCore -j ${{ env.ncpus }}
+          if [[ "${{ matrix.oop-jit }}" == "On" ]]; then
+            if [[ "${{ runner.os }}" == "Linux" ]]; then
+              ninja clang clang-repl llvm-jitlink-executor orc_rt -j ${{ env.ncpus }}
+            elif [[ "${{ runner.os }}" == "macOS" ]]; then
+              ninja clang clang-repl llvm-jitlink-executor orc_rt_osx -j ${{ env.ncpus }}
+            fi
+          fi
```
mcbarton, line 74: Why do we have target name for orc_rt difference for MacOS? The target name should be independent of the system your building on.

.github/actions/Build_LLVM/action.yml
```diff
@@ -49,6 +49,10 @@ runs:
             git apply -v ../patches/llvm/clang${{ matrix.clang-runtime }}-*.patch
             echo "Apply clang${{ matrix.clang-runtime }}-*.patch patches:"
           fi
+          if [[ "${{ matrix.oop-jit }}" == "On" ]]; then
+            git apply -v ../patches/llvm/clang20-1-out-of-process-jit-execution.patch
+            echo "Apply out-of-process-jit-execution.patch:"
+          fi
```
mcbarton, line 55: We should apply this patch regardless of whether matrix.oop-jit=on.  If you only apply it if CPPINTEROP_WITH_OOP_JIT=ON , this would imply if you apply the patch, DCPPINTEROP_WITH_OOP_JIT=OFF, then somehow something will go wrong.

Also, why are we not applying this patch on Windows, and testing CPPINTEROP_WITH_OOP_JIT=ON/OFF?

.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -40,15 +40,29 @@ runs:
...
+          else
+              cmake -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}     \
+                -DCPPINTEROP_INCLUDE_DOCS=${{ matrix.documentation }} \
+                -DLLVM_DIR=$LLVM_BUILD_DIR/lib/cmake/llvm    \
+                -DClang_DIR=$LLVM_BUILD_DIR/lib/cmake/clang  \
+                -DBUILD_SHARED_LIBS=ON                       \
+                -DCODE_COVERAGE=${{ env.CODE_COVERAGE }}     \
+                -DCMAKE_INSTALL_PREFIX=$CPPINTEROP_DIR       \
+                -DLLVM_ENABLE_WERROR=On                      \
+                -DCPPINTEROP_WITH_OOP_JIT=OFF                \
```
mcbarton: Feels like CPPINTEROP_WITH_OOP_JIT should have a default value, like we have for cppinterop_use_repl/cppinterop_use_cling

.github/workflows/main.yml
```diff
@@ -32,6 +32,15 @@ jobs:
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
             coverage: true
+          - name: ubu24-arm-gcc12-clang-repl-20-oop
```
mcbarton, line 35: We need a Ubuntu x86 job and Windows job surely? 

Is there a reason why the Emscripten jobs don't try and use this feature you've added?

patches/llvm/clang20-1-out-of-process-jit-execution.patch
```diff
@@ -0,0 +1,710 @@
+diff --git a/clang/include/clang/Interpreter/Interpreter.h b/clang/include/clang/Interpreter/Interpreter.h
```
mcbarton: This is quite a large patch. Can you confirm that it has been upsteamed so will appear in llvm 21?

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -146,6 +146,23 @@ TEST(InterpreterTest, Process) {
   clang_Interpreter_dispose(CXI);
 }
 
+// #ifdef CPPINTEROP_WITH_OOP_JIT
```
mcbarton: Why is this test currently commented out?

README.md
```diff
@@ -120,6 +120,13 @@ git apply -v clang{version}-*.patch
 
 on Windows.
 
+If you want to have out-of-process JIT execution enabled in CppInterOp, then apply this patch on Linux and MacOS environment. 
+> Note that this patch will not work for Windows because out-of-process JIT execution is currently implemented for Linux and MacOS only.
+
+```bash
+git apply -v ../CppInterOp/patches/llvm/out-of-process-jit-execution.patch
```
mcbarton: The patch name doesn't match what you have in the PR

README.md
```diff
@@ -120,6 +120,13 @@ git apply -v clang{version}-*.patch
 
 on Windows.
 
+If you want to have out-of-process JIT execution enabled in CppInterOp, then apply this patch on Linux and MacOS environment. 
+> Note that this patch will not work for Windows because out-of-process JIT execution is currently implemented for Linux and MacOS only.
```
mcbarton, line 124: Although we have numerous tests disabled for Windows, this is due to an issue writing to standard output. Do we really want add a feature, we cannot currently support for Windows.

README.md
```diff
@@ -120,6 +120,13 @@ git apply -v clang{version}-*.patch
 
 on Windows.
```
mcbarton, line 122: You have updated the readme, but not the documentation in the docs folder that appears online.

mcbarton (changes requested):

.github/actions/Build_LLVM/action.yml
Vipul-Cariappa: It is different in LLVM itself. I guess it might have to do with cross-compiling.

.github/actions/Build_LLVM/action.yml
Vipul-Cariappa: Ok. The thing is, I would want CppInterOp to be compatible with upstream LLVM as well as the patched LLVM. Therefore, I was thinking about having 2 different CI workflows for version 20 with OOP-JIT and without OOP-JIT.
OOP-JIT is not supported on Windows.

---

Ideally, this is temporary till the patch is directly merged into LLVM and a new version is released.

.github/workflows/main.yml
Vipul-Cariappa: We only need Ubuntu and OSX. This is not supported in WASM nor Windows.

README.md
Vipul-Cariappa: We have thumbs up from Vassil.

.github/actions/Build_LLVM/action.yml
mcbarton: I could be wrong, and @vgvassilev can correct me if I am wrong, but for such large patches, we have had them merged into llvm first, before applying them to previous versions in CppInterOp.

You could have a new llvm cache like you want. One where you build OOP-JIT and without OOP-JIT. This could only really be done after I find a way to make space in the cache, since we currently are unlikely to have enough space for all these new llvm cache builds. I have an idea on how to free up some space, and just need to find some time to do. I will try and do it tomorrow, as I am busy today.

.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -40,15 +40,29 @@ runs:
...
+                ../
+          fi
```
mcbarton:
```suggestion
              cmake -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}     \
                -DCPPINTEROP_INCLUDE_DOCS=${{ matrix.documentation }} \
                -DLLVM_DIR=$LLVM_BUILD_DIR/lib/cmake/llvm    \
                -DClang_DIR=$LLVM_BUILD_DIR/lib/cmake/clang  \
                -DBUILD_SHARED_LIBS=ON                       \
                -DCODE_COVERAGE=${{ env.CODE_COVERAGE }}     \
                -DCMAKE_INSTALL_PREFIX=$CPPINTEROP_DIR       \
                -DLLVM_ENABLE_WERROR=On                      \
                -DCPPINTEROP_WITH_OOP_JIT=${{ matrix.oop-jit }}                 \
                ../
```

.github/actions/Build_LLVM/action.yml
mcbarton:
```suggestion
          if [[ "${{ matrix.oop-jit }}" == "On" ]]; then
            if [ "${{ matrix.os }}}" = "macos"*  ]; then SUFFIX="_osx"; fi
            ninja llvm-jitlink-executor orc_rt${SUFFIX}
          fi
```

CMakeLists.txt
```diff
@@ -310,6 +310,13 @@ include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})
 separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
 add_definitions(${LLVM_DEFINITIONS_LIST})
 
+if(CPPINTEROP_WITH_OOP_JIT)
```
mcbarton, line 313: If set to ON when on Windows or Emscripten then this should error with a message about those platforms not supporting that option.

unittests/CppInterOp/InterpreterTest.cpp
kr-2003: previously there was some bug in out of process Jit execution. but now that has been resolved, this test can be uncommented now.

unittests/CppInterOp/InterpreterTest.cpp
kr-2003: yes, i added only this for now to test it out. In the future, all the tests should be ran against out-of-process JIT.

patches/llvm/clang20-2-out-of-process-jit-execution.patch
```diff
@@ -0,0 +1,780 @@
...
++#include "llvm/ExecutionEngine/Orc/SimpleRemoteEPC.h"
++#include "llvm/Support/Error.h"
++
++#include <cstdint>
++#include <memory>
++#include <string>
++
++llvm::Expected<std::unique_ptr<llvm::orc::SimpleRemoteEPC>>
++launchExecutor(llvm::StringRef ExecutablePath, bool UseSharedMemory,
++               llvm::StringRef SlabAllocateSizeString);
```
Vipul-Cariappa, line 63:
```suggestion
+launchExecutor(llvm::StringRef ExecutablePath, bool UseSharedMemory,
+               llvm::StringRef SlabAllocateSizeString, int stdin_fd = 0, int stdout_fd = 1, int stderr_fd = 2);
```

patches/llvm/clang20-2-out-of-process-jit-execution.patch
```diff
@@ -0,0 +1,780 @@
...
++    close(ToExecutor[WriteEnd]);
++    close(FromExecutor[ReadEnd]);
++
++    // Close read ends of stdout and stderr pipes
++    close(StdoutPipe[ReadEnd]);
++    close(StderrPipe[ReadEnd]);
++
++    // Redirect stdout and stderr to the pipes
++    dup2(StdoutPipe[WriteEnd], STDOUT_FILENO);
++    dup2(StderrPipe[WriteEnd], STDERR_FILENO);
```
Vipul-Cariappa, line 322:
```suggestion
+ if (stdout_fd != 1) ...
+    dup2(stdout_fd, STDOUT_FILENO);
+    dup2(StderrPipe[WriteEnd], STDERR_FILENO);
```

PR #638, opened 2025-06-20
[docs] update header include to new style

aaronj0:


mcbarton:
@aaronj0 Sorry about stopping the workflows running in this PR. I did it as I need to rebuild the cache on main at the moment, since some of the builds got deleted. Without rebuilding the cache, the space we have to stored llvm builds would be overwhelmed quite quickly. If you want I will restart the workflow in this PR in about 1.5 hours, once the cache is rebuild. You can also restart them yourself once this workflow finishes running https://github.com/compiler-research/CppInterOp/actions/runs/15765979794 which is rebuilding the cache.

aaronj0:
> @aaronj0 Sorry about stopping the workflows running in this PR. I did it as I need to rebuild the cache on main at the moment, since some of the builds got deleted. Without rebuilding the cache, the space we have to stored llvm builds would be overwhelmed quite quickly. If you want I will restart the workflow in this PR in about 1.5 hours, once the cache is rebuild. You can also restart them yourself once this workflow finishes running https://github.com/compiler-research/CppInterOp/actions/runs/15765979794 which is rebuilding the cache.

no worries, thanks!

PR #640, opened 2025-06-20
fix: codegen for templates with multiple parameter packs

Vipul-Cariappa:
# Description

adapted from cling

## Fixes # (issue)

Fixes 2 issues in cppyy.

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

Included.

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -616,8 +616,6 @@ static Decl* GetScopeFromType(QualType QT) {
     Type = Type->getUnqualifiedDesugaredType();
     if (auto* ET = llvm::dyn_cast<EnumType>(Type))
       return ET->getDecl();
-    if (auto* FnType = llvm::dyn_cast<FunctionProtoType>(Type))
-      Type = const_cast<clang::Type*>(FnType->getReturnType().getTypePtr());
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3422,6 +3448,8 @@ static Decl* InstantiateTemplate(TemplateDecl* TemplateD,
   // This will instantiate tape<T> type and return it.
   SourceLocation noLoc;
   QualType TT = S.CheckTemplateIdType(TemplateName(TemplateD), noLoc, TLI);
+  if (TT.isNull())
+    return nullptr;
```
vgvassilev, line 3452: I'd drop that if we can't cover it with tests.

Vipul-Cariappa: This is required; it is due to a template instantiation failing in Eigen (crashes).
Also, fixing the instantiation in this PR does not make sense. Will look into that once I get to the eigen tests.
And writing a test that is supposed to fail is bad again.

vgvassilev: Ok. Understood

vgvassilev: Let’s merge it then.

PR #642, opened 2025-06-23
add `GetTemplateInstantiatedFunctions`

Vipul-Cariappa:
# Description

lookup instantated templated functions

## Fixes # (issue)

1 test in cppyy

## Type of change

- [ ] Bug fix
- [x] New feature

## Testing

Included

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -863,6 +863,21 @@ void GetFunctionTemplatedDecls(TCppScope_t klass,
   GetClassDecls<FunctionTemplateDecl>(klass, methods);
 }
 
+void GetTemplateInstantiatedFunctions(const std::string& name,
+                                      TCppScope_t scope,
+                                      std::vector<TCppFunction_t>& methods) {
+  std::vector<TCppFunction_t> tmpl_funcs;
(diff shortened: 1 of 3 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -863,6 +863,21 @@ void GetFunctionTemplatedDecls(TCppScope_t klass,
   GetClassDecls<FunctionTemplateDecl>(klass, methods);
 }
 
+void GetTemplateInstantiatedFunctions(const std::string& name,
```
vgvassilev, line 866: Why would we need all possible instantiations? That encourages iteration other expensive operations.

Vipul-Cariappa: Required [here](https://github.com/compiler-research/cppyy/blob/aa3efe698f60f8d50026d0ef1f44e131912f564e/test/test_advancedcpp.py#L618).

In cppyy, if there is an instantiation of a templated function, then it is treated as an overloaded function. This also prevents instantiation because we have already extracted the instantiated functions.

vgvassilev: We should avoid iteration. Perhaps we can use some level of FindInstantiation or similar, which does not push everything onto a vector…

PR #650, opened 2025-06-27
Improve Cpp::Construct pointer logic for memory arena flags

aaronj0:
Also handle both ctors and class scopes. Upstreamed from https://github.com/root-project/root/pull/18546


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3783,22 +3783,26 @@ void Deallocate(TCppScope_t scope, TCppObject_t address, TCppIndex_t count) {
...
-      return arena;
-    }
+  TCppFunction_t ctor;
+  if (Cpp::IsConstructor(scope))
+    ctor = scope;
+  else if (Cpp::IsClass(scope)) {
+    if (!HasDefaultConstructor(scope))
+      return nullptr;
+    ctor = Cpp::GetDefaultConstructor(scope);
+  } else
```
vgvassilev: We can invert the rule here to return a nullptr when not a class, not a ctor, and a class without default ctor.

```suggestion
  if (!Cpp::IsConstructor(scope) || !Cpp::IsClass(scope))
    return nullptr;
  if (Cpp::IsClass(scope) && !HasDefaultConstructor(scope))
    return nullptr;

  TCppFunction_t ctor = nullptr;
  if (Cpp::Class(scope))
    ctor = Cpp::GetDefaultConstructor(scope);
  else // a ctor
    ctor = scope
```

mcbarton:
@aaronj0 Ignore the failing Emscripten job. It wasn't anything you did. It happened on one of my PRs. The Emscripten static library for llvm 19 is a little demanding on Ubuntu Github runners. If you rerun the job it should pass.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3783,22 +3783,27 @@ void Deallocate(TCppScope_t scope, TCppObject_t address, TCppIndex_t count) {
...
-    JC.InvokeConstructor(&obj, count, {}, nullptr);
-    return obj;
+  if (JitCall JC = MakeFunctionCallable(&interp, ctor)) {
+    if (JC.getKind() != Cpp::JitCall::kConstructorCall)
+      return nullptr;
+    // invoke the constructor (placement/heap) in one shot
+    // flag is non-null for placement new, null for normal new
+    void* flag = arena ? reinterpret_cast<void*>(1) : nullptr;
+    void* result = arena;
+    JC.InvokeConstructor(&result, count, {}, flag);
```
vgvassilev: We need to add the arg names as comments to know what’s what.

```suggestion
    void* is_arena = arena ? reinterpret_cast<void*>(1) : nullptr;
    void* result = arena;
    JC.InvokeConstructor(&result, count, /*args=*/{}, is_arena);
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3783,22 +3783,27 @@ void Deallocate(TCppScope_t scope, TCppObject_t address, TCppIndex_t count) {
...
+    ctor = Cpp::GetDefaultConstructor(scope);
+  else // a ctor
+    ctor = scope;
 
-    void* obj = nullptr;
-    JC.InvokeConstructor(&obj, count, {}, nullptr);
-    return obj;
+  if (JitCall JC = MakeFunctionCallable(&interp, ctor)) {
+    if (JC.getKind() != Cpp::JitCall::kConstructorCall)
+      return nullptr;
```
vgvassilev: Shouldn't this be an assert?

aaronj0: I will drop this check since it is taken care of in `InvokeConstructor`:

https://github.com/compiler-research/CppInterOp/blob/347490300317faa6ea10e05e37c124deddfdbf29/include/CppInterOp/CppInterOp.h#L228-L230

vgvassilev (approved):
Nice! LGTM!

aaronj0:
@mcbarton There is something wrong with the Emscripten build jobs, they have been running for over an hour and (seems to have) stopped progressing past `[ 55%] Built target googletest`

aaronj0:
I am going ahead and merging this PR since the test that was extended is not run on windows, nor emscripten builds with llvm < 20 (which cover the 3 jobs that are stalled), but the issue on the CI must be investigated

PR #651, opened 2025-06-27
add `IsClassDefined`

Vipul-Cariappa:


include/CppInterOp/CppInterOp.h
```diff
@@ -265,6 +265,9 @@ CPPINTEROP_API bool IsNamespace(TCppScope_t scope);
 /// Checks if the scope is a class or not.
 CPPINTEROP_API bool IsClass(TCppScope_t scope);
 
+/// Checks if the klass has a definition
+CPPINTEROP_API bool IsClassDefined(TCppScope_t klass);
```
vgvassilev, line 269: How is that different from `IsComplete`?

Vipul-Cariappa: `IsComplete` does what I want. Sorry, I missed that API. Will close this PR.

PR #655, opened 2025-07-01
Reduce CppInterOp Emscripten shared library size

mcbarton:
# Description

@vgvassilev This change locally reduced the Emscripten shared library size to 47/48 Mb (depends of whether you use dir -lh or du -lh). It still passed all tests and xeus-cpp still could run all the notebook cells it could before (exception throwing is broken in CppInterOps deployment at the moment for some reason). I have cleared the cache of all llvm 20 Emscripten builds for this PR.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -320,10 +320,12 @@ jobs:
           {
             git apply -v emscripten-clang19-2-shift-temporary-files-to-tmp-dir.patch
             git apply -v emscripten-clang19-3-remove-zdefs.patch
+            git apply -v emscripten-clang19-4-enable_exception_handling.patch
           }
           elseif ( "${{ matrix.clang-runtime }}" -imatch "20" )
           {
             git apply -v emscripten-clang20-2-shift-temporary-files-to-tmp-dir.patch
+            git apply -v emscripten-clang20-3-enable_exception_handling.patch
           }
           cd build
           echo "Apply clang${{ matrix.clang-runtime }}-*.patch patches:"
@@ -587,6 +589,14 @@ jobs:
           emrun --browser="Google Chrome" --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  CppInterOpTests.html
           echo "Running DynamicLibraryManagerTests in Google Chrome"          
           emrun --browser="Google Chrome" --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  DynamicLibraryManagerTests.html
+          sudo safaridriver --enable
+          python -m pip install selenium
+          echo "Running CppInterOpTests in Safari"
+          emrun --no_browser --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  CppInterOpTests.html &
+          python ../../../scripts/browser_tests_safari.py CppInterOpTests.html
+          echo "Running DynamicLibraryManagerTests in Safari"          
+          emrun --no_browser --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  DynamicLibraryManagerTests.html &
+          python ../../../scripts/browser_tests_safari.py DynamicLibraryManagerTests.html
         else
           export ARCHITECHURE=$(uname -m)
           if [[ "$ARCHITECHURE" != "aarch64" ]]; then
@@ -727,6 +737,12 @@ jobs:
           emrun --browser="Google Chrome" --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  CppInterOpTests.html
           echo "Running DynamicLibraryManagerTests in Google Chrome"          
           emrun --browser="Google Chrome" --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  DynamicLibraryManagerTests.html
+          echo "Running CppInterOpTests in Safari"
+          emrun --no_browser --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  CppInterOpTests.html &
+          python ../../../scripts/browser_tests_safari.py CppInterOpTests.html
+          echo "Running DynamicLibraryManagerTests in Safari"          
+          emrun --no_browser --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  DynamicLibraryManagerTests.html &
+          python ../../../scripts/browser_tests_safari.py DynamicLibraryManagerTests.html
         else
           export ARCHITECHURE=$(uname -m)
           if [[ "$ARCHITECHURE" != "aarch64" ]]; then
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -168,6 +168,14 @@ jobs:
         emrun --browser="Google Chrome" --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  CppInterOpTests.html
         echo "Running DynamicLibraryManagerTests in Google Chrome"          
         emrun --browser="Google Chrome" --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  DynamicLibraryManagerTests.html
+        sudo safaridriver --enable
+        python -m pip install selenium
+        echo "Running CppInterOpTests in Safari"
+        emrun --no_browser --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  CppInterOpTests.html &
+        python ../../../scripts/browser_tests_safari.py CppInterOpTests.html
+        echo "Running DynamicLibraryManagerTests in Safari"          
+        emrun --no_browser --kill_exit --timeout 60 --browser-args="--headless --no-sandbox"  DynamicLibraryManagerTests.html &
+        python ../../../scripts/browser_tests_safari.py DynamicLibraryManagerTests.html
 
         cd ../../
         emmake make -j ${{ env.ncpus }} install
--- a/Emscripten-build-instructions.md
+++ b/Emscripten-build-instructions.md
@@ -64,6 +64,7 @@ On Windows execute the following
 cd .\llvm-project\
 cp -r ..\patches\llvm\emscripten-clang20*
 git apply -v emscripten-clang20-2-shift-temporary-files-to-tmp-dir.patch
+git apply -v emscripten-clang20-3-enable_exception_handling.patch
 ```
 
 We are now in a position to build an emscripten build of llvm by executing the following on Linux
@@ -225,7 +226,7 @@ It is possible to run the Emscripten tests in a headless browser. To do this we 
     cd ./unittests/CppInterOp/
 ```
 
-We will run our tests in a fresh installed browser. Installing the browsers, and running the tests within the installed browsers will be platform dependent. To do this on MacOS execute the following
+We will run our tests in a fresh installed browser. Installing the browsers, and running the tests within the installed browsers will be platform dependent. To do this for Chrome and Firefox on MacOS execute the following
 
 ```bash
 wget "https://download.mozilla.org/?product=firefox-latest&os=osx&lang=en-US" -O Firefox-latest.dmg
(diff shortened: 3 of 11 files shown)


anutosh491:
It's a compile time optimization flag I think. 

Do we need to force it like this everywhere or can we have cmake handle this ?

anutosh491:
Yeah but we still don't want users to look into/pass flags correct during the build correct?

 We should be able to have cmake handle this.

mcbarton:
> Yeah but we still don't want users to look into/pass flags correct during the build correct?
> 
> We should be able to have cmake handle this.

I really don't understand the point your trying to make. Using EMCC_CFLAGS in this way to perfectly legitimate way of doing the Emscripten build. Emscripten is designed to be effected by modifying this flag (see https://emscripten.org/docs/tools_reference/emcc.html#environment-variables). If a user changes what we have in the documentation for this flag, and messes something up, then that is on them.

anutosh491:
My point is simple. We already have compile time and link time flags being passed through cmake, I don't see a need to pass this externally.

vgvassilev:
I think what @anutosh491 means is that we need to make this flag as part of our default compiler flags in the CMakeLists.txt.

mcbarton:
@vgvassilev @anutosh491 I first tried locally using `-DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4 -Oz" -DCMAKE_C_FLAGS="-Oz" -DCMAKE_STATIC_LINKER_FLAGS="-Oz"`  (since I thought this might be the closest thing to EMCC_CFLAGS) and it crashed 15% into the build. I then tried `-DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4 -Oz" -DCMAKE_C_FLAGS="-Oz"` which although I could build llvm with this option, had no effect on the size of CppInterOps shared library, or the llvm libraries. I do not know how to proceed now.

mcbarton:
Putting a note here in case I forget. Using EMCC_CFLAGS="-flto" in the llvm build doesn't end up effecting the shared library size on its own. Used in conjuncation with Oz then the size falls slightly to 45/46 Mb.

mcbarton:
I have updated this PR to use the flto flag as well, and to compile+link CppInterOp shared library with these flags too. This reduces the shared library size to 44/45 Mb.

.github/workflows/emscripten.yml
```diff
@@ -194,6 +194,9 @@ jobs:
                         -DLLVM_ENABLE_LIBPFM=OFF                        \
                         -DCLANG_BUILD_TOOLS=OFF                         \
                         -DLLVM_NATIVE_TOOL_DIR=$NATIVE_DIR 		\
+                        -DCMAKE_C_FLAGS_RELEASE="-Oz -g0" \
+                        -DCMAKE_CXX_FLAGS_RELEASE="-Oz -g0" \
+                        -DLLVM_ENABLE_LTO=Full \
```
anutosh491, line 198: I am not sure, we necessarily need to change the way llvm is built for this.

lib/CppInterOp/CMakeLists.txt
```diff
@@ -127,10 +127,16 @@ if(EMSCRIPTEN)
...
+  target_compile_options(clangCppInterOp
+    PRIVATE "SHELL: -Oz"
+    PRIVATE "SHELL: -flto"
+  )
   target_link_options(clangCppInterOp
     PRIVATE "SHELL: -s WASM_BIGINT"
     PRIVATE "SHELL: -s SIDE_MODULE=1"
     PRIVATE "SHELL: ${SYMBOLS_LIST}"
+    PRIVATE "SHELL: -Oz"
+    PRIVATE "SHELL: -flto"
```
anutosh491, line 139: Hmm why is lto being involved here ? I've tried using it in the past and I remember it behaves notoriously. 
How is it helping us here ?

.github/workflows/emscripten.yml
mcbarton: We absolutely need to change the llvm build to get this reduction in size of Emscripten shared library. The majority of the size reduction this PR provides comes from adding these flags to the llvm build.

lib/CppInterOp/CMakeLists.txt
mcbarton: Adding these flags gets us a further reduction in the size of the Emscripten shared library of a few Mb. Needed them for both compile and link flags for it to have any effect. I didn't see any bad side effects your comment suggests (I don't know exactly what you mean by behaves notoriously).

vgvassilev:
Can we add some check in the CI to fire when we break this. Maybe something that checks if the file is greater than 50 and fail the build?

mcbarton:
> Can we add some check in the CI to fire when we break this. Maybe something that checks if the file is greater than 50 and fail the build?

I have added a check, which on my local system will error if the size is greater than 46 Mb. Need to wait for ci to pass, to check I implemented correctly in the workflow file.

vgvassilev:
> Hi, please refrain from merging this PR before a set of reviews on my end.
> 
> Emscripten-forge applies few optimization flags itself (similar to what pyodide does) to add a basic level of optimization for every package while building.
> 
> https://github.com/emscripten-forge/recipes/blob/e04b1067f9d7e4e56d742e5d3efa2131ed8a0315/recipes/recipes/emscripten_emscripten-wasm32/activate.sh#L50-L68
> 
> So if this goes in (and after a release if we build the latest cppinterop on emscripten-forge) It's possible these flags go against each other/ overwrite each other.

I do not see how these flags can go against each other. We have tests that are running and if the tests are good I do not see a reason to hold things off. Right now we destroy people's network by forcing them to download 300mb, ~100 of which is CppInterOp. If this PR cuts the CppInterOp size by half while the tests still run this is something we need urgently, right?

anutosh491:
Hey @mcbarton can you educate me with some questions

1) What's the current size of libclangCppInterOp.so being built on main ? Guessing its somewhere around 70-80 MB

2) Now as we see that CppInterop's deployment runs on iphone .... which means that the above size would suffice somehow, I am trying to understand what exactly is being optimized here ?

Cause according to my attempts 
i) I see that using o2 gets it to 80 MB
ii) Using oz gets it to 73 MB
iii) flto doesn't have any strong affect
iv) even getting rid of the huge list of symbols we export isn't really having any strong affect on size.

So only making changes to cppinterop might not be good enough ... which gets us to llvm. 

Now I am trying to understand what exactly is being optimized through the above flags to get job done cause libclangInterpreter.a which we link against is anyways not too huge (412K on emscripten-forge) though the transitive dependencies being pulled in might be heavy (I see libclang-cpp.a is 147M)

mcbarton:
@vgvassilev @anutosh491 If I build locally according to the instructions in the readme of this PR, this it the deployment I get https://mcbarton.github.io/xeus-cpp-demo/lab/index.html.  Everything which works for me when follow mains Emscripten instructions works in this deployment too. 

The 'SIMD Acceleration through WebAssembly' example doesn't work in the link, but it doesn't work locally if you build according to the instructions on main (e.g. no modified llvm or CppInterOp). That is the reason I have opened this issue https://github.com/compiler-research/CppInterOp/issues/680

mcbarton:
> @vgvassilev @anutosh491 If I build locally according to the instructions in the readme of this PR, this is the deployment I get https://mcbarton.github.io/xeus-cpp-demo/lab/index.html. Everything which works for me when I follow mains Emscripten instructions works in this deployment too.
> 
> The 'SIMD Acceleration through WebAssembly' example doesn't work in the link (fails to find wasm_simd128.h), but it doesn't work locally if you build according to the instructions on main (e.g. no modified llvm or CppInterOp). That is the reason I have opened this issue #680

With the changes in this PR, the shared library size is down to 39 Mb. The ci has been changed to check whether any changes make the library greater than 40 Mb.

.github/workflows/emscripten.yml
```diff
@@ -175,7 +175,6 @@ jobs:
                         -DLLVM_EXTERNAL_CLING_SOURCE_DIR=../../cling  \
                         -DCMAKE_BUILD_TYPE=Release \
                         -DLLVM_HOST_TRIPLE=wasm32-unknown-emscripten \
-                        -DLLVM_ENABLE_ASSERTIONS=ON                        \
```
mcbarton, line 178: I removed LLVM_ENABLE_ASSERTIONS from the build instructions since off is the default for a release build.

vgvassilev: Can we do that in a separate PR?

lib/CppInterOp/CMakeLists.txt
```diff
@@ -127,10 +127,16 @@ if(EMSCRIPTEN)
   set_target_properties(clangCppInterOp
     PROPERTIES NO_SONAME 1
   )
+  target_compile_options(clangCppInterOp
+    PRIVATE "SHELL: -Oz"
+    PRIVATE "SHELL: -flto"
```
vgvassilev, line 132: Can't these become part of `CMAKE_X_FLAGS_RELEASE` flags, too?

mcbarton: I will give it a go and find out.

.github/workflows/emscripten.yml
mcbarton: @vgvassilev unless you have a specific reason for wanting this in a separate PR, I would prefer it in this PR. This PR is about reducing the shared library size, and assertions are one thing which is causing the shared library to be large. They are responsible for 5 Mb of the size reduction.

vgvassilev: Ok!

lib/CppInterOp/CMakeLists.txt
mcbarton: I tried this out and couldn't get it to work. For some reason if I do it via cmake options then Ubuntu both x86 and arm cannot pass xeus-cpps Emscripten tests for llvm 20. I have reverted back to the original solution now, and waiting to see the ci pass

vgvassilev: How's that even possible...

mcbarton: Its not supposed to he possible according to the WebAssembly/Emscripten standard. The result should be independent of the platform you cross compile from, but I have come accross a few examples where this is found not to be true in reality (and shows exactly why we should be cross compiling from multiple platforms in the ci to tests these assumptions). In the process of trying to get the cmake approach to work I found out that the symbols you need to export on each platform manually to get the CppInterOps tests to pass are not the same (e.g. you can not export _ZN4llvm15SmallVectorBaseIjE8set_sizeEm , and the Emscripten tests will build and pass on Ubuntu, but wont build on MacOS or Windows)

mcbarton: @vgvassilev Here is a workflow run where Ubuntu arm and x86 fail for llvm 20 Emscripten builds using cmake options https://github.com/compiler-research/CppInterOp/actions/runs/16596946952 despite the other platforms passing

mcbarton:
The old way of doing it is still able to pass the ci for all jobs.

anutosh491:
Hi, 

Some remarks based on my comment here (https://github.com/compiler-research/CppInterOp/pull/655#issuecomment-3077043264)

One of my questions there was 

> I am trying to understand what exactly is being optimized here ?

To be fair, I still don't understand what exactly brings the size :\ 

Just thinking about what can bring down the size from 80 to 45

1) Just making changes to cppinterop bring a change of about +- 5MB (which says we need to make changes to llvm too)
2) Changing from 02 to 0z for any project can also probably get a change of +- 5 to 10 MB (these flags are not very vastly different)
3) I can see that LLVM_ENABLE_ASSERTIONS would also bring the size down.
4) which means LTO on llvm does help but I am not sure I can pin point how it does so

I would have been really happy if we could pin point on what exactly changes in llvm when we us lto (in combination with oz) and how does it get down cppinterop's size as a whole but I guess this is OK.

Let's try this approach out and try to backtrack how it helps :)

PR #657, opened 2025-07-04
fix JITcall codegen to handle move semantics

Vipul-Cariappa:
# Description

Required for std::unique_ptr

Taken from Cling

## Fixes # (issue)

1 test in cppyy

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -2160,12 +2160,32 @@ void make_narg_call(const FunctionDecl* FD, const std::string& return_type,
       }
     }
 
+    CXXRecordDecl* rtdecl = QT->getAsCXXRecordDecl();
     if (refType != kNotReference) {
       callbuf << "(" << type_name.c_str()
               << (refType == kLValueReference ? "&" : "&&") << ")*("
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2160,12 +2160,32 @@ void make_narg_call(const FunctionDecl* FD, const std::string& return_type,
...
+      // By-value construction; this may either copy or move, but there is no
+      // information here in terms of intent. Thus, simply assume that the
+      // intent is to move if there is no viable copy constructor (ie. if the
+      // code would otherwise fail to even compile). There does not appear to be
+      // a simple way of determining whether a viable copy constructor exists,
+      // so check for the most common case: the trivial one, but not uniquely
+      // available, while there is a move constructor.
+
+      // move construction as needed for classes (note that this is implicit)
+      static bool included_utility = false;
```
vgvassilev: I don't think this would work well when we have a stack of interpreters, that is changing the active interpreter. Perhaps a simple lookup will help, clang::Sema already `getStdNamespace`.

PR #658, opened 2025-07-08
[GSoC 2025] Enabling Out-Of-Process JIT Execution in CppInterOp

kr-2003:
# Description

> This PR supports the GSoC project - "Implementing Debugger Support for xeus-cpp".

This PR introduces creation of interpreter that runs JIT code in different process(out-of-process). This is the [PR](https://github.com/llvm/llvm-project/pull/144064/) on LLVM that supports this feature on LLVM side. Since, this feature on LLVM is expected to be released with 21.x, till then CppInterOp will be using a patch for LLVM build.

## Changes
1. OOP LLVM Patch for the LLVM build.
2. ``CreateInterpreter`` when passed with appropriate flags, spins up an executor that runs code remotely.
3. Input, Output and Error redirection handled before creating interpreter.
4. Tests for OOP.
5. CI changes to test for OOP execution.

## Type of change

- [x] New feature

## Testing

1. Most of tests for in-process are being replicated and tested on out-of-process. Tests that are failing are being skipped for now.

## Checklist


mcbarton:
> ## [Codecov](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/658?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research) Report
> Attention: Patch coverage is `38.82353%` with `52 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 77.85%. Comparing base [(`9c94230`)](https://app.codecov.io/gh/compiler-research/CppInterOp/commit/9c9423069529caea5aee79b61580fe9a7b1eb73b?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research) to head [(`e46d9ce`)](https://app.codecov.io/gh/compiler-research/CppInterOp/commit/e46d9ce71fc87653e90791e135d52b17f072227e?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research).
> > Report is 5 commits behind head on main.
> 
> [Files with missing lines](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/658?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)	Patch %	Lines
> [lib/CppInterOp/CppInterOpInterpreter.h](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/658?src=pr&el=tree&filepath=lib%2FCppInterOp%2FCppInterOpInterpreter.h&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research#diff-bGliL0NwcEludGVyT3AvQ3BwSW50ZXJPcEludGVycHJldGVyLmg=)	22.91%	[37 Missing ⚠️ ](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/658?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)
> [lib/CppInterOp/CppInterOp.cpp](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/658?src=pr&el=tree&filepath=lib%2FCppInterOp%2FCppInterOp.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research#diff-bGliL0NwcEludGVyT3AvQ3BwSW50ZXJPcC5jcHA=)	62.06%	[11 Missing ⚠️ ](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/658?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)
> [lib/CppInterOp/Compatibility.h](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/658?src=pr&el=tree&filepath=lib%2FCppInterOp%2FCompatibility.h&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research#diff-bGliL0NwcEludGVyT3AvQ29tcGF0aWJpbGl0eS5o)	50.00%	[4 Missing ⚠️ ](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/658?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)
> Additional details and impacted files
> 🚀 New features to boost your workflow:

@kr-2003 please rebase your PR, and address the lack of test coverage for your PR

kr-2003:
The code coverage for CppInterOp is done via ``ubu22-arm-gcc12-clang-repl-20-coverage``.
But, OOP JIT only works for ``ubuntu-x86_64`` and ``macos-arm``. Therefore my changes in CppInterOp that are related to OOP are not covered by codecov.

Is it possible for us to move code coverage from ubuntu-arm to ubuntu-x86??

And I think it should be more reasonable to have code coverage done via the compute that has more chances of supporting any change, i.e. ``ubuntu-x86_64``??

CMakeLists.txt
```diff
@@ -391,10 +395,16 @@ string(REPLACE "-Wcovered-switch-default" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}
...
 string(REPLACE "." ";" VERSION_LIST "${CPPINTEROP_VERSION}")
+string(REPLACE "\;" ";" VERSION_LIST "${VERSION_LIST}")
+
 list(GET VERSION_LIST 0 CPPINTEROP_VERSION_MAJOR)
 list(GET VERSION_LIST 1 CPPINTEROP_VERSION_MINOR)
 list(GET VERSION_LIST 2 CPPINTEROP_VERSION_PATCH)
 
+if(NOT CPPINTEROP_VERSION_PATCH STREQUAL "0")
+  add_definitions(-DCPPINTEROP_VERSION_PATCH)
+endif()
```
Vipul-Cariappa: Modification to the `VERSION` file feels like an additional unnecessary step to me. Can't we just `-DLLVM_BUILT_WITH_OOP_JIT` and use this define, instead of checking the VERSION and defining `CPPINTEROP_VERSION_PATCH`?

lib/CppInterOp/Compatibility.h
```diff
@@ -371,6 +439,12 @@ inline void codeComplete(std::vector<std::string>& Results,
 #endif
 }
 
+#ifndef _WIN32
+#ifdef CPPINTEROP_VERSION_PATCH
+inline pid_t getExecutorPID() { return /*llvm*/ getLastLaunchedExecutorPID(); }
```
Vipul-Cariappa: We may also want to expose the `get_nth_pid` from LLVM.

lib/CppInterOp/CppInterOpInterpreter.h
```diff
@@ -145,6 +151,43 @@ class Interpreter {
...
+
+  static std::unique_ptr<FILE, FileDeleter>& getStderrFile() {
+    static std::unique_ptr<FILE, FileDeleter> stderr_file = nullptr;
+    return stderr_file;
+  }
+
+  static bool& getOutOfProcess() {
+    static bool outOfProcess = false;
+    return outOfProcess;
+  }
```
Vipul-Cariappa: I don't think this should be `static`. What if we have multiple instances of the `Interpreter`. Will this work in such use cases?

Vipul-Cariappa (commented):
Few comments for now.
You should try to address the github-actions bot's comments too.

lib/CppInterOp/CppInterOpInterpreter.h
kr-2003: I created ``static`` variables because these files have to be instantiated before creating Interpreter class. And that should happen in this function ``static std::unique_ptr<Interpreter> create``. 
The ``create()`` method is static - it doesn't have access to instance members.
The ``create()`` method needs to initialize the temp files before creating the Interpreter instance

kr-2003: And also, we can't move the instantiation out of ``create`` method, because we have to pass the fd here.
https://github.com/compiler-research/CppInterOp/blob/ec70e5b99f840a96711c57b81b6e337dc89e8ba7/lib/CppInterOp/CppInterOpInterpreter.h#L231
Before, that we need to have the fd's ready.

kr-2003: Yeah, you are right about having multiple instances of the Interpreter. Multiple instances will have the same fds.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -4,13 +4,14 @@ if (EMSCRIPTEN)
   # Omitting CUDATest.cpp since Emscripten build currently has no GPU support
   # For Emscripten builds linking to gtest_main will not suffice for gtest to run
   # the tests and an explicitly main.cpp is needed
-  set(EXTRA_TEST_SOURCE_FILES main.cpp)
```
mcbarton, line 7: This change was only originally needed for Emscripten builds, and you tell me that oop doesn't support oop, so why would this PR include this change?

.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -60,6 +61,9 @@ runs:
           if [[ "${os}" != "macos"* ]]; then
             valgrind --show-error-list=yes --track-origins=yes --error-exitcode=1 unittests/CppInterOp/CppInterOpTests/unittests/bin/${{ env.BUILD_TYPE }}/CppInterOpTests
           fi
+          if [[ "${{ matrix.oop-jit }}" == "On" ]]; then
+            ./unittests/CppInterOp/CppInterOpTests/unittests/bin/${{ env.BUILD_TYPE }}/CppInterOpTests --use-oop-jit
```
mcbarton, line 65: This should be built into the cmake build system, if this flag is needed for some reason when oop is on.

README.md
```diff
@@ -160,6 +160,28 @@ $env:LLVM_DIR= $PWD.Path
...
+              -DLLVM_ENABLE_ASSERTIONS=ON                         \
+              -DCLANG_ENABLE_STATIC_ANALYZER=OFF                  \
+              -DCLANG_ENABLE_ARCMT=OFF                            \
+              -DCLANG_ENABLE_FORMAT=OFF                           \
+              -DCLANG_ENABLE_BOOTSTRAP=OFF                        \
+              ../llvm
+
+## For Linux
+cmake --build . --target clang clang-repl llvm-jitlink-executor orc_rt-x86_64 --parallel $(nproc --all)
+## For MacOS
```
mcbarton: You have put for Linux and MacOS, but have told me that is only supported on Ubuntu x86 and MacOS arm.  This should be made very clear in the instructions.

CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ include(GNUInstallDirs)
 
   ## Set Cmake packages search order
 
+
```
mcbarton: There are a few random added blank lines throughout. Please remove.

CMakeLists.txt
```diff
@@ -298,6 +299,13 @@ include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})
 separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
 add_definitions(${LLVM_DEFINITIONS_LIST})
 
+string(REGEX REPLACE "/build/lib/cmake/llvm$" "" LLVM_SOURCE_DIR "${LLVM_DIR}")
+add_definitions(-DLLVM_SOURCE_DIR="${LLVM_SOURCE_DIR}")
+
+if(LLVM_BUILT_WITH_OOP_JIT)
```
mcbarton, line 304: I cannot see any check throughout the PR, that it is checked that this has only been turned on for systems that support it. Can you show me where this is done?

.github/workflows/main.yml
```diff
@@ -236,14 +255,14 @@ jobs:
...
 
-    - name: Cache LLVM-${{ matrix.clang-runtime }} and ${{ matrix.cling == 'On' && 'Cling' || 'Clang-REPL' }} build
-      uses: actions/cache/save@v4
-      if: ${{ steps.cache.outputs.cache-hit != 'true' }}
-      with:
-        path: |
-          llvm-project
-          ${{ matrix.cling=='On' && 'cling' || '' }}
-        key: ${{ steps.cache.outputs.cache-primary-key }}
+    # - name: Cache LLVM-${{ matrix.clang-runtime }} and ${{ matrix.cling == 'On' && 'Cling' || 'Clang-REPL' }} build
```
mcbarton: This can be uncommented now llvm 16 and 17 support has been dropped.

mcbarton (changes requested):

unittests/CppInterOp/CMakeLists.txt
kr-2003: Because the main.cpp file handles the extra flags passed for gtest. Here in our case: ``--use-oop-jit``. 
I don't understand this line though:
> and you tell me that oop doesn't support oop

mcbarton: I meant why were you changing something that was changing this, so main.cpp is including for non Emscripten builds now. You answered it, but I don't understand why this --use-oop-jit option is handled outside the cmake.

kr-2003: Yeah, you are right that cmake should handle this. But, to have difference between the tests, we are using ``--use-oop-jit`` option. Otherwise, our test util will not be able to understand which tests to run. So, it kind of gives the user power to whether run OOP tests or not. Even with your cmake approach, I don't think that we can get rid of ``--use-oop-jit`` option because my current approach is something like this:
https://github.com/compiler-research/CppInterOp/blob/fda48a4699a7ec89dd29afa51a75c54ae6ee9a80/unittests/CppInterOp/Utils.cpp#L66-L73

kr-2003: So, I need this option to create interpreters differently for both in-process and out-process approach.

README.md
kr-2003: addressed

CMakeLists.txt
kr-2003: addressed

CMakeLists.txt
kr-2003: addressed

.github/workflows/main.yml
kr-2003: ok

PR #660, opened 2025-07-09
Moving code coverage from ubunut-arm to ubuntu-x86

kr-2003:
# Description

This PR makes the code coverage using the ubuntu's x86 machine instead of using arm machine.

## Type of change

- [x] New feature

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,15 +23,6 @@ jobs:
       matrix:
         include:
           # Ubuntu Arm Jobs
-          - name: ubu22-arm-gcc12-clang-repl-20-coverage
-            os: ubuntu-22.04-arm
-            compiler: gcc-12
-            clang-runtime: '20'
-            cling: Off
-            cppyy: Off
-            llvm_enable_projects: "clang"
-            llvm_targets_to_build: "host;NVPTX"
-            coverage: true
           - name: ubu24-arm-gcc12-clang-repl-20
             os: ubuntu-24.04-arm
             compiler: gcc-12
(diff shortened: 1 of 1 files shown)


mcbarton (changes requested):
This was requested as apparantly opp only works ob Ubuntu x86 (still waiting to hear why). It makes more sense to me to move coverage to the oop job in that PR.

kr-2003:
> This was requested as apparantly opp only works ob Ubuntu x86 (still waiting to hear why). It makes more sense to me to move coverage to the oop job in that PR.

The reason why it works for ub-x86 and not ub-arm, is because the LLVM side of code does not handle the ub-arm case. And there might be some issues with ubuntu-arm supporting, that we might not know.

https://github.com/llvm/llvm-project/blob/337bc916435f40b0e6f32e197237f007a4c6ee57/clang/tools/clang-repl/ClangRepl.cpp#L117C1-L135C1

```cpp
    if (SystemTriple.isOSBinFormatELF())
      OrcRuntimePath =
          BasePath.str().str() + "/lib/x86_64-unknown-linux-gnu/liborc_rt.a";
    else if (SystemTriple.isOSBinFormatMachO())
      OrcRuntimePath = BasePath.str().str() + "/lib/darwin/liborc_rt_osx.a";
    else
      return llvm::make_error<llvm::StringError>(
          "Out-of-process execution is not supported on non-unix platforms",
          llvm::inconvertibleErrorCode());
```

anutosh491:
> This was requested as apparantly opp only works ob Ubuntu x86 (still waiting to hear why). It makes more sense to me to move coverage to the oop job in that PR.

This stands discussed in an internal channel between GSoC mentors (me, Vipul, Vassil) and our mentee (Abhinav) .

Currently Abhinav plans on running tests 3 times in the oop build
1) inprocess without valgrind
2) inprocess with valgrind
3) outprocess without valgrind

And we would need to switch to ubuntu-x86 for code coverage

mcbarton:
Yes this can be closed now

PR #661, opened 2025-07-11
Fix bugs related to printing types

Vipul-Cariappa:
# Description

## Fixes # (issue)

Helps fix up to 5 tests in cppyy.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -3645,7 +3645,11 @@ std::string GetFunctionArgDefault(TCppFunction_t func,
   if (PI->hasDefaultArg()) {
     std::string Result;
     llvm::raw_string_ostream OS(Result);
-    Expr* DefaultArgExpr = const_cast<Expr*>(PI->getDefaultArg());
+    Expr* DefaultArgExpr = nullptr;
+    if (PI->hasUninstantiatedDefaultArg())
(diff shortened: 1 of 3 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -517,18 +517,28 @@ std::string GetCompleteName(TCppType_t klass) {
   auto& C = getSema().getASTContext();
   auto* D = (Decl*)klass;
 
+  PrintingPolicy Policy = C.getPrintingPolicy();
```
vgvassilev, line 520: I wonder if it would be a good idea to improve the to-string printing consistency by modifying the default printing policy at `CreateIntepreter` time?

Vipul-Cariappa: If the intention is to; not keep modifying the `PrintingPolicy`, it may not be possible. We need different amounts of information at different times, using different APIs. Where we will have to modify the `PrintingPolicy`.

vgvassilev: Can we open an issue to track this idea?

Vipul-Cariappa: Opened an issue at https://github.com/compiler-research/CppInterOp/issues/662

vgvassilev (approved):
LGTM!

PR #672, opened 2025-07-15
[wip] Add llvm 21 support

mcbarton:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

## Testing

## Checklist


mcbarton:
> Hey @mcbarton ,
> 
> Are you working on this ?
> 
> I think I saw quite some errors errors while trying to build cppinterop against latest llvm 21 few days back.

@anutosh491 With the exception of the Windows native build you can build CppInterOp against llvm 21 with the changes in this PR (Windows fails building llvm). The tests do not pass however. Given the changes in this PR are minimal, then the source of these errors are likely on the llvm side. I will update this PR soon to get rid of the Windows Emscripten cross compile patch since it is no longer needed, and to update the documentation. Not sure I have any spare time at the moment though to work out the source of the failing tests.

Vipul-Cariappa (commented):
LGTM. Will approve if the CI is green.
@mcbarton, does this also test cppyy with LLVM 21?
@vgvassilev, could you take a look at the `.patch` files in this PR?

patches/llvm/Windows-emscripten-clang21-1-CrossCompile.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/llvm/cmake/modules/CrossCompile.cmake b/llvm/cmake/modules/CrossCompile.cmake
```
mcbarton: This patch needs removing now since we no longer need it. The other llvm patch is one we have used for Emscripten builds for both llvm 19 and 20 (it isn't needed for functionality, but helps the xeus-cpp deployment have a clear folder when executing cells). I think we also need to add the exception handling patch to this PR too, so that the deployment will work correctly.

mcbarton:
Once we have the ci green, we probably need to remove a few ci jobs, since we are hitting the maximum cache size again. My suggestion is to remove the Ubuntu x86 llvm 18 and osx x86 llvm 18 jobs. This will scrap us just under the maximum I think, and we can continue to support these from the cppyy repos.

mcbarton:
I have opened up an issue in llvm about the failing Windows build for llvm 21 here https://github.com/llvm/llvm-project/issues/158716

anutosh491:
hey @mcbarton here's the updated pr on llvm for the patch : https://github.com/llvm/llvm-project/pull/132670

Hope this helps if we need to keep it for some more time on llvm 21

mcbarton:
> @mcbarton can you rebase and fix the size threshold? I’d like to get that in and make a release. We should fix the emscripten size in a separate effort to avoid blocking the support of llvm21.

@vgvassilev I have rebased the PR, and removed the check for the size of the Emscripten shared library size, which I fix after release. The only thing that should be left to get the ci to go green would be to fix the native osx jobs which fail with the following error https://github.com/compiler-research/CppInterOp/actions/runs/19036208336/job/54361154631?pr=672#step:12:578 . I tried fixing it about 2 weeks ago without success. I have enabled it so this PR can be edited by maintainers. If anyone knows how to fix, then they can feel free to commit the fix to the PR.

anutosh491:
> would you like me to work out a patch for llvm 21 which allows CppInterOp macos jobs to pass, or try and modify CppInterOp to work with this llvm commit which currently breaks the macos jobs? Not sure if the commit description highlights we can work with this change to llvm.

Wait I'm confused. So it is a cppinterop thing and not a llvm thing ? 

As in were we doing something wrong and we need to start maintaining a patch or is that patch at fault and we need to fix it on llvm's side ?

That looks like a 300 line odd patch and not sure anyone of us knows what it does exactly, so do we really want to maintain it ? Let's fix it the organic way if we can, can we ?

anutosh491:
Why not post our concerns on that PR and address the root cause ?

anutosh491:
Hmm, I am not sure we should maintain random patches which we don't understand or wasn't introduced by one of us :\

I can give it a look, once I back from vacations (Jan 2026)

mcbarton:
Will fix the failing Emscripten jobs after the Christmas break. CppInterOpTest/InProcessJIT.InterpreterTestEmscriptenExceptionHandling passes for llvm 20, but not for llvm 21 when building with wasm exceptions. CppInterOpTest/InProcessJIT.VariableReflectionTestVariableOffsetsWithInheritance fails/crashes for llvm 20, but not for llvm 21 where is passes when building with wasm exceptions.

mcbarton:
> Hey @mcbarton,
> 
> i just went through the commit you had pointed out ([llvm/llvm-project#138234](https://github.com/llvm/llvm-project/pull/138234)) I don't think the commit needs to be reverted, rather we need to adapt as per that commit.
> 
> We just need this introduce this diff to get the test passing on macos
> 
> ```
> diff --git a/lib/CppInterOp/Paths.cpp b/lib/CppInterOp/Paths.cpp
> index f3980c8..bd9099d 100644
> --- a/lib/CppInterOp/Paths.cpp
> +++ b/lib/CppInterOp/Paths.cpp
> @@ -158,7 +158,13 @@ void CopyIncludePaths(const clang::HeaderSearchOptions& Opts,
>    /// User specified include entries.
>    for (unsigned i = 0, e = Opts.UserEntries.size(); i != e; ++i) {
>      const HeaderSearchOptions::Entry& E = Opts.UserEntries[i];
> -    if (E.IsFramework && E.Group != frontend::Angled)
> +    if (E.IsFramework && E.Group != frontend::Angled && E.Group != frontend::System)
> ```
> 
> What happened is clang’s Darwin driver change now passes default framework search paths via -iframework/-internal-iframework, which get stored in HeaderSearchOptions::UserEntries as IsFramework=true with Group=frontend::System (not Angled). cppinterop assumed ‘framework implies Angled’ and aborted.
> 
> So if I add such logs such above
> 
> ```
>     if (E.IsFramework && E.Group != frontend::Angled) {
>       llvm::errs() << "[cppinterop] framework include entry with non-Angled group: "
>                   << "group=" << int(E.Group)
>                   << " path=" << E.Path << "\n";
>     }
> ```
> 
> I would see, this at the tests
> 
> ```
> ......
> 1: [ RUN      ] CppInterOpTest/InProcessJIT.InterpreterTestIncludePaths
> 1: [cppinterop] framework include entry with non-Angled group: group=2 path=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks
> 1: [cppinterop] framework include entry with non-Angled group: group=2 path=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/SubFrameworks
> 1: [cppinterop] framework include entry with non-Angled group: group=2 path=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/Library/Frameworks
> .......
> ```
> 
> This proves we must allow IsFramework entries in the System group (and ideally treat them as -iframework when emitting flags)
> 
> Try it out and hopefully we can remove the patch you had introduced reverting the llvm commit ! Tests pass locally on my maocs after the 1 line diff.

Hi @anutosh491

I have checked 'Allow edits and access to secrets by maintainers' on this PR, so if you want to apply your fix directly to the PR then you can. 

After that we should only have to get the Emscripten workflow to pass. For fixing the Emscripten workflow I don't mind disabling CppInterOpTest/InProcessJIT.VariableReflectionTestVariableOffsetsWithInheritance for llvm 20 when built with wasm exceptions, but we need to fix CppInterOpTest/InProcessJIT.InterpreterTestEmscriptenExceptionHandling for llvm 21 when built with wasm exceptions.

CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ set(CMAKE_CXX_EXTENSIONS NO)
 option(CPPINTEROP_USE_CLING "Use Cling as backend" OFF)
 option(CPPINTEROP_USE_REPL "Use clang-repl as backend" ON)
 option(CPPINTEROP_ENABLE_TESTING "Enable the CppInterOp testing infrastructure." ON)
+if(EMSCRIPTEN)
+  option(CPPINTEROP_ENABLE_WASM_EXCEPTIONS "Link using wasm exceptions flag" OFF)
+endif()
```
anutosh491: I don't think this should be an option here. it's not a xeus-cpp thing. An exception handling model should be consistent. 

- If cppinterop is meant to built with the wasm EH model, all packages around it will be built with the same.

- if cppinterop is meant to built with the emscripten EH model, all packages around it will be built with the same.

That means if we give this option for cppinterop, we need to do the same while building llvm and that 's just more confusion and more maintaince. 

The deal is simple. Emscripten 4-x comes with a better wasm EH model (than the emscripten EH model which was being used) and hence its recommended to use that, we use just that and don't give no options !

anutosh491: > but we need to fix CppInterOpTest/InProcessJIT.InterpreterTestEmscriptenExceptionHandling for llvm 21 when built with wasm exceptions.

As we're switching the EH model here. We just use the flag for wasm exceptions instead of emscripten exceptions 

```
--- a/unittests/CppInterOp/InterpreterTest.cpp
+++ b/unittests/CppInterOp/InterpreterTest.cpp
@@ -183,10 +183,8 @@ TYPED_TEST(CppInterOpTest, InterpreterTestEmscriptenExceptionHandling) {
   std::vector<const char*> Args = {
     "-std=c++20",
     "-v",
-    "-fexceptions",
-    "-fcxx-exceptions",
-    "-mllvm", "-enable-emscripten-cxx-exceptions",
-    "-mllvm", "-enable-emscripten-sjlj"
+    "-fwasm-exceptions",
+    "-mllvm", "-wasm-enable-sjlj"
   };
   
```
And that would be it !

We should change the test name too. From `InterpreterTestEmscriptenExceptionHandling` to `InterpreterTestWasmExceptionHandling`

mcbarton: Although quant stack has decided to make the change to building the stack in Emscripten forge with wasm exceptions doesn't mean it needs to force that on others. The build instructions in this repo are very clear about how to build with and without wasm exceptions (and that you need wasm exceptions if you plan to build xeus-cpp), so I don't plan on dropping this option from the PR. The idea that its superior is a matter of opinion.

Putting all this discussion on the option to one side, and changing the test as you suggest (the way it is in the PR already), and building everything with none of my changes in this PR relating to wasm exceptions and just ` EMCC_CFLAGS="-sSUPPORT_LONGJMP=wasm -fwasm-exceptions"` (like emscripten forge does it https://github.com/emscripten-forge/recipes/blob/35f77d9fb19fa5f536e707b3713742f331cf3931/recipes/recipes/emscripten_emscripten-wasm32/activate.sh#L54) will not get the test pass the llvm 21 (I tried this before Christmas).

anutosh491: I've written what I think is the way to go ahead. 

Which is a separate PR getting a working wasm exceptions based build for llvm & cppinterop build and then confirming all the tests pass (using llvm 20.1.8) and also the notebook deployments work (i've pasted that the raytracer doesn't work)

Once that is done we think about getting the emscripten tests right for llvm 21. 

1) The EH model is a choice, **I find literally no use case providing multiple options**. I would just stick to the one on emscripten-forge where about 300-400 packages are being built. Cppinterop is not a special case correct ?
2) The only currently existing emscripten usecase for cppinterop is xeus-cpp-lite and I would stick with whatever xeus-cpp needs.

If that's not the case, I am not sure I'd be up for offering further help on failing tests. All the best.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -111,9 +111,8 @@ void* __clang_Interpreter_SetValueWithAlloc(void* This, void* OutVal,
 #endif
 
 #if CLANG_VERSION_MAJOR > 18
-    extern "C" void __clang_Interpreter_SetValueNoAlloc(void* This,
-                                                        void* OutVal,
-                                                        void* OpaqueType, ...);
+extern "C" void __clang_Interpreter_SetValueNoAlloc(void* This, void* OutVal,
```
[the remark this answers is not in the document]
mcbarton: This change was only a format change made because the clang format workflow kept on failing because of it, despite the PR not even changing these lines, so closing this.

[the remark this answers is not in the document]
mcbarton: This change was only a format change made because the clang format workflow kept on failing because of it, despite the PR not even changing these lines, so closing this.

anutosh491 (approved):
Ahh okay, this doesn't have the emscripten changes anymore ! Should be good then !

PR #678, opened 2025-07-18
fix JitCall codegen for function with lambda as return value

Vipul-Cariappa:
# Description

## Fixes # (issue)

Helps fix a test in cppyy

## Type of change

- [x] Bug fix
- [x] New feature

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1966,9 +1966,24 @@ void collect_type_info(const FunctionDecl* FD, QualType& QT,
   PrintingPolicy Policy(C.getPrintingPolicy());
   Policy.SuppressElaboration = true;
   refType = kNotReference;
-  if (QT->isRecordType() && forArgument) {
-    get_type_as_string(QT, type_name, C, Policy);
-    return;
+  if (QT->isRecordType()) {
+    if (forArgument) {
+      get_type_as_string(QT, type_name, C, Policy);
+      return;
+    }
+    if (auto* CXXRD = QT->getAsCXXRecordDecl()) {
+      if (CXXRD->isLambda()) {
+        std::string fn_name;
+        llvm::raw_string_ostream stream(fn_name);
+        Policy.FullyQualifiedName = true;
+        Policy.SuppressUnwrittenScope = true;
+        FD->getNameForDiagnostic(stream, Policy,
+                                 /*Qualified=*/false);
+        type_name = "__internal_CppInterOp::function<decltype(" + fn_name +
+                    ")>::result_type";
+        return;
+      }
+    }
   }
   if (QT->isFunctionPointerType()) {
     std::string fp_typedef_name;
@@ -3222,6 +3237,17 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
     Args[NumArgs + 1] = nullptr;
     llvm::cl::ParseCommandLineOptions(NumArgs + 1, Args.get());
   }
+
+  I->declare(R"(
+    namespace __internal_CppInterOp {
+    template <typename Signature>
+    struct function;
+    template <typename Res, typename... ArgTypes>
+    struct function<Res(ArgTypes...)> {
+      typedef Res result_type;
+    };
+    }  // namespace __internal_CppInterOp
+  )");
 
   sInterpreters->emplace_back(I, /*Owned=*/true);
 
--- a/unittests/CppInterOp/FunctionReflectionTest.cpp
+++ b/unittests/CppInterOp/FunctionReflectionTest.cpp
@@ -2154,6 +2154,19 @@ TEST(FunctionReflectionTest, GetFunctionCallWrapper) {
   EXPECT_TRUE(obj);
   Cpp::Destruct(obj, MyNameSpace_TemplatedEnum_instantiated);
   obj = nullptr;
+
+  Cpp::Declare(R"(
+    auto get_fn(int x) { return [x](int y){ return x + y; }; }
+  )");
+
+  Cpp::TCppScope_t get_fn = Cpp::GetNamed("get_fn");
+  EXPECT_TRUE(get_fn);
+
+  auto get_fn_callable = Cpp::MakeFunctionCallable(get_fn);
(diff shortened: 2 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1927,9 +1935,27 @@ void collect_type_info(const FunctionDecl* FD, QualType& QT,
...
+    if (forArgument) {
+      get_type_as_string(QT, type_name, C, Policy);
+      return;
+    }
+    if (auto* CXXRD = QT->getAsCXXRecordDecl()) {
+      if (CXXRD->isLambda()) {
+        DeclarationName DFunction = &getASTContext().Idents.get("function");
+        auto result = getSema().getStdNamespace()->lookup(DFunction);
+        if (result.empty())
+          Cpp::Declare("#include <functional>");
```
vgvassilev: I feel uncomfortable including <utility> and now <functional>. I am not sure how else to solve this but is there some way to write our own minimal `std::function`-like class that does what we need it to?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1399,6 +1399,14 @@ TCppScope_t LookupDatamember(const std::string& name, TCppScope_t parent) {
   return 0;
 }
 
+bool IsLambdaClass(TCppType_t type) {
+  QualType QT = QualType::getFromOpaquePtr(type);
+  if (auto* CXXRD = QT->getAsCXXRecordDecl()) {
+    return CXXRD->isLambda();
```
vgvassilev: FYI `isLambdaCallOperator` in clang.

Vipul-Cariappa: [`isLambdaCallOperator`](https://clang.llvm.org/doxygen/ASTLambda_8h_source.html#l00028) takes a `CXXMethodDecl`, but we are using `QualType` to check here.
Also, we are not checking if the function is the lambda call operator; instead, we are checking if the given class is a lambda class.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1931,9 +1939,38 @@ void collect_type_info(const FunctionDecl* FD, QualType& QT,
...
-    get_type_as_string(QT, type_name, C, Policy);
-    return;
+  if (QT->isRecordType()) {
+    if (forArgument) {
+      get_type_as_string(QT, type_name, C, Policy);
+      return;
+    }
+    if (auto* CXXRD = QT->getAsCXXRecordDecl()) {
+      if (CXXRD->isLambda()) {
+        static bool flag = false;
```
vgvassilev: No statics anymore. We have a stack of interpreters. Maybe we feed that as part of `Cpp::CreateIntepreter`...

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3191,6 +3214,19 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
     llvm::cl::ParseCommandLineOptions(NumArgs + 1, Args.get());
   }
 
+  if (GpuArgs.empty()) {
```
vgvassilev: Why is this irrelevant for gpu?

Vipul-Cariappa: This crashes with the `--cuda` flag. Atleast on my machine.

vgvassilev: Why?

Vipul-Cariappa: Apparently the crash is a specific problem with my machine. The CI is green for Ubuntu.
But fails on WASM.

anutosh491:
Weird it says the exception handling test fails. I can recreate the exact same example in the lite deployment 

image

mcbarton:
> Weird ..... it says the exception handling test fails. I can recreate the exact same example in the lite deployment
> 
> https://github.com/compiler-research/CppInterOp/blob/515c1c8c6aa6e5847471df3058f779b4a0725c86/unittests/CppInterOp/InterpreterTest.cpp#L155-L179
> 
> image

@anutosh491 could this have anything to do with the fact that exception handling in CppInterOps deployment doesn't currently work (I'm assuming your example comes from xeus-cpps deployment)? That requires this PR to go in https://github.com/compiler-research/CppInterOp/pull/686 (patch taken from Emscripten forge).

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3191,6 +3214,17 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
...
 
+  I->declare(R"(
+    namespace __internal_CppInterOp {
+    template <typename Signature>
+    struct function;
+    template <typename Res, typename... ArgTypes>
+    struct function<Res(ArgTypes...)> {
+      typedef Res result_type;
+    };
+    }  // namespace __internal_CppInterOp
```
vgvassilev, line 3249: These types of codes can go in a separate header file called `CppInterOpRuntime.h` that gets included at interpreter startup time.

vgvassilev:
> Hi @anutosh491, @mcbarton, Could you please look at this. Why is this failing on Emscripten? The only Emscripten relevant change is the `I->declare(...)` in `Cpp::CreateInterpreter`.

I think this can be `ifdef`-ed if it's too hard to fix for emscripten...

mcbarton:
@Vipul-Cariappa @vgvassilev The llvm 19 Emscripten jobs are now passing the test they were previously failing for, now the exception handling patch has been backported. The ci should hopefully go pass now.

lib/CppInterOp/CppInterOp.cpp
vgvassilev: Do we want to do this as part of this PR or we can do it in a subsequent PR?

mcbarton: I'd rather an issue opened about this, and we fix in a subsequent PR just so we can get the Emsripten ci passing again on main.

vgvassilev: Can you open that issue?

mcbarton: The issue has been opened now.

vgvassilev (approved):
LGTM!

PR #684, opened 2025-07-22
Run browser tests in safari

mcbarton:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

## Testing

## Checklist

### DIFF
--- a/unittests/CppInterOp/CMakeLists.txt
+++ b/unittests/CppInterOp/CMakeLists.txt
@@ -1,3 +1,4 @@
+file(COPY ${CMAKE_SOURCE_DIR}/scripts/browser_tests_safari.py DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
 if (EMSCRIPTEN)
   # So we create a html file, as well as the javascript file
   set(CMAKE_EXECUTABLE_SUFFIX ".html")
--- a/unittests/CppInterOp/CMakeLists.txt
+++ b/unittests/CppInterOp/CMakeLists.txt
@@ -1,4 +1,3 @@
-file(COPY ${CMAKE_SOURCE_DIR}/scripts/browser_tests_safari.py DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
 if (EMSCRIPTEN)
   # So we create a html file, as well as the javascript file
   set(CMAKE_EXECUTABLE_SUFFIX ".html")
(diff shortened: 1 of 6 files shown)


mcbarton:
This PR is ready for review. Safari has no headless mode unlike other browsers, so doing automated testing with it is done a little differently to other browsers. I run emrun with the no_browser option, and then use safari driver with selenium to run the webpage we want.

mcbarton:
@anutosh491 @vgvassilev pinging for review. Also please the review the PR adding Safari browser testing in xeus-cpp. We can detect failures just like for other browsers. The only difference between this and other browsers, is we use safari driver and selenium to open the needed webpage, since Safari doesn't have a headless browser mode.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+file(COPY ${CMAKE_SOURCE_DIR}/scripts/browser_tests_safari.py DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
vgvassilev: Why do we need to copy this in the build folder?

mcbarton: Just so that after I have run the emrun command I don't give a relative path to the python script. Technically you don't need it, but it makes the python command look tidier.

vgvassilev: I'd prefer to not copy it since it can cause invalidation issues in development areas if one gets edited.

mcbarton: The PR has now been updated to use the relative path instead of copying the file. I think I got the relative path correct, but will know for certain once the ci finishes running.

vgvassilev: We probably have some github action magic to give the full path to the source dir.

mcbarton: We could make it nicer in the ci using the environment variable for the github workspace, but feels like its best to use relative path there too, given that is what a user will have to do when following the documentation

mcbarton: I got the relative path correct, but forgot to modify the static library Emscripten tests to use thee relative path in the ci. This has been fixed now, and the ci should pass this time.

vgvassilev (approved):
LGTM!

PR #689, opened 2025-07-30
fix look up of shadow constructor

Vipul-Cariappa:
# Description

Shadow constructor need to be reconstructed for the base classes

## Fixes # (issue)

Helps fix 1 test in cppyy

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -873,12 +873,10 @@ static void GetClassDecls(TCppScope_t klass,
       if (CXXCD->isDeleted())
         continue;
 
-      CXXConstructorDecl* Result =
-          getSema().findInheritingConstructor(SourceLocation(), CXXCD, CUSD);
       // Result is appended to the decls, i.e. CXXRD, iterator
       // non-shadowed decl will be push_back later
       // methods.push_back(Result);
-      (void)Result;
+      getSema().findInheritingConstructor(SourceLocation(), CXXCD, CUSD);
     }
   }
 }


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -855,8 +855,21 @@ static void GetClassDecls(TCppScope_t klass,
...
     else if (auto* USD = dyn_cast<UsingShadowDecl>(DI))
-      if (auto* MD = dyn_cast<DeclType>(USD->getTargetDecl()))
-        methods.push_back(MD);
+      if (auto* MD = dyn_cast<DeclType>(USD->getTargetDecl())) {
+        if (auto* CUSD = dyn_cast<ConstructorUsingShadowDecl>(DI)) {
+          if (auto* CXXCD =
+                  dyn_cast_or_null<CXXConstructorDecl>(CUSD->getTargetDecl())) {
+            CXXConstructorDecl* Result = getSema().findInheritingConstructor(
+                SourceLocation(), CXXCD, CUSD);
+            // Result is appended to the decls, i.e. CXXRD, iterator
```
vgvassilev: The implementation is awkward... Would that mean that the iterators are invalidated?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -855,8 +855,21 @@ static void GetClassDecls(TCppScope_t klass,
     if (auto* MD = dyn_cast<DeclType>(DI))
       methods.push_back(MD);
     else if (auto* USD = dyn_cast<UsingShadowDecl>(DI))
```
vgvassilev: I guess we could simplify the levels of indentation if we did:
```cpp
for (Decl* DI : CXXRD->decls()) {
  if (auto* USD = dyn_cast<UsingShadowDecl>(DI))
    DI = USD->getTargetDecl();
  ...
```

vgvassilev: Why my approach here did not work?

Vipul-Cariappa: We want `ConstructorUsingShadowDecl` to pass to `findInheritingConstructor`. `UsingShadowDecl::getTargetDecl` will return `NamedDecl`, in this case `CXXConstructorDecl`.

vgvassilev: Yes, but then you will do the usual checks with dyn_cast, right?

Vipul-Cariappa: No, we cannot do it, because we require the original `DI` later, in line 862.

vgvassilev: Oh, ok.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -854,9 +854,32 @@ static void GetClassDecls(TCppScope_t klass,
...
+      }
+      if (CXXCD->isDeleted())
+        continue;
+
+      CXXConstructorDecl* Result =
+          getSema().findInheritingConstructor(SourceLocation(), CXXCD, CUSD);
+      // Result is appended to the decls, i.e. CXXRD, iterator
+      // non-shadowed decl will be push_back later
+      // methods.push_back(Result);
+      (void)Result;
```
```suggestion
      // Result is appended to the decls, i.e. CXXRD, iterator
      // non-shadowed decl will be push_back later
      // methods.push_back(Result);
      getSema().findInheritingConstructor(SourceLocation(), CXXCD, CUSD);

```

PR #698, opened 2025-08-11
[Thread-Safety] Add Mutex per Interpreter

Vipul-Cariappa:
Big changes in this PR. Some parts of the explanation might be like a conversation.

## Mutex lock per interpreter.

**Question:** Do we need a per-interpreter lock?
**Answer:** Yes. Ideally, that would improve the performance in multi-threaded cases.

**Question:** Does the InterOp API fully support dispatching queries (some kind of lookup) or compiling code across multiple interpreters at the same time?
**Answer:** Actually, (I guess) No. Our API only supports stack-like access of multiple interpreters.
_Example code that will not work:_
```c++
auto interpreter_1 = Cpp::CreateInterpreter();
Cpp::Compile(R"(
void f() {}
)");
auto f = Cpp::GetNamed("f");

auto interpreter_2 = Cpp::CreateInterpreter();
Cpp::Compile(R"(
void ff() {}
)");

auto ff = Cpp::GetNamed("ff");

auto f_callable = Cpp::MakeFunctionCallable(f); // This fails because:
// MakeFunctionCallable uses the interpreter instance 2 at line
// https://github.com/compiler-research/CppInterOp/blob/dba60ff3d829551b1c86128593ca817af78d67dc/lib/CppInterOp/CppInterOp.cpp#L3148-L3150
```
But you can make the above example work, if the user rotates the [`sInterpreters`](https://github.com/compiler-research/CppInterOp/blob/dba60ff3d829551b1c86128593ca817af78d67dc/lib/CppInterOp/CppInterOp.cpp#L127).
_Example:_
```c++
auto interpreter_1 = Cpp::CreateInterpreter();
Cpp::Compile(R"(
void f() {}
)");
auto f = Cpp::GetNamed("f");

auto interpreter_2 = Cpp::CreateInterpreter();
Cpp::Compile(R"(
void ff() {}
)");

auto ff = Cpp::GetNamed("ff");

Cpp::ActivateInterpreter(interpreter_1); // look at https://github.com/compiler-research/CppInterOp/blob/dba60ff3d829551b1c86128593ca817af78d67dc/lib/CppInterOp/CppInterOp.cpp#L3273-L3287
auto f_callable = Cpp::MakeFunctionCallable(f);
```

**Question:** Ok. So is there a way for InterOp is maintain the rotation thing?
**Answer:** Would not be a easy thing to achieve. From my initial view on the matter, it might be possible. But a more important question; Should we support such usecases?

**Question:** Ok. So given the limitation of using multiple interpreter but only as a stack-access. Do we need a mutex lock per interpreter?
**Answer:** No, the user can only access the top most interpreter at a time. We don't need a mutex per interpreter. (Let me know if my reasoning is wrong somewhere)

## Testing this by running in parallel

gtest does not support running in parallel (I could not find anything with my google search). But there is this project, [gtest-parallel](https://github.com/google/gtest-parallel), that can split the tasks into separate isolated processes. But that is not what we want. We want to split the tests to run in parallel threads that share the same interpreter instance.

## Code recovery RAII

@vgvassilev, I need some pointers on incorporating the codegen error recovery RAII. I don't think we should combine the two into the same class, but let me know your opinion. And how should I go about doing it?

### DIFF
--- a/include/CppInterOp/CppInterOp.h
+++ b/include/CppInterOp/CppInterOp.h
@@ -389,7 +389,7 @@ CPPINTEROP_API std::string GetQualifiedCompleteName(TCppScope_t klass);
 CPPINTEROP_API std::vector<TCppScope_t> GetUsingNamespaces(TCppScope_t scope);
 
 /// Gets the global scope of the whole C++  instance.
-CPPINTEROP_API TCppScope_t GetGlobalScope();
+CPPINTEROP_API TCppScope_t GetGlobalScope(TInterp_t interp = nullptr);
 
 /// Strips the typedef and returns the underlying class, and if the
 /// underlying decl is not a class it returns the input unchanged.
@@ -398,18 +398,28 @@ CPPINTEROP_API TCppScope_t GetUnderlyingScope(TCppScope_t scope);
 /// Gets the namespace or class (by stripping typedefs) for the name
 /// passed as a parameter, and if the parent is not passed,
 /// then global scope will be assumed.
+/// Looks up the name in parent, if parent is nullptr,
+/// interp is used to select the interpreter if multiple in-use.
+/// interp is ignored if parent is non-null.
 CPPINTEROP_API TCppScope_t GetScope(const std::string& name,
-                                    TCppScope_t parent = nullptr);
+                                    TCppScope_t parent = nullptr,
+                                    TInterp_t interp = nullptr);
 
 /// When the namespace is known, then the parent doesn't need
 /// to be specified. This will probably be phased-out in
 /// future versions of the interop library.
-CPPINTEROP_API TCppScope_t GetScopeFromCompleteName(const std::string& name);
+/// interp is used to select the interpreter if multiple in-use.
+CPPINTEROP_API TCppScope_t GetScopeFromCompleteName(const std::string& name,
+                                                    TInterp_t interp = nullptr);
 
 /// This function performs a lookup within the specified parent,
 /// a specific named entity (functions, enums, etcetera).
+/// Looks up the name in parent, if parent is nullptr,
+/// interp is used to select the interpreter if multiple in-use.
+/// interp is ignored if parent is non-null.
 CPPINTEROP_API TCppScope_t GetNamed(const std::string& name,
-                                    TCppScope_t parent = nullptr);
+                                    TCppScope_t parent = nullptr,
+                                    TInterp_t interp = nullptr);
 
 /// Gets the parent of the scope that is passed as a parameter.
 CPPINTEROP_API TCppScope_t GetParentScope(TCppScope_t scope);
@@ -493,8 +503,12 @@ CPPINTEROP_API bool IsTemplatedFunction(TCppFunction_t func);
 
 /// This function performs a lookup to check if there is a
 /// templated function of that type.
+/// Looks up the name in parent, if parent is nullptr,
+/// interp is used to select the interpreter if multiple in-use.
+/// interp is ignored if parent is non-null.
 CPPINTEROP_API bool ExistsFunctionTemplate(const std::string& name,
-                                           TCppScope_t parent = nullptr);
+                                           TCppScope_t parent = nullptr,
+                                           TInterp_t interp = nullptr);
 
 /// Sets a list of all the constructor for a scope/class that is
 /// supplied as a parameter.
@@ -542,7 +556,8 @@ CPPINTEROP_API bool IsDestructor(TCppConstFunction_t method);
 CPPINTEROP_API bool IsStaticMethod(TCppConstFunction_t method);
 
 ///\returns the address of the function given its potentially mangled name.
-CPPINTEROP_API TCppFuncAddr_t GetFunctionAddress(const char* mangled_name);
+CPPINTEROP_API TCppFuncAddr_t GetFunctionAddress(const char* mangled_name,
+                                                 TInterp_t interp = nullptr);
 
 ///\returns the address of the function given its function declaration.
 CPPINTEROP_API TCppFuncAddr_t GetFunctionAddress(TCppFunction_t method);
@@ -643,7 +658,9 @@ CPPINTEROP_API TCppType_t GetCanonicalType(TCppType_t type);
 
 /// Used to either get the built-in type of the provided string, or
 /// use the name to lookup the actual type.
-CPPINTEROP_API TCppType_t GetType(const std::string& type);
+/// interp is used to select the interpreter if multiple in-use.
+CPPINTEROP_API TCppType_t GetType(const std::string& type,
+                                  TInterp_t interp = nullptr);
 
 ///\returns the complex of the provided type.
 CPPINTEROP_API TCppType_t GetComplexType(TCppType_t element_type);
@@ -727,10 +744,10 @@ CPPINTEROP_API void UseExternalInterpreter(TInterp_t I);
 
 /// Adds a Search Path for the Interpreter to get the libraries.
 CPPINTEROP_API void AddSearchPath(const char* dir, bool isUser = true,
-                                  bool prepend = false);
+                                  bool prepend = false, TInterp_t I = nullptr);
 
 /// Returns the resource-dir path (for headers).
-CPPINTEROP_API const char* GetResourceDir();
+CPPINTEROP_API const char* GetResourceDir(TInterp_t I = nullptr);
 
 /// Uses the underlying clang compiler to detect the resource directory.
 /// In essence calling clang -print-resource-dir and checks if it ends with
@@ -751,46 +768,52 @@ DetectSystemCompilerIncludePaths(std::vector<std::string>& Paths,
 
 /// Secondary search path for headers, if not found using the
 /// GetResourceDir() function.
-CPPINTEROP_API void AddIncludePath(const char* dir);
+CPPINTEROP_API void AddIncludePath(const char* dir, TInterp_t I = nullptr);
 
 // Gets the currently used include paths
 ///\param[out] IncludePaths - the list of include paths
 ///
 CPPINTEROP_API void GetIncludePaths(std::vector<std::string>& IncludePaths,
                                     bool withSystem = false,
-                                    bool withFlags = false);
+                                    bool withFlags = false,
+                                    TInterp_t I = nullptr);
 
 /// Only Declares a code snippet in \c code and does not execute it.
 ///\returns 0 on success
-CPPINTEROP_API int Declare(const char* code, bool silent = false);
+CPPINTEROP_API int Declare(const char* code, bool silent = false,
+                           TInterp_t I = nullptr);
 
 /// Declares and executes a code snippet in \c code.
 ///\returns 0 on success
-CPPINTEROP_API int Process(const char* code);
+CPPINTEROP_API int Process(const char* code, TInterp_t I = nullptr);
 
 /// Declares, executes and returns the execution result as a intptr_t.
 ///\returns the expression results as a intptr_t.
-CPPINTEROP_API intptr_t Evaluate(const char* code, bool* HadError = nullptr);
+CPPINTEROP_API intptr_t Evaluate(const char* code, bool* HadError = nullptr,
+                                 TInterp_t I = nullptr);
 
 /// Looks up the library if access is enabled.
 ///\returns the path to the library.
-CPPINTEROP_API std::string LookupLibrary(const char* lib_name);
+CPPINTEROP_API std::string LookupLibrary(const char* lib_name,
+                                         TInterp_t I = nullptr);
 
 /// Finds \c lib_stem considering the list of search paths and loads it by
 /// calling dlopen.
 /// \returns true on success.
-CPPINTEROP_API bool LoadLibrary(const char* lib_stem, bool lookup = true);
+CPPINTEROP_API bool LoadLibrary(const char* lib_stem, bool lookup = true,
+                                TInterp_t I = nullptr);
 
 /// Finds \c lib_stem considering the list of search paths and unloads it by
 /// calling dlclose.
 /// function.
-CPPINTEROP_API void UnloadLibrary(const char* lib_stem);
+CPPINTEROP_API void UnloadLibrary(const char* lib_stem, TInterp_t I = nullptr);
 
 /// Scans all libraries on the library search path for a given potentially
 /// mangled symbol name.
 ///\returns the path to the first library that contains the symbol definition.
-CPPINTEROP_API std::string
-SearchLibrariesForSymbol(const char* mangled_name, bool search_system /*true*/);
+CPPINTEROP_API std::string SearchLibrariesForSymbol(const char* mangled_name,
+                                                    bool search_system /*true*/,
+                                                    TInterp_t I = nullptr);
 
 /// Inserts or replaces a symbol in the JIT with the one provided. This is
 /// useful for providing our own implementations of facilities such as printf.
@@ -801,7 +824,8 @@ SearchLibrariesForSymbol(const char* mangled_name, bool search_system /*true*/);
 ///
 ///\returns true on failure.
 CPPINTEROP_API bool InsertOrReplaceJitSymbol(const char* linker_mangled_name,
-                                             uint64_t address);
+                                             uint64_t address,
+                                             TInterp_t I = nullptr);
 
 /// Tries to load provided objects in a string format (prettyprint).
 CPPINTEROP_API std::string ObjToString(const char* type, void* obj);
@@ -838,9 +862,10 @@ GetClassTemplateInstantiationArgs(TCppScope_t templ_instance,
 
 /// Instantiates a function template from a given string representation. This
 /// function also does overload resolution.
+///\param[in] interp - is used to select the interpreter if multiple in-use.
 ///\returns the instantiated function template declaration.
-CPPINTEROP_API TCppFunction_t
-InstantiateTemplateFunctionFromString(const char* function_template);
+CPPINTEROP_API TCppFunction_t InstantiateTemplateFunctionFromString(
+    const char* function_template, TInterp_t interp = nullptr);
 
 /// Finds best overload match based on explicit template parameters (if any)
 /// and argument types.
@@ -937,7 +962,7 @@ CPPINTEROP_API void CodeComplete(std::vector<std::string>& Results,
 /// Reverts the last N operations performed by the interpreter.
 ///\param[in] N The number of operations to undo. Defaults to 1.
 ///\returns 0 on success, non-zero on failure.
-CPPINTEROP_API int Undo(unsigned N = 1);
+CPPINTEROP_API int Undo(unsigned N = 1, TInterp_t interp = nullptr);
 
 } // end namespace Cpp
 
--- a/include/CppInterOp/CppInterOp.h
+++ b/include/CppInterOp/CppInterOp.h
@@ -702,6 +702,11 @@ CreateInterpreter(const std::vector<const char*>& Args = {},
 ///\returns false on failure or if \c I is not tracked in the stack.
 CPPINTEROP_API bool DeleteInterpreter(TInterp_t I = nullptr);
 
+/// Take ownership of an interpreter instance.
+///\param[in] I - the interpreter to be taken, if nullptr, returns the last.
+///\returns nullptr on failure or if \c I is not tracked in the stack.
+CPPINTEROP_API TInterp_t TakeInterpreter(TInterp_t I = nullptr);
+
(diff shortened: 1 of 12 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -86,9 +87,76 @@ using namespace clang;
...
+             "Internal Error: A thread holds lock for the current "
+             "interpreter, but no information available on which thread "
+             "holds the lock.\nPlease report this as a bug.\n");
+    }
+    assert(!CurrentThreadId &&
+           "Internal Error: Lock released but thread id not cleared.\nPlease "
+           "report this as a bug.\n");
+    CurrentThreadId = std::this_thread::get_id();
+    ThreadIdLock.unlock();
+  }
```
Vipul-Cariappa: This complicated code allows us to lock the same mutex multiple times, provided it is the same thread.
Example:
```c++
void foo() {
  RAII lock(mutex);
  clang::WriteToAST();
}

void bar() {
  RAII lock(mutex);
  clang::WriteToAST_2();
  foo(); // foo will try to lock mutex that is already locked. blocking the thread completely.
}

int main() { bar(); }
```

With this change, we check which thread is trying to lock the mutex, and if this thread previously locked the mutex, then it is safe to use without the second lock.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3238,23 +3377,26 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
...
+    I->declare(R"(
+      namespace __internal_CppInterOp {
+      template <typename Signature>
+      struct function;
+      template <typename Res, typename... ArgTypes>
+      struct function<Res(ArgTypes...)> {
+        typedef Res result_type;
+      };
+      }  // namespace __internal_CppInterOp
+    )");
```
Vipul-Cariappa: Note to myself: Did not mean to commit this.

include/CppInterOp/CppInterOp.h
```diff
@@ -643,7 +643,7 @@ CPPINTEROP_API TCppType_t GetCanonicalType(TCppType_t type);
 
 /// Used to either get the built-in type of the provided string, or
 /// use the name to lookup the actual type.
-CPPINTEROP_API TCppType_t GetType(const std::string& type);
+CPPINTEROP_API TCppType_t GetType(const std::string& name);
```
vgvassilev: Can we move these seemingly unrelated changes in a separate PR?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -224,11 +301,18 @@ std::string Demangle(const std::string& mangled_name) {
   return demangle;
 }
 
-void EnableDebugOutput(bool value /* =true*/) { llvm::DebugFlag = value; }
+void EnableDebugOutput(bool value /* =true*/) {
+  LOCK(getInterpInfo());
```
vgvassilev: I had a different mental model for this. I was thinking that we can have a `StateMutatingSection` RAII object which in turn has a mutex-based locking (if multithreading was enabled) and a 'PushTransactionRAII` (for pch/modules). The object should take a parameter if the change is in clang AST which might trigger a derserialiazation or llvm-only. In the latter case we can skip pushing a transaction.

There will be cases where we need one of the two concepts and not both. We need to design a flexible solution addressing these needs without having to introduce both concepts at the same time when they are not needed.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -125,8 +193,17 @@ struct InterpreterInfo {
 
 // std::deque avoids relocations and calling the dtor of InterpreterInfo.
 static llvm::ManagedStatic<std::deque<InterpreterInfo>> sInterpreters;
+static std::mutex InterpreterStackLock;
```
vgvassilev: No statics. This needs to go in the InterpreterInfo class as we will lock per interpreter. That won't be sufficient if the interfaces are changing the llvm global state as done in `EnableDebugOutput`. Fortunately the llvm global state is not a lot and maybe we can figure out how to address that on a later stage...

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -893,17 +996,19 @@ void GetFunctionTemplatedDecls(TCppScope_t klass,
...
+  LOCK(getInterpInfo());
   if (auto* CXXRD = llvm::dyn_cast_or_null<CXXRecordDecl>(D))
     return CXXRD->hasDefaultConstructor();
 
   return false;
 }
 
-TCppFunction_t GetDefaultConstructor(compat::Interpreter& interp,
-                                     TCppScope_t scope) {
+static TCppFunction_t GetDefaultConstructor(compat::Interpreter& interp,
```
vgvassilev: That should be a separate PR.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2270,12 +2400,14 @@ void make_narg_call(const FunctionDecl* FD, const std::string& return_type,
...
       // so check for the most common case: the trivial one, but not uniquely
       // available, while there is a move constructor.
-
-      // include utility header if not already included for std::move
-      DeclarationName DMove = &getASTContext().Idents.get("move");
-      auto result = getSema().getStdNamespace()->lookup(DMove);
-      if (result.empty())
-        Cpp::Declare("#include <utility>");
+      {
+        LOCK(getInterpInfo());
```
vgvassilev: The lock should be done in `Declare`. The next 2-3 lines should not be changing the compiler state (unless modules are involved)...

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2270,12 +2400,14 @@ void make_narg_call(const FunctionDecl* FD, const std::string& return_type,
...
+        // include utility header if not already included for std::move
+        DeclarationName DMove = &getASTContext().Idents.get("move");
+        auto result = getSema().getStdNamespace()->lookup(DMove);
+        if (result.empty())
+          Cpp::Declare("#include <utility>");
```
vgvassilev: That should go in the special runtime header we will include in the interpreter initialization time... Which in turn will end up forcing us to implement our our std::move-like construct.

lib/CppInterOp/CppInterOp.cpp
vgvassilev: Yeah, I remember asking why in the non-GPU version... Please move the fix into another separate PR.

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: I believe we should lock the AST when we do any named lookups. What if some other thread is writing to the AST at the same time? These things should be atomic. Either we finish the read and perform the write, or finish the write and then perform the read. We cannot be writing and reading the symbol table at the same time.
But say we want to check the number of arguments of a function, that would not change once the function is parsed. So we can read that data without any locks.

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: I can do this in the same PR, but as a different commit. We can rebase and merge.

The thing is: We need to compute the path to this initialization file/header. And this should succeed with `make install` and without the `install` step too. So developers can use it without the `install` step.

Let me know if you want a different PR, or if you are fine with me doing it in this PR as a different commit?

vgvassilev: It is pretty similar to the clang resouce header files, eg. `stddef.h`. It is okay to require such a runtime header file -- we can delay that and hardcode it but ideally still needs to happen at the initialization time. The real challenge is testing when/if this goes to llvm -- we cannot have a runtime that depends on stl.

Let's move that discussion in a separate PR.

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: This is the lock for `sInterpreters`. We will need to lock this `sInterpreters` stack whenever we access it. This should be a global lock. `Cpp::CreateInterpreter` pushes the `Cpp::Interpreter` into `sInterpreters`. And if two threads call `Cpp::CreateInterpreter` at the same time, then the insertion should be atomic.

Similarly, we can have another global LLVM lock to take care of the `EnableDebugOutput` cases.

Any comments?

vgvassilev: Good point -- did not realize that...

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -125,8 +133,17 @@ struct InterpreterInfo {
 
 // std::deque avoids relocations and calling the dtor of InterpreterInfo.
 static llvm::ManagedStatic<std::deque<InterpreterInfo>> sInterpreters;
+static std::mutex InterpreterStackLock;
+
```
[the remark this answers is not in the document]
vgvassilev: Can we add the relevant suppression comments for clang-tidy?

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -361,3 +361,22 @@ if (llvm::sys::RunningOnValgrind())
   delete ExtInterp;
 #endif
 }
+
+TEST(InterpreterTest, MultipleInterpreter) {
```
vgvassilev, line 375: We will need a proper readthedocs design document, too :)

mcbarton:
A note for this PR. The Emscripten build of CppInterOp doesn't currently support threads
https://github.com/compiler-research/CppInterOp/blob/6cf1c0662575e59ba7962c537d138d2d52000990/Emscripten-build-instructions.md?plain=1#L95
It is possible to turns threads on, and modify CppInterOp, so that it builds with threads. I did this locally yesterday (I did this in the past before we had tests). The tests can run, but several of our currently passing tests fail with threads enabled. We also get a warning about shared memory for the passing tests. I can put a PR showing my current progress, but I don't know how far I would get enabling past what I already have since shared library with pthreads is labelled as experimental on Emscripten, so we might be trying to fight possible bugs in Emscripten.

unittests/CppInterOp/InterpreterTest.cpp
mcbarton, line 375: Its probably fine to disable this test for llvm 19 Emscripten builds, since the llvm 20 Emscripten build passes. There are already many tests we have passing for llvm 20, but not llvm 19.

lib/.clang-tidy
```diff
@@ -0,0 +1,2 @@
+Checks: >
+  -cppcoreguidelines-avoid-non-const-global-variables
```
vgvassilev: I meant // NOLINTNEXTLINE or something..

mcbarton:
> ## [Codecov](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/698?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research) Report
> 
> ❌ Patch coverage is `94.02985%` with `12 lines` in your changes missing coverage. Please review. ✅ Project coverage is 80.34%. Comparing base ([`6cf1c06`](https://app.codecov.io/gh/compiler-research/CppInterOp/commit/6cf1c0662575e59ba7962c537d138d2d52000990?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)) to head ([`e6a33f8`](https://app.codecov.io/gh/compiler-research/CppInterOp/commit/e6a33f8201a28ae8bc9dcf43491d8057d5045c9b?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)).
> [Files with missing lines](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/698?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research) 	Patch % 	Lines
> [lib/CppInterOp/CppInterOp.cpp](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/698?src=pr&el=tree&filepath=lib%2FCppInterOp%2FCppInterOp.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research#diff-bGliL0NwcEludGVyT3AvQ3BwSW50ZXJPcC5jcHA=) 	94.02% 	[12 Missing ⚠️ ](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/698?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)
> Additional details and impacted files
> 
> 		
> 		
> 		
> 		
> 🚀 New features to boost your workflow:

@Vipul-Cariappa can you add tests so that all lines of your patch are covered by tests?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -129,16 +139,80 @@ struct InterpreterInfo {
   InterpreterInfo& operator=(const InterpreterInfo&) = delete;
 };
 
+// NOLINTBEGIN
```
```suggestion
// NOLINTBEGIN (cppcoreguidelines-avoid-non-const-global-variables)
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -129,16 +139,80 @@ struct InterpreterInfo {
...
 // std::deque avoids relocations and calling the dtor of InterpreterInfo.
-static llvm::ManagedStatic<std::deque<InterpreterInfo>> sInterpreters;
+static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
+    sInterpreters;
+static llvm::ManagedStatic<
+    std::unordered_map<clang::ASTContext*, InterpreterInfo*>>
+    sInterpreterASTMap;
+static std::recursive_mutex InterpreterStackLock;
+static std::recursive_mutex LLVMLock;
+// NOLINTEND
```
```suggestion
// NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -129,16 +139,83 @@ struct InterpreterInfo {
...
+  return *(*sInterpreterASTMap)[&D->getASTContext()]->Interpreter;
 }
+static compat::Interpreter& getInterp(const void* D) {
+  return *getInterpInfo(D).Interpreter;
+}
+
 static clang::Sema& getSema() { return getInterp().getCI()->getSema(); }
+static clang::Sema& getSema(const clang::Decl* D) {
+  if (!D)
+    return getSema();
```
vgvassilev: Do we need the getSema() with no arguments? Ideally we should drop it so that developers don’t rely on it when adding code here.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -129,16 +139,83 @@ struct InterpreterInfo {
...
+  return getInterpInfo(D).Interpreter->getSema();
+}
+static clang::Sema& getSema(const void* D) { return getInterp(D).getSema(); }
+
 static clang::ASTContext& getASTContext() { return getSema().getASTContext(); }
+static clang::ASTContext& getASTContext(const clang::Decl* D) {
+  if (!D)
+    return getASTContext();
```
vgvassilev: Likewise.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3549,16 +3697,20 @@
...
 std::string ObjToString(const char* type, void* obj) {
+  LOCK(getInterpInfo()); // FIXME: not enough information to lock the corrent
+                         // interpreter
   return getInterp().toString(type, obj);
 }
 
-static Decl* InstantiateTemplate(TemplateDecl* TemplateD,
-                                 TemplateArgumentListInfo& TLI, Sema& S,
-                                 bool instantiate_body) {
+Decl* InstantiateTemplate(TemplateDecl* TemplateD,
```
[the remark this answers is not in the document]
vgvassilev: Seems a good suggestion.

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: It is required in things like `Cpp::Declare`. Here, we declare/compile the code into the top-most interpreter in the stack. The call to `getSema` without any arguments resolves this.

Maybe we should change the API to accept an interpreter instance in such cases. If the interpreter instance is null, then we use the top of the stack.
Any thoughts?

vgvassilev: Yes, I want to discourage people using blindly the `getSema()` or `getASTContext()`. We should probably call the other ones explicitly...

vgvassilev (commented):
We are very close to be done. Please address the clang-tidy errors and order the commits with the appropriate commit message.

vgvassilev:
> > ## [Codecov](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/698?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research) Report
> > ❌ Patch coverage is `89.45869%` with `37 lines` in your changes missing coverage. Please review. ✅ Project coverage is 80.29%. Comparing base ([`d85eec0`](https://app.codecov.io/gh/compiler-research/CppInterOp/commit/d85eec031210a9e8ee6b1b74db249d31138ace8b?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)) to head ([`3cf9aec`](https://app.codecov.io/gh/compiler-research/CppInterOp/commit/3cf9aec338e23741c45fdabc2173b41a3741c87d?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)).
> > [Files with missing lines](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/698?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research) 	Patch % 	Lines
> > [lib/CppInterOp/CppInterOp.cpp](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/698?src=pr&el=tree&filepath=lib%2FCppInterOp%2FCppInterOp.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research#diff-bGliL0NwcEludGVyT3AvQ3BwSW50ZXJPcC5jcHA=) 	89.45% 	[37 Missing ⚠️ ](https://app.codecov.io/gh/compiler-research/CppInterOp/pull/698?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=compiler-research)
> > Additional details and impacted files
> > 🚀 New features to boost your workflow:
> 
> @vgvassilev Just over 10% over this PR is not covered by tests according to this message. Maybe @Vipul-Cariappa could add more tests to increase this coverage too before this goes in?

Good point. I’d guess most of the uncovered lines are due to preexisting lack of coverage but would be nice to improve the coverage where possible.

mcbarton:
@Vipul-Cariappa Disabling it is the way to go. We do not currently support threads for the wasm build, and enabling them will quite an involved project from my investigations so far. My progress so far managed to get the tests to run with threads, but we end up with tests failing which currently pass. Also using threads with an Emscripten shared library is currently labelled as experimental.

mcbarton:
> Skipped the test in WASM. @vgvassilev, I guess this can be approved now. To be merged once the CI is green.

I don't know how many commits this is behind main, but it looks like it needs rebasing.

vgvassilev:
> Skipped the test in WASM. @vgvassilev, I guess this can be approved now. To be merged once the CI is green.

You don’t want to spend 30 mins writing these tests to make our coverage happy?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3547,13 +3826,20 @@ bool InsertOrReplaceJitSymbol(compat::Interpreter& I,
...
+  if (interp)
+    return InsertOrReplaceJitSymbol(*interp, linker_mangled_name, address);
+  return InsertOrReplaceJitSymbol(getInterp(NULLPTR), linker_mangled_name,
+                                  address);
 }
 
 std::string ObjToString(const char* type, void* obj) {
-  return getInterp().toString(type, obj);
+  LOCK(getInterpInfo(NULLPTR)); // FIXME: not enough information to lock the
+                                // corrent interpreter
```
```suggestion
                                // current interpreter
```
Please move the comment above the line.

vgvassilev (approved):
Well done!

LGTM, modulo the comments. Consider squashing all into one commit or condensing a good atomic minimal set of commits...

PR #700, opened 2025-08-11
Remove unneeded MakeResourceDir

mcbarton:
# Description

MakeResourceDir is in the compatibility header file despite the fact it is not dependent on what llvm version you are using, or whether you are using cling or clang-repl. Therefore this PR removes is, and just replaces the function in the code.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/lib/CppInterOp/Compatibility.h
+++ b/lib/CppInterOp/Compatibility.h
@@ -398,13 +398,6 @@ inline std::string FixTypeName(const std::string type_name) {
   return type_name;
 }
 
-inline std::string MakeResourceDir(llvm::StringRef Dir) {
-  llvm::SmallString<128> P(Dir);
-  llvm::sys::path::append(P, CLANG_INSTALL_LIBDIR_BASENAME, "clang",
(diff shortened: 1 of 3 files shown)


lib/CppInterOp/Compatibility.h
```diff
@@ -393,13 +393,6 @@ class SynthesizingCodeRAII {
 
 namespace compat {
 
-inline std::string MakeResourceDir(llvm::StringRef Dir) {
```
mcbarton, line 396: An alternative to removing MakeResourceDir like I've done here would be to move it to a different header file. It doesn't really belong in compatibility.h anymore since its not dependent on llvm version or operating system. @vgvassilev Is it ok just removing it, or would you prefer it be moved somewhere else? If its the latter, which header file do you suggest it gets moved to.

vgvassilev: No, this lgtm!

PR #713, opened 2025-09-16
Remove restriction on number of processes for Emscripten static library build on MacOS Github runners

mcbarton:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -706,9 +706,9 @@ jobs:
         fi
         os="${{ matrix.os }}"
         if [[ "${os}" != macos* ]] ; then
-          emmake make -j ${{ env.ncpus }} check-cppinterop
+          EMCC_CORES=1 emmake make -j 1 check-cppinterop
         else
-          EMCC_CORES=2 emmake make -j 2 check-cppinterop
+          emmake make -j ${{ env.ncpus }} check-cppinterop
         fi
         cd ./unittests/CppInterOp/
         # Explaination of options for emrun


.github/workflows/emscripten.yml
```diff
@@ -708,7 +708,7 @@ jobs:
         if [[ "${os}" != macos* ]] ; then
           EMCC_CORES=1 emmake make -j 1 check-cppinterop
         else
-          EMCC_CORES=2 emmake make -j 2 check-cppinterop
+          emmake make -j ${{ env.ncpus }} check-cppinterop
```
Vipul-Cariappa, line 711: Why are we not doing the same for other OS?

mcbarton: Because the Linux jobs still crash the ci if you remove the restriction for them. This PR allows us at least remove the restriction on the MacOS jobs while we are able to. Once this is in I plan to open another PR which removes the restriction on the Linux jobs, and I will just keep rebasing that PR every time a new PR goes in until it passes.

Vipul-Cariappa: So, is it a problem with emscripten?

mcbarton: No you can do these build with enough resources (e.g. locally on your laptop which will have more memory). There is no problem with Emscripten. The Github runners are too limited in resources to allow you to do these builds with multiple jobs (at least for now with Linux).

PR #721, opened 2025-09-24
Fix Thread-Safety

Vipul-Cariappa:
This PR fixes the problems with the old implementation.
We now globally lock all the interpreters when we iterate over all the interpreters' bump allocators to resolve which interpreter a `QualType` belongs to. Similarly, we also lock all the interpreters to resolve the interpreter from a `Decl`.

I have also adopted most of the tests to run in parallel, hopefully detecting failures early.


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +150,146 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
...
+// static clang::Sema& getSema(const void* D) { return getInterp(D).getSema(); }
+
+// static clang::ASTContext& getASTContext(const clang::Decl* D) {
+//   return getSema(D).getASTContext();
+// }
+// static clang::ASTContext& getASTContext(const void* D) {
+//   return getSema(D).getASTContext();
+// }
+
+int Declare(compat::Interpreter& I, const char* code, bool silent);
```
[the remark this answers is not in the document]
Vipul-Cariappa: Cannot do this, Ref: https://github.com/compiler-research/CppInterOp/blob/0aa859bb7b122d46564d81ad090d5dcdee00aa56/lib/CppInterOp/CXCppInterOp.cpp#L327

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2045,27 +2183,32 @@ TCppType_t GetType(const std::string& name, TInterp_t interp) {
...
+
+  auto* D = (Decl*)klass;
+  return GetTypeFromScope(D, *getInterpInfo(D));
+}
+
 // Internal functions that are not needed outside the library are
 // encompassed in an anonymous namespace as follows.
 namespace {
-static unsigned long long gWrapperSerial = 0LL;
+static std::atomic<unsigned long long> gWrapperSerial = 0LL;
```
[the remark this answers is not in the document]
vgvassilev: Shouldn't that be per-interpreter state?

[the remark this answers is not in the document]
Vipul-Cariappa: Yes, ideally it should be. But we cannot make it per-interpreter if we want to support cling. At least it is complicated. I say will leave as a `static atomic`.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +158,122 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
 static llvm::ManagedStatic<
     std::unordered_map<clang::ASTContext*, InterpreterInfo*>>
     sInterpreterASTMap;
-static std::recursive_mutex InterpreterStackLock;
-static std::recursive_mutex LLVMLock;
+static int IsSomeThreadWriting = 0;
```
vgvassilev, line 161: We need probably a better name here `IsThreadWriting` is not too bad.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +158,122 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
...
+static std::mutex IsSomeThreadWritingMutex;
+static std::mutex InterpreterStackLock;
+static std::mutex LLVMLock;
```
vgvassilev, line 164: Please add a comment explaining why these variables is fine to be process-wide globals.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +158,122 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
...
 // NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
 
+class LOCK {
```
vgvassilev, line 167: Why all caps?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +158,122 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
...
+    M.unlock();
+  }
+  LOCK() = delete;
+  LOCK(const LOCK&) = delete;
+  LOCK(LOCK&&) = delete;
+  LOCK operator=(const LOCK&) = delete;
+  LOCK operator=(LOCK&&) = delete;
+};
+
+struct WAIT_ON_ALL {
```
vgvassilev, line 190: All caps again, we need doxygen style documentation.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +158,122 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
...
+  // Cpp::Declare("class Klass;") // add a declaration
+  // thread 2:
+  // Cpp::Declare("class Klass{};") // add a definition
+  // thread 3:
+  // auto result = Cpp::GetNamed("Klass");
+  // if (result)
+  //   getInterpInfo(result); // getASTContext may read wrong data
+  //    if definition is being written,
+  //    and the declaration has already been written
+  WAIT_ON_ALL WAIT;
```
vgvassilev, line 228: We should name that as somethingRaii.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +158,122 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
...
-    return getInterpInfo(D);
-  if (sInterpreters->size() == 1)
-    return *sInterpreters->back();
-  for (auto& item : *sInterpreterASTMap) {
-    if (item.first->getAllocator().identifyObject(D))
-      return *item.second;
+  if (auto* D = QT->getAsTagDecl()) {
+    WAIT_ON_ALL WAIT;
+    {
+      std::lock_guard<std::mutex> Lock(InterpreterStackLock);
```
vgvassilev, line 246: Why that needs to be in its own scope and not in the if-stmt scope?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +158,122 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
...
+      return *(*sInterpreterASTMap)[&D->getASTContext()];
+    }
+  }
+  WAIT_ON_ALL WAIT;
```
vgvassilev, line 250: I guess you probably want to make the ctor take a lambda which is in the scope you pass after?

lib/CppInterOp/CppInterOp.cpp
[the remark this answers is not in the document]
vgvassilev: Why `Declare` there is different than for example `process`: https://github.com/compiler-research/CppInterOp/blob/0aa859bb7b122d46564d81ad090d5dcdee00aa56/lib/CppInterOp/CXCppInterOp.cpp#L342

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -148,69 +158,122 @@ static llvm::ManagedStatic<std::deque<std::unique_ptr<InterpreterInfo>>>
...
-  return getSema(D).getASTContext();
-}
-static clang::ASTContext& getASTContext(const void* D) {
-  return getSema(D).getASTContext();
-}
+int Declare(compat::Interpreter& I, const char* code, bool silent);
+static TCppType_t GetTypeFromScope(clang::Decl* D, compat::Interpreter& I);
+static TCppIndex_t GetNumBases(compat::Interpreter& I, clang::CXXRecordDecl* D);
+static TCppScope_t GetBaseClass(clang::CXXRecordDecl* CXXRD, TCppIndex_t ibase);
+static TCppScope_t GetGlobalScope(compat::Interpreter& I);
```
vgvassilev, line 276: Why do we need these forward declarations?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -416,13 +478,14 @@ bool IsComplete(TCppScope_t scope) {
 
   Decl* D = static_cast<Decl*>(scope);
 
-  LOCK(getInterpInfo(D));
+  auto& IF = getInterpInfo(D);
```
vgvassilev, line 481: We should spell out the types when we can't deduce then from the rhs.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -696,16 +763,19 @@ std::string GetQualifiedName(TCppType_t klass) {
 // FIXME: Figure out how to merge with GetCompleteName.
 std::string GetQualifiedCompleteName(TCppType_t klass) {
   auto* D = (Decl*)klass;
-  auto& C = getSema(D).getASTContext();
+  auto& IF = getInterpInfo(D);
+  auto& C = IF.getASTContext();
```
vgvassilev, line 767: Likewise.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -645,15 +709,18 @@ std::string GetName(TCppType_t klass) {
 
 std::string GetCompleteName(TCppType_t klass) {
   auto* D = (Decl*)klass;
-  auto& C = getSema(D).getASTContext();
+  auto& IF = getInterpInfo(D);
+  auto& C = IF.getASTContext();
```
vgvassilev, line 713: Likewise.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -860,38 +936,47 @@ TCppScope_t GetParentScope(TCppScope_t scope) {
   return (TCppScope_t)P;
 }
 
-TCppIndex_t GetNumBases(TCppScope_t klass) {
-  auto* D = (Decl*)klass;
-
+static TCppIndex_t GetNumBases(compat::Interpreter& I,
```
vgvassilev, line 939: Why we have these clones of their counterparts taking `I`. I mean I know why but why only a handful of them and not all?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2045,27 +2167,33 @@ TCppType_t GetType(const std::string& name, TInterp_t interp) {
...
+  auto* D = static_cast<Decl*>(klass);
+  return GetTypeFromScope(D, *getInterpInfo(D));
+}
+
 // Internal functions that are not needed outside the library are
 // encompassed in an anonymous namespace as follows.
 namespace {
-static unsigned long long gWrapperSerial = 0LL;
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+std::atomic<unsigned long long> gWrapperSerial = 0LL;
```
vgvassilev, line 2196: Should that not be unsigned long long?

unittests/CppInterOp/EnumReflectionTest.cpp
```diff
@@ -8,11 +8,13 @@
...
 #include "gtest/gtest.h"
 
+#include <utility>
+
 using namespace TestUtils;
 using namespace llvm;
 using namespace clang;
 
-TEST(ScopeReflectionTest, IsEnumScope) {
+static void EnumReflectionTest_IsEnumScope() {
```
vgvassilev, line 17: These changes are incorrect. We should ignore clang-tidy here.

unittests/CppInterOp/EnumReflectionTest.cpp
```diff
@@ -338,13 +340,13 @@ TEST(EnumReflectionTest, GetEnums) {
...
+  Cpp::Declare(code.c_str(), false, I);
   std::vector<std::string> enumNames1, enumNames2, enumNames3, enumNames4;
-  Cpp::TCppScope_t globalscope = Cpp::GetScope("", 0);
-  Cpp::TCppScope_t Animals_scope = Cpp::GetScope("Animals", 0);
-  Cpp::TCppScope_t myClass_scope = Cpp::GetScope("myClass", 0);
-  Cpp::TCppScope_t unsupported_scope = Cpp::GetScope("myVariable", 0);
+  Cpp::TCppScope_t globalscope = Cpp::GetScope("", nullptr, I);
+  Cpp::TCppScope_t Animals_scope = Cpp::GetScope("Animals", nullptr, I);
+  Cpp::TCppScope_t myClass_scope = Cpp::GetScope("myClass", nullptr, I);
+  Cpp::TCppScope_t unsupported_scope = Cpp::GetScope("myVariable", nullptr, I);
```
vgvassilev, line 349: I'd think the interfaces became very clunky here. We need something better...

include/CppInterOp/CppInterOp.h
```diff
@@ -733,7 +733,7 @@ CPPINTEROP_API bool ActivateInterpreter(TInterp_t I);
 /// Clang-REPL, etcetera). In practice, the selected interpreter should not
 /// matter, since the library will function in the same way.
 ///\returns the current interpreter instance, if any.
-CPPINTEROP_API TInterp_t GetInterpreter();
+CPPINTEROP_API TInterp_t GetInterpreter(TInterp_t I = nullptr);
```
Vipul-Cariappa, line 736: Return value will be the same as argument. This should not be done.

PR #726, opened 2025-10-01
[interp] Register runtime symbols for clang-repl

aaronj0:
Registering this runtime symbol fixes JIT session errors when calling `Cpp::Evaluate` if the process does not see symbols from `libClangCppInterOp.so`.

### DIFF
--- a/unittests/CppInterOp/InterpreterTest.cpp
+++ b/unittests/CppInterOp/InterpreterTest.cpp
@@ -79,8 +79,22 @@ TYPED_TEST(CppInterOpTest, InterpreterTestEvaluate) {
   bool HadError;
   EXPECT_TRUE(Cpp::Evaluate("#error", &HadError) == (intptr_t)~0UL);
   EXPECT_TRUE(HadError);
+  // for llvm < 19 this tests different overloads of
+  // __clang_Interpreter_SetValueNoAlloc
   EXPECT_EQ(Cpp::Evaluate("int i = 11; ++i", &HadError), 12);
-  EXPECT_FALSE(HadError) ;
+  EXPECT_FALSE(HadError);
+  EXPECT_EQ(Cpp::Evaluate("double a = 12.; a", &HadError), 12.);
+  EXPECT_FALSE(HadError);
+  EXPECT_EQ(Cpp::Evaluate("float b = 13.; b", &HadError), 13.);
+  EXPECT_FALSE(HadError);
+  EXPECT_EQ(Cpp::Evaluate("long double c = 14.; c", &HadError), 14.);
+  EXPECT_FALSE(HadError);
+  EXPECT_EQ(Cpp::Evaluate("long double d = 15.; d", &HadError), 15.);
+  EXPECT_FALSE(HadError);
+  EXPECT_EQ(Cpp::Evaluate("unsigned long long e = 16; e", &HadError), 16);
+  EXPECT_FALSE(HadError);
+  EXPECT_NE(Cpp::Evaluate("struct S{} s; s", &HadError), (intptr_t)~0UL);
+  EXPECT_FALSE(HadError);
 }
 
 TYPED_TEST(CppInterOpTest, InterpreterTestDeleteInterpreter) {


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3330,6 +3333,35 @@ CPPINTEROP_API JitCall MakeFunctionCallable(TCppConstFunction_t func) {
 }
 
 namespace {
+inline static bool DefineAbsoluteSymbol(compat::Interpreter& I,
```
vgvassilev: No need of inline in the source file.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3442,6 +3474,11 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
        sInterpreters->back().get()});
 
   assert(sInterpreters->size() == sInterpreterASTMap->size());
+// define __clang_Interpreter_SetValueNoAlloc in the JIT dylib for clang-repl
+#ifndef CPPINTEROP_USE_CLING
+  DefineAbsoluteSymbol(*I, "__clang_Interpreter_SetValueNoAlloc",
+                       (uint64_t)&__clang_Interpreter_SetValueNoAlloc);
```
vgvassilev: We will need the entire runtime not only that function.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -90,6 +90,34 @@
...
+#else
+void* __clang_Interpreter_SetValueWithAlloc(void* This, void* OutVal,
+                                            void* OpaqueType);
+#endif
+
+#if CLANG_VERSION_MAJOR >= 19
+    extern "C" void __clang_Interpreter_SetValueNoAlloc(void* This,
+                                                        void* OutVal,
+                                                        void* OpaqueType, ...);
+#elif CLANG_VERSION_MAJOR == 18
```
```suggestion
#if CLANG_VERSION_MAJOR > 18
    extern "C" void __clang_Interpreter_SetValueNoAlloc(void* This,
                                                        void* OutVal,
                                                        void* OpaqueType, ...);
#else
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -90,6 +90,34 @@
 #include <unistd.h>
 #endif // WIN32
 
+//  Runtime symbols required if the library using JIT (Cpp::Evaluate) does not
+//  link to llvm
+#if !defined(CPPINTEROP_USE_CLING) && !defined(EMSCRIPTEN)
+struct __clang_Interpreter_NewTag {
+} __ci_newtag;
+#if CLANG_VERSION_MAJOR >= 22
```
mcbarton: 22? We don't have llvm 21 support yet, so don't understand this reference to llvm 22 being in the PR.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -90,6 +90,34 @@
...
+extern "C" void* __clang_Interpreter_SetValueWithAlloc(void* This, void* OutVal,
+                                                       void* OpaqueType)
+#else
+void* __clang_Interpreter_SetValueWithAlloc(void* This, void* OutVal,
+                                            void* OpaqueType);
+#endif
+
+#if CLANG_VERSION_MAJOR >= 19
```
mcbarton: Its hard to tell viewing this on Github with the clang tidy comments, but is there a reason this isn't being taken care of in the Compatibility header file?

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -84,6 +84,38 @@ TEST(InterpreterTest, Evaluate) {
   EXPECT_FALSE(HadError) ;
 }
 
+TEST(InterpreterTest, EvaluateExtensive) {
+#ifdef EMSCRIPTEN
+  GTEST_SKIP() << "Test fails for Emscipten builds";
+#endif
```
mcbarton: All the tests at the start of this test are the same as InterpreterTest, Evaluate. Can we not remove that old test, and just replace it with this one?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -90,6 +90,34 @@
 #include <unistd.h>
 #endif // WIN32
 
+//  Runtime symbols required if the library using JIT (Cpp::Evaluate) does not
+//  link to llvm
+#if !defined(CPPINTEROP_USE_CLING) && !defined(EMSCRIPTEN)
+struct __clang_Interpreter_NewTag {
```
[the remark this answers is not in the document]
vgvassilev: We should add the relevant clang-tidy suppression.

lib/CppInterOp/Compatibility.h
```diff
@@ -31,6 +31,34 @@ static inline char* GetEnv(const char* Var_Name) {
 #endif
 }
 
+//  Runtime symbols required if the library using JIT (Cpp::Evaluate) does not
+//  link to llvm
+#if !defined(CPPINTEROP_USE_CLING) && !defined(EMSCRIPTEN)
+struct __clang_Interpreter_NewTag {
+} __ci_newtag;
+#if CLANG_VERSION_MAJOR >= 22
```
vgvassilev:
```suggestion
#if CLANG_VERSION_MAJOR > 21
```

lib/CppInterOp/CppInterOp.cpp
aaronj0: done

lib/CppInterOp/CppInterOp.cpp
aaronj0: This symbol was externed recently upstream. I felt uncomfortable adding a declaration that would be incorrect for newer LLVM. Solved with Vassil's comment to make that check `>21`

lib/CppInterOp/CppInterOp.cpp
aaronj0: These symbols must only be declared once in CppInterOp.cpp under an anonymous namespace. If in moved to that header, this will be redefined in multiple files including `Compatibility.h`

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3169,6 +3197,30 @@ CPPINTEROP_API JitCall MakeFunctionCallable(TCppConstFunction_t func) {
...
+  llvm::orc::ExecutionSession& ES = Jit.getExecutionSession();
+  JITDylib& DyLib = *Jit.getProcessSymbolsJITDylib().get();
+
+  llvm::orc::SymbolMap InjectedSymbols{
+      {ES.intern(linker_mangled_name), ExecutorSymbolDef(ExecutorAddr(address),
+                               JITSymbolFlags::Exported)}
+  };
+
+  if (Error Err = DyLib.define(absoluteSymbols(InjectedSymbols))) {
+    logAllUnhandledErrors(std::move(Err), errs(),
```
vgvassilev, line 3227: ~~That should be covered if we add a test that tries to define the same symbol twice.~~

This is a static function and since the error logic is coming from llvm probably not worth the efforts into testing this branch...

unittests/CppInterOp/InterpreterTest.cpp
aaronj0: Thanks, I've addressed this comment.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3330,6 +3333,37 @@
...
+  char GlobalPrefix = DL.getGlobalPrefix();
+  std::string tmp(linker_mangled_name);
+  if (GlobalPrefix != '\0') {
+    tmp = std::string(1, GlobalPrefix) + tmp;
+  }
+  auto Name = ES.intern(tmp);
+  InjectedSymbols[Name] =
+      ExecutorSymbolDef(ExecutorAddr(address), JITSymbolFlags::Exported);
+
+  if (Error Err = DyLib.define(absoluteSymbols(InjectedSymbols))) {
```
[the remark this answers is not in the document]
vgvassilev: Can you click on "resolve" if these are resolved throughout?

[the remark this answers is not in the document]
aaronj0: done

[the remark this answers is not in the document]
aaronj0: done

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3280,6 +3337,88 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
 
   sInterpreters->emplace_back(I, /*Owned=*/true);
 
+// Define runtime symbols in the JIT dylib for clang-repl
+#if !defined(CPPINTEROP_USE_CLING) && !defined(EMSCRIPTEN)
+  DefineAbsoluteSymbol(*I, "__ci_newtag",
+                       reinterpret_cast<uint64_t>(&__ci_newtag));
+// llvm > 22 has this defined as a C symbol that does not require mangling
+#if CLANG_VERSION_MAJOR >= 22
```
mcbarton: @aaronj0 It still feels very weird to have a reference to llvm 22 in the repo, before we even have llvm 21 support. If this needs to stay it should be consistent with the other reference to llvm 22 you make in this PR (e.g. #if CLANG_VERSION_MAJOR > 21).

As an aside, if you have managed to get CppInterOp working with llvm 21, do you know what is needed for MacOS to get this PR over the line https://github.com/compiler-research/CppInterOp/pull/672 ? If yes, can you please apply a fix to the PR (I have enabled 'Allow edits and access to secrets by maintainers ')? If you don't know what is exactly causing the failure, but have an idea about what might fix the problem, still feel free to apply the attempted fix to the PR.

aaronj0: Thanks for the catch! I missed updating the max version here and it should match the `#if CLANG_VERSION_MAJOR > 21` seen when declaring the symbols. That should work.

> do you know what is needed for MacOS to get this PR over the line https://github.com/compiler-research/CppInterOp/pull/672

I don't use Mac for development so I'm unaware of the issue you face in your PR. 

> If you don't know what is exactly causing the failure, but have an idea about what might fix the problem, still feel free to apply the attempted fix to the PR.

Unfortunately I won't have any free cycles until the EOY. If some time frees up then sure :)

PR #730, opened 2025-10-15
Add CppInterOp API dispatch mechanism

aaronj0:
This PR defines the mechanism which enables dispatching CppInterOp's API without linking to it, preventing any LLVM or Clang symbols from being leaked into the client application.

This allows us to run cppyy without linking to CppInterOp motivated by the use case in ROOT. Can be used to deploy CppInterOp in an environment where dynamic linking is not favourable and the only option is dlopen'ing `libClangCppInterOp.so` with `RTLD_LOCAL`


unittests/CMakeLists.txt
```diff
@@ -34,24 +34,24 @@ function(add_cppinterop_unittest name)
...
-  target_link_libraries(${name} PRIVATE ${ARG_LIBRARIES} ${gtest_libs} ${link_pthreads_lib})
-endif()
-if(EMSCRIPTEN)
-  # Without this cmake will try and get node to run the html file.
-  # This guarantees that it runs the js file, and uses emsdks node.
-  add_test(NAME cppinterop-${name} COMMAND $ENV{EMSDK_NODE} ${name}.js)
-else()
-  add_test(NAME cppinterop-${name} COMMAND ${name})
-endif()
+  if(WIN32)
```
mcbarton, line 45: You have have an entire commit for this format change to cmake, but I have no idea what it has to do with adding a CppInterOp API for a  dispatch mechanism. Please revert this change and put it a separate PR if you think this change needs to be made

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -80,6 +80,9 @@ endif()
 set_source_files_properties(InterpreterTest.cpp PROPERTIES COMPILE_DEFINITIONS
   "LLVM_BINARY_DIR=\"${LLVM_BINARY_DIR}\""
 )
+set_source_files_properties(DispatchAPITest.cpp PROPERTIES COMPILE_DEFINITIONS
+  "CPPINTEROP_LIB_DIR=\"${CMAKE_BINARY_DIR}/lib/libclangCppInterOp.so\""
```
mcbarton: I'm assuming your going to make this test file inclusion dependent on whether CppInterOp was built as a shared or static library?

mcbarton (dismissed):

unittests/CppInterOp/CMakeLists.txt
mcbarton: The library will only have a .so extension on Linux and  Emscripten. The extension will need to be ${CMAKE_SHARED_LIBRARY_SUFFIX}

include/CppInterOp/CppInterOpDispatch.h
```diff
@@ -0,0 +1,520 @@
...
+
+#define LOAD_CPP_FUNCTION(func_name)                                           \
+  func_name =                                                                  \
+      reinterpret_cast<CppAPIType::func_name>(dlGetProcAddress(#func_name));
+
+// macro that allows declaration and loading of all CppInterOp API functions in
+// a consistent way. This is used as our dispatched API list, along with the
+// name-address pair table
+#define FOR_EACH_CPP_FUNCTION(DO)                                              \
+  DO(CreateInterpreter)                                                        \
```
vgvassilev: This can take a pair <`CreateInterpreter`, `Type`> and in this case you can write the `using` clauses automatically. We can also use clang-tablegen to generate these for every function in the namespace `Cpp`.

unittests/CMakeLists.txt
aaronj0: As this is a draft, I have included commits that will move into a seperate PR

lib/CppInterOp/CppInterOpDispatch.cpp
```diff
@@ -0,0 +1,159 @@
...
+        size_t count = 0;
+        while (api_function_table[count].name != nullptr)
+          ++count;
+        map.reserve(count);
+
+        for (size_t i = 0; i < count; ++i) {
+          map[api_function_table[i].name] = api_function_table[i].address;
+        }
+        return map;
+      }();
```
Vipul-Cariappa: Why do we need this?
Can we use `api_function_table` directly?

lib/CppInterOp/CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ endif()
   add_llvm_library(clangCppInterOp
     DISABLE_LLVM_LINK_LLVM_DYLIB
     CppInterOp.cpp
+    CppInterOpDispatch.cpp
```
vgvassilev: Maybe we should name the infrastructure we built with something closer to what it does, eg. `DyldDispatch`.

include/CppInterOp/CppDispatch.h
```diff
@@ -0,0 +1,597 @@
...
+#define CPPINTEROP_CPPINTEROPDISPATCH_H
+
+#include <cstddef>
+#include <cstdint>
+#include <set>
+#include <string>
+#include <vector>
+
+#include <cstdlib>
+#include <dlfcn.h>
```
aaronj0: needs something like:

```cpp
#ifdef _WIN32
#include <io.h>
#include <Psapi.h>
#define RTLD_DEFAULT ((void *)::GetModuleHandle(NULL))
//#define dlsym(library, function_name) ::GetProcAddress((HMODULE)library, function_name)
#define dlopen(library_name, flags) ::LoadLibrary(library_name)
#define dlclose(library) ::FreeLibrary((HMODULE)library)
#endif
```

include/CppInterOp/CppInterOpDispatch.h
```diff
@@ -0,0 +1,597 @@
...
+#include <dlfcn.h>
+#include <iostream>
+#include <mutex>
+
+#include <CppInterOp/CppInterOp.h>
+
+// Configured by CMake, can be overridden by defining before including this
+// header
+#ifndef CPPINTEROP_LIBRARY_PATH
+#define CPPINTEROP_LIBRARY_PATH "@CPPINTEROP_LIBRARY_PATH@"
```
vgvassilev: That does not seem correct. The library can be installed or relocated.

include/CppInterOp/CppInterOpDispatch.h
vgvassilev: Ping on this. We want to be writing the using clauses automatically.

lib/CppInterOp/CppInterOpDispatch.cpp
```diff
@@ -0,0 +1,144 @@
...
+    {"GetTypeAsString", (__CPP_FUNC)Cpp::GetTypeAsString},
+    {"GetCanonicalType", (__CPP_FUNC)Cpp::GetCanonicalType},
+    {"HasTypeQualifier", (__CPP_FUNC)Cpp::HasTypeQualifier},
+    {"RemoveTypeQualifier", (__CPP_FUNC)Cpp::RemoveTypeQualifier},
+    {"GetUnderlyingType", (__CPP_FUNC)Cpp::GetUnderlyingType},
+    {"IsRecordType", (__CPP_FUNC)Cpp::IsRecordType},
+    {"IsFunctionPointerType", (__CPP_FUNC)Cpp::IsFunctionPointerType},
+    {"GetVariableType", (__CPP_FUNC)Cpp::GetVariableType},
+    {"GetNamed", (__CPP_FUNC)Cpp::GetNamed},
+    {"GetScopeFromType", (__CPP_FUNC)Cpp::GetScopeFromType},
```
vgvassilev: These can be written by a macro, too.

include/CppInterOp/CppInterOpDispatch.h
aaronj0: This is not the build location, but a new flag. The path can be overridden to the installed/relocated location either with CMake or by setting an env var. The header has no way of knowing that...

lib/CppInterOp/CppInterOpDispatch.cpp
aaronj0: Mostly but we need to append the entries for overloaded APIs:

```cpp
    {"MakeFunctionCallable",
     (__CPP_FUNC) static_cast<Cpp::JitCall (*)(Cpp::TCppConstFunction_t)>(
         &Cpp::MakeFunctionCallable)},
    {"GetFunctionAddress",
     (__CPP_FUNC) static_cast<Cpp::TCppFuncAddr_t (*)(Cpp::TCppFunction_t)>(
         &Cpp::GetFunctionAddress)},
 ```

File: include/CppInterOp/CppInterOpDispatch.h (hunk 3 above)
Comment (aaronj0) [reply include/CppInterOp/CppInterOpDispatch.h]: Looks like this is indeed possible.

```cpp
#define EXTERN_CPP_FUNC_SIMPLE(func_name) \
    extern CppAPIType::func_name func_name;

#define EXTERN_CPP_FUNC_OVERLOADED(func_name, signature) \
    extern CppAPIType::func_name func_name;

#define LOAD_CPP_FUNCTION_SIMPLE(func_name) \
    func_name = reinterpret_cast<CppAPIType::func_name>(dlGetProcAddress(#func_name));

#define LOAD_CPP_FUNCTION_OVERLOADED(func_name, signature) \
    func_name = reinterpret_cast<CppAPIType::func_name>(dlGetProcAddress(#func_name));

#define DECLARE_CPP_NULL_SIMPLE(func_name) \
    CppAPIType::func_name func_name = nullptr;

#define DECLARE_CPP_NULL_OVERLOADED(func_name, signature) \
    CppAPIType::func_name func_name = nullptr;

// macro that allows declaration and loading of all CppInterOp API functions in
// a consistent way. This is used as our dispatched API list, along with the
// name-address pair table
#define FOR_EACH_CPP_FUNCTION_SIMPLE(DO)                                       \
  DO(CreateInterpreter)                                                        \
  DO(GetInterpreter)                                                           \
// ...  rest of non overloaded API
  DO(EnableDebugOutput)

#define FOR_EACH_CPP_FUNCTION_OVERLOADED(DO)                                   \
  DO(MakeFunctionCallable, Cpp::JitCall (*)(Cpp::TCppConstFunction_t))        \
  DO(GetFunctionAddress, Cpp::TCppFuncAddr_t (*)(Cpp::TCppFunction_t))

#define EXTRACT_NAME_OVERLOADED(name, sig) name

#define FOR_EACH_CPP_FUNCTION(DO)                                              \
  FOR_EACH_CPP_FUNCTION_SIMPLE(DO)                                             \
  FOR_EACH_CPP_FUNCTION_OVERLOADED(EXTRACT_NAME_OVERLOADED)

#define DECLARE_TYPE_SIMPLE(func_name) \
    using func_name = decltype(&Cpp::func_name);

#define DECLARE_TYPE_OVERLOADED(func_name, signature) \
    using func_name = signature;
namespace CppDispatch {
// Forward all type aliases
using TCppIndex_t = ::Cpp::TCppIndex_t;
...

using JitCall = ::Cpp::JitCall;
} // end namespace CppDispatch

namespace CppAPIType {
    FOR_EACH_CPP_FUNCTION_SIMPLE(DECLARE_TYPE_SIMPLE)
    FOR_EACH_CPP_FUNCTION_OVERLOADED(DECLARE_TYPE_OVERLOADED)
} // end namespace CppAPIType

#undef DECLARE_TYPE_SIMPLE
#undef DECLARE_TYPE_OVERLOADED
```

aaronj0: Here `FOR_EACH_CPP_FUNCTION_SIMPLE` and `FOR_EACH_CPP_FUNCTION_OVERLOADED` serve as the only ground truth for where we keep the API list. If you have an overloaded API, you put it in the overloaded version as `<API name, function pointer type>`

aaronj0: Of course all usages of the previous macro split up:

```diff
- FOR_EACH_CPP_FUNCTION(LOAD_CPP_FUNCTION)
+ FOR_EACH_CPP_FUNCTION_SIMPLE(LOAD_CPP_FUNCTION_SIMPLE)
+ FOR_EACH_CPP_FUNCTION_OVERLOADED(LOAD_CPP_FUNCTION_OVERLOADED)

- FOR_EACH_CPP_FUNCTION(EXTERN_CPP_FUNC)
+ FOR_EACH_CPP_FUNCTION_SIMPLE(EXTERN_CPP_FUNC_SIMPLE)
+ FOR_EACH_CPP_FUNCTION_OVERLOADED(EXTERN_CPP_FUNC_OVERLOADED)
```

But that keeps the whole thing automatic

include/CppInterOp/CppInterOpDispatch.h
vgvassilev: That is still a problem, however, I do not understand why you need to define it if it is provided from the outside...

include/CppInterOp/CppInterOpDispatch.h
vgvassilev: Probably you don't need such a complexity but a X-macro that expands. You can also take the type of the function using `decltype` and probably handle the overloads there, too. It seems that you already use decltype but I do not understand why this can be a simpler macro setup..

include/CppInterOp/CppInterOpDispatch.h
```diff
@@ -0,0 +1,298 @@
...
+  DO(GetFunctionAddress, Cpp::TCppFuncAddr_t (*)(Cpp::TCppFunction_t))
+
+#define EXTRACT_NAME_OVERLOADED(name, sig) name
+
+#define FOR_EACH_CPP_FUNCTION(DO)                                              \
+  FOR_EACH_CPP_FUNCTION_SIMPLE(DO)                                             \
+  FOR_EACH_CPP_FUNCTION_OVERLOADED(EXTRACT_NAME_OVERLOADED)
+
+#define DECLARE_TYPE_SIMPLE(func_name)                                         \
+  using func_name = decltype(&Cpp::func_name);
```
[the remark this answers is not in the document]
vgvassilev: These are good suggestions..

include/CppInterOp/CppDispatch.h
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    handle = dlopen(libPath, RTLD_LOCAL | RTLD_NOW);
+    if (!handle) {
+      std::cerr << "[CppInterOp] Failed to load library from " << libPath
+                << ": " << dlerror() << '\n';
+      return;
+    }
+
+    getCppProcAddress = reinterpret_cast<void* (*)(const char*)>(
```
vgvassilev:
```suggestion
    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
    getCppProcAddress = reinterpret_cast<void* (*)(const char*)>(
```

include/CppInterOp/CppDispatch.h
```diff
@@ -0,0 +1,298 @@
...
+}
+namespace CppDispatch {
+FOR_EACH_CPP_FUNCTION_SIMPLE(EXTERN_CPP_FUNC_SIMPLE)
+FOR_EACH_CPP_FUNCTION_OVERLOADED(EXTERN_CPP_FUNC_OVERLOADED)
+
+/// Initialize all CppInterOp API from the dynamically loaded library
+/// (RTLD_LOCAL) \param[in] customLibPath Optional custom path to
+/// libclangCppInterOp.so \returns true if initialization succeeded, false
+/// otherwise
+inline bool init_functions(const char* customLibPath = nullptr) {
```
vgvassilev: That should be a test case, right?

lib/CppInterOp/CMakeLists.txt
vgvassilev: ping.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -11,6 +11,14 @@ else()
   set(EXTRA_PATH_TEST_BINARIES /CppInterOpTests/unittests/bin/$<CONFIG>/)
 endif()
 
+# Add the DispatchAPITest only when building shared libraries
+if (BUILD_SHARED_LIBS)
+  set_source_files_properties(DispatchAPITest.cpp PROPERTIES COMPILE_DEFINITIONS
+    "CPPINTEROP_LIB_DIR=\"${CMAKE_BINARY_DIR}/lib/libclangCppInterOp${CMAKE_SHARED_LIBRARY_SUFFIX}\""
```
vgvassilev: Please write a test and try to run that test against installed CppInterOp with the build folder deleted.

unittests/CppInterOp/DispatchAPITest.cpp
```diff
@@ -0,0 +1,57 @@
+#include "Utils.h"
+
+#include "CppInterOp/CppInterOpDispatch.h"
+
+#include "gtest/gtest.h"
+
+#include <dlfcn.h>
```
vgvassilev: You probably do not need this header file.

include/CppInterOp/CppInterOpDispatch.h
aaronj0: > Probably you don't need such a complexity but a X-macro that expands.

Can you give me a clearer idea on that?

> You can also take the type of the function using decltype and probably handle the overloads there, too.

Well, yes I can make this simpler if I write the signatures for all API and take the decltype of the function obtained on static_cast. I've split up the macros to avoid static_cast on non-overloaded methods, as that seems like an unnecessary operation...

unittests/CppInterOp/DispatchAPITest.cpp
aaronj0: > You probably do not need this header file.

Thanks, that's a remnant from earlier

lib/CppInterOp/CMakeLists.txt
aaronj0: `#include <CppInterOp/DyldDispatch.h>`?

include/CppInterOp/CppInterOpDispatch.h
vgvassilev: Something like https://github.com/root-project/cling/blob/9f8a2243cca3e26f8ec32daf2c6022265ce5d1af/include/cling/Interpreter/Value.h#L27

lib/CppInterOp/CMakeLists.txt
vgvassilev: Yes, that looks good.

include/CppInterOp/Dispatch.h
```diff
@@ -0,0 +1,252 @@
...
+
+namespace CppInternal {
+namespace Dispatch {
+
+// FIXME: This is required for the types, but we should move the types
+// into a separate namespace and only use that scope (CppImpl::Types)
+using namespace CppImpl;
+
+#define X(name, type) extern CppAPIType::name name;
+CPPINTEROP_API_MAP
```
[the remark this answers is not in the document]
vgvassilev: We will need a NOLINT(cppcoreguidelines-avoid-non-const-global-variables)

lib/CppInterOp/Dispatch.cpp
```diff
@@ -0,0 +1,27 @@
+//------------------------------------------------------------------------------
+// CppInterOp Dispatch Implementation
+// author:  Aaron Jomy <aaron.jomy@cern.ch>
```
vgvassilev:
```suggestion
// CppInterOp Dispatch Implementation
```

include/CppInterOp/Dispatch.h
```diff
@@ -0,0 +1,252 @@
...
+///
+///\returns the function address of the requested API, or nullptr if not found
+extern "C" CPPINTEROP_API void (
+    *CppGetProcAddress(const unsigned char* procname))(void);
+
+// macro that allows declaration and loading of all CppInterOp API functions in
+// a consistent way. This is used as our dispatched API list, along with the
+// name-address pair table
+#define CPPINTEROP_API_MAP                                                     \
+  X(CreateInterpreter, decltype(&CppImpl::CreateInterpreter))                  \
```
vgvassilev:
```suggestion
  DISPATCH_API(CreateInterpreter, decltype(&CppImpl::CreateInterpreter))                  \
```
maybe...

include/CppInterOp/Dispatch.h
```diff
@@ -0,0 +1,252 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CppInterOp Dispatch Mechanism
+// author:  Aaron Jomy <aaron.jomy@cern.ch>
```
vgvassilev:
```suggestion
// CppInterOp Dispatch Mechanism
```

include/CppInterOp/Dispatch.h
```diff
@@ -0,0 +1,252 @@
...
+//
+//===----------------------------------------------------------------------===//
+#ifndef CPPINTEROP_DISPATCH_H
+#define CPPINTEROP_DISPATCH_H
+
+#include <cstdlib>
+#include <dlfcn.h>
+#include <iostream>
+#include <mutex>
+#include <string>
```
vgvassilev: More generic headers come after less generic ones.

include/CppInterOp/Dispatch.h
```diff
@@ -0,0 +1,252 @@
...
+
+// TODO: implement overload that takes an existing opened DL handle
+inline void* dlGetProcAddress(const char* name,
+                              const char* customLibPath = nullptr) {
+  if (!name)
+    return nullptr;
+
+  static std::once_flag loaded;
+  static void* handle = nullptr;
+  static void* (*getCppProcAddress)(const char*) = nullptr;
```
vgvassilev: Maybe add a comment that this is not checked with multiple interpreters setup.

include/CppInterOp/Dispatch.h
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    handle = dlopen(libPath, RTLD_LOCAL | RTLD_NOW);
+    if (!handle) {
+      std::cerr << "[CppInterOp] Failed to load library from " << libPath
+                << ": " << dlerror() << '\n';
+      return;
+    }
+
+    getCppProcAddress = reinterpret_cast<void* (*)(const char*)>(
```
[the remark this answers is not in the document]
vgvassilev: Add a nolint clause here.

include/CppInterOp/CppInterOp.h
```diff
@@ -945,6 +945,9 @@ CPPINTEROP_API int Undo(unsigned N = 1);
 CPPINTEROP_API pid_t GetExecutorPID();
 #endif
 
-} // end namespace Cpp
+} // namespace CppImpl
 
+#ifndef CPPINTEROP_DISPATCH_H
+namespace Cpp = CppImpl;
+#endif
```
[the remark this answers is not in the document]
aaronj0: Not true, used by clients

include/CppInterOp/CppInterOpDispatch.h
aaronj0: Addressed

include/CppInterOp/CppDispatch.h
aaronj0: The windows specific code has been added and compiled but is untested due to the absence of a shared lib build on the CI (or so it seems) The CppInterOpDispatch test executable is not built.

include/CppInterOp/CppInterOpDispatch.h
aaronj0: Dropped

lib/CppInterOp/CppInterOpDispatch.cpp
aaronj0: Addressed

include/CppInterOp/Dispatch.h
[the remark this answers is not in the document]
aaronj0: Done

include/CppInterOp/Dispatch.h
```diff
@@ -0,0 +1,267 @@
...
+//===----------------------------------------------------------------------===//
+#ifndef CPPINTEROP_DISPATCH_H
+#define CPPINTEROP_DISPATCH_H
+
+#include <cstdlib>
+#include <dlfcn.h>
+#include <iostream>
+#include <mutex>
+#include <string>
+
```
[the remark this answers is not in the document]
aaronj0: Addressed

include/CppInterOp/Dispatch.h
```diff
@@ -0,0 +1,267 @@
...
+    }
+  });
+
+  return getCppProcAddress ? getCppProcAddress(name) : nullptr;
+}
+
+// Used for the extern clauses below
+// FIXME: drop the using clauses
+namespace CppAPIType {
+#define X(name, type) using name = type;
```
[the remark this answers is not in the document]
aaronj0: Addressed

include/CppInterOp/Dispatch.h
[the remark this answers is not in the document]
aaronj0: Done

aaronj0:
Not sure how dynamic loading works on emscripten but the following error indicates something wrong with that step for the dispatch test executable:

```
[ 95%] Linking CXX executable CppInterOpTestsDispatch.html
wasm-ld: error: attempted static link of dynamic object ../../lib/libclangCppInterOp.so
em++: error: '/Users/runner/work/CppInterOp/CppInterOp/emsdk/upstream/bin/wasm-ld -o CppInterOpTestsDispatch.wasm CMakeFiles/CppInterOpTestsDispatch.dir/EnumReflectionTest.cpp.o CMakeFiles/CppInterOpTestsDispatch.dir/FunctionReflectionTest.cpp.o CMakeFiles/CppInterOpTestsDispatch.dir/InterpreterTest.cpp.o CMakeFiles/CppInterOpTestsDispatch.dir/JitTest.cpp.o CMakeFiles/CppInterOpTestsDispatch.dir/ScopeReflectionTest.cpp.o CMakeFiles/CppInterOpTestsDispatch.dir/TypeReflectionTest.cpp.o CMakeFiles/CppInterOpTestsDispatch.dir/Utils.cpp.o CMakeFiles/CppInterOpTestsDispatch.dir/VariableReflectionTest.cpp.o CMakeFiles/CppInterOpTestsDispatch.dir/main.cpp.o ../googletest-prefix/src/googletest-build/lib//libgtest.a ../googletest-prefix/src/googletest-build/lib//libgmock.a ../../lib/libclangCppInterOp.so -L/Users/runner/work/CppInterOp/CppInterOp/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten -L/Users/runner/work/CppInterOp/CppInterOp/emsdk/upstream/emscripten/src/lib -lGL-getprocaddr -lal -lhtml5 -lstubs -lnoexit -lc -ldlmalloc -lcompiler_rt -lc++-noexcept -lc++abi-noexcept -lsockets -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr /var/folders/sz/cwqy23057413nkkfyjbx066h0000gn/T/tmp68oq2eqelibemscripten_js_symbols.so --strip-debug --export=_emscripten_stack_alloc --export=__wasm_call_ctors --export=emscripten_stack_get_current --export=_emscripten_stack_restore --export-if-defined=__start_em_asm --export-if-defined=__stop_em_asm --export-if-defined=__start_em_lib_deps --export-if-defined=__stop_em_lib_deps --export-if-defined=__start_em_js --export-if-defined=__stop_em_js --export-if-defined=main --export-if-defined=__main_argc_argv --export-table -z stack-size=65536 --no-growable-memory --initial-heap=16777216 --no-entry --table-base=1 --global-base=1024' failed (returned 1)
make[3]: *** [unittests/CppInterOp/CppInterOpTestsDispatch.html] Error 1
make[2]: *** [unittests/CppInterOp/CMakeFiles/CppInterOpTestsDispatch.dir/all] Error 2
```

PR #731, opened 2025-10-15
Try `find_package(gtest)` when configuring testing

aaronj0:
From https://github.com/root-project/root/pull/20011/commits/72b979a1646b4c758b0cc2393087a132af9816fc Prevents failures in offline builds that cannot fetch gtest. We now first look for gtest and suggest disabling testing if it is not found

To be fixed to work in general in our testing env


vgvassilev:
We found the bug in ROOT and I believe we can close that now.

aaronj0:
Indeed

PR #735, opened 2025-10-21
Update coverage bot to Ubuntu 24.04 x86 bot

mcbarton:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,15 +23,6 @@ jobs:
       matrix:
         include:
           # Ubuntu Arm Jobs
-          - name: ubu22-arm-gcc12-clang-repl-20-coverage
-            os: ubuntu-22.04-arm
-            compiler: gcc-12
-            clang-runtime: '20'
-            cling: Off
-            cppyy: Off
-            llvm_enable_projects: "clang"
-            llvm_targets_to_build: "host;NVPTX"
-            coverage: true
           - name: ubu24-arm-gcc12-clang-repl-20
             os: ubuntu-24.04-arm
             compiler: gcc-12
@@ -66,6 +57,15 @@ jobs:
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
           # Ubuntu X86 Jobs
+          - name: ubu24-x86-gcc12-clang-repl-20-coverage
+            os: ubuntu-24.04
+            compiler: gcc-12
+            clang-runtime: '20'
+            cling: Off
+            cppyy: Off
+            llvm_enable_projects: "clang"
+            llvm_targets_to_build: "host;NVPTX"
+            coverage: true
           - name: ubu24-x86-gcc12-clang-repl-20
             os: ubuntu-24.04
             compiler: gcc-12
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -261,10 +261,13 @@ jobs:
...
         vers="${CC#*-}"
-        lcov --directory build/ --capture --output-file coverage.info --gcov-tool /usr/bin/gcov-${vers}
-        lcov --remove coverage.info '/usr/*' "${HOME}"'/.cache/*' ${{ github.workspace }}'/llvm-project/*' ${{ github.workspace }}'/unittests/*' --output-file coverage.info
+        lcov --directory build/ --capture --output-file coverage.info --gcov-tool /usr/bin/gcov-${vers} --ignore-errors mismatch
+        lcov --remove coverage.info '/usr/*' ${{ github.workspace }}'/llvm-project/*' --ignore-errors unused --output-file coverage.info
+        lcov --remove coverage.info '${{ github.workspace }}/unittests/*' --ignore-errors unused --output-file coverage.info
+        lcov --remove coverage.info  '${{ github.workspace }}/build/*' --ignore-errors unused --output-file coverage.info
         # output coverage data for debugging (optional)
         lcov --list coverage.info
+        rm -rf ./build/
```
mcbarton, line 270: This is a bit of a fudge but without it codecov will ignore the coverage report the ci makes, and makes its own (which will include the unittests folder in the coverage report)

.github/workflows/main.yml
```diff
@@ -261,10 +261,13 @@ jobs:
         # Create lcov report
         # capture coverage info
         vers="${CC#*-}"
-        lcov --directory build/ --capture --output-file coverage.info --gcov-tool /usr/bin/gcov-${vers}
-        lcov --remove coverage.info '/usr/*' "${HOME}"'/.cache/*' ${{ github.workspace }}'/llvm-project/*' ${{ github.workspace }}'/unittests/*' --output-file coverage.info
+        lcov --directory build/ --capture --output-file coverage.info --gcov-tool /usr/bin/gcov-${vers} --ignore-errors mismatch
```
Vipul-Cariappa, line 264: Why do we have `--ignore-errors mismatch` flag?
The `lcov --remove` is fine.

mcbarton: Without these changes the ci crashes when you upgrade, since you get a newer lcov version is more restrictive in what it will allow you to get away with. All the changes here were found to be necessary.

Vipul-Cariappa: Looking at the docs https://manpages.debian.org/unstable/lcov/lcov.1.en.html#mismatch:
I feel we should try and avoid it. Do we have a CI run without that flag?

mcbarton: Avoiding this flag would mean dealing with all the mismatch warnings we currently have in the ci now (see https://github.com/compiler-research/CppInterOp/actions/runs/18694188958/job/53307234747?pr=735#step:13:52). If you see all the warnings they are in the test files, and to me dealing with them doesn't feel like it gives any benefit given they aren't effecting the end result, and they can easily be ignored with a flag which lcov itself suggested.

Vipul-Cariappa: That run still passes. Just gives you a few warnings. I believe that is fine. We should try and fix those warnings instead of ignoring them. Will approve it if you can remove `--ignore-errors` flag. We can fix the warnings later.

mcbarton: That run passes in the link I gave above because` --ignore-errors mismatch` is there 😅 . It will error without that flag. We cannot fix them later. If we remove that flag then they would need to be fixed in this PR.

mcbarton: Without that flag is treats all those warnings as errors. `--ignore-errors mismatch` is literally telling lcov to not treat the mismatches as errors, but warnings instead.

Vipul-Cariappa: Hmm. Ok.

PR #736, opened 2025-10-21
[cmake] Drop unecessary branches, improve format/indenting

aaronj0:


lib/CppInterOp/CMakeLists.txt
```diff
@@ -18,25 +18,28 @@ else()
...
+)
+
+# Set sources based on whether Cling or Clang-REPL is used
+if (CPPINTEROP_USE_CLING)
+  set(LLVM_OPTIONAL_SOURCES ${LLVM_OPTIONAL_SOURCES} ${DLM})
+  set(DLM)
+endif(CPPINTEROP_USE_CLING)
+if (CPPINTEROP_USE_REPL)
+  #Use DLM optional sources
+endif(CPPINTEROP_USE_REPL)
```
mcbarton: I know it was in the cmake previously, but I don't understand why the cmake has an if statement which does nothing. @aaronj0 can you remove it?

aaronj0: Sure, looks like that came from the original commit: https://github.com/compiler-research/CppInterOp/commit/b0ebf800e8a25a6635804cdce75ee2eedb45dcec. Perhaps something that was expected back then. I've dropped that branch and folded the extra condition, so its a lot more readable now. Good catch.

PR #739, opened 2025-10-29
Enable cling based emscripten build

mcbarton:
# Description

If you build roots llvm 20 against cling master with the changes in https://github.com/root-project/root/pull/20231, then you can cross compile a CppInterOp which makes use of cling to wasm using Emscripten. CppInterOps Emscripten tests can build, but the tests fail immediately due to a popen error here in cling https://github.com/root-project/cling/blob/903295516f19014a0097bfd516cb23295ef1a404/lib/Interpreter/CIFactory.cpp#L160

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

## Testing

## Checklist


anutosh491:
+1

I don't see a good reason behind doing this.

P.S: In my PyData Paris last month, I just educated people to move away from xeus-cling and use xeus-cpp because of how actively it is being developed and people can raise bugs and get them addressed. So I don't think I would be someone contributing to any effort or maintainance if this goes in :\

PR #753, opened 2025-11-14
Update emsdk used to 4.0.9

mcbarton:
# Description

This PR updates CppInterOps Emscripten workflows to emsdk version 4.0.9 . With these changes locally CppInterOps Emscripten build passed all tests, but works is needed in xeus-cpp before this PR will be ready to go in.

### DIFF
--- a/cmake/modules/GoogleTest.cmake
+++ b/cmake/modules/GoogleTest.cmake
@@ -34,8 +34,8 @@ endif()
 ExternalProject_Add(
   googletest
   GIT_REPOSITORY https://github.com/google/googletest.git
-  GIT_SHALLOW 1
-  GIT_TAG v1.17.0
+  GIT_SHALLOW FALSE
+  GIT_TAG fa8438ae6b70c57010177de47a9f13d7041a6328
   UPDATE_COMMAND ""
   # # Force separate output paths for debug and release builds to allow easy
   # # identification of correct lib in subsequent TARGET_LINK_LIBRARIES commands
--- a/environment-wasm.yml
+++ b/environment-wasm.yml
@@ -1,9 +1,9 @@
 name: CppInterOp-wasm
 channels:
-  - https://prefix.dev/emscripten-forge-dev
+  - https://prefix.dev/emscripten-forge-4x
   - https://prefix.dev/conda-forge
 dependencies:
-  - emscripten-abi==3.1.73
+  - emscripten-abi==4.0.9
   - nlohmann_json
   - nlohmann_json-abi
   - xeus-lite
--- a/unittests/CppInterOp/FunctionReflectionTest.cpp
+++ b/unittests/CppInterOp/FunctionReflectionTest.cpp
@@ -1547,7 +1547,7 @@ TYPED_TEST(CppInterOpTest, FunctionReflectionTestGetFunctionAddress) {
 
   std::vector<Decl*> Decls;
   std::string code = "int f1(int i) { return i * i; }";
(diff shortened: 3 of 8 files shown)


environment-wasm.yml
```diff
@@ -1,9 +1,9 @@
 name: CppInterOp-wasm
 channels:
-  - https://prefix.dev/emscripten-forge-dev
+  - https://prefix.dev/emscripten-forge-4x
```
mcbarton, line 3: This change has to be made until Emscripten forges emsdk 4x branch is merged into its main.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -1455,7 +1455,7 @@ TYPED_TEST(CppInterOpTest, FunctionReflectionTestGetFunctionAddress) {
 
   std::vector<Decl*> Decls;
   std::string code = "int f1(int i) { return i * i; }";
-  std::vector<const char*> interpreter_args = {"-include", "new"};
+  std::vector<const char*> interpreter_args = {"-include", "new", "-Xclang", "-iwithsysroot/include/compat"};
```
mcbarton, line 1550: Given these extra flags are only needed for the Emscripten case, it might be worth making the args dependent on what platform we are targeting. Same for the other test.

mcbarton:
Its not clear why googletest fails to build in the Emscripten ci with this change (see https://github.com/compiler-research/CppInterOp/actions/runs/19363850250/job/55410413357?pr=753#step:9:310). This error doesn't happen locally for me. I will debug (probably next week).

cmake/modules/GoogleTest.cmake
```diff
@@ -34,8 +34,8 @@ endif()
 ExternalProject_Add(
   googletest
   GIT_REPOSITORY https://github.com/google/googletest.git
-  GIT_SHALLOW 1
-  GIT_TAG v1.17.0
+  GIT_SHALLOW FALSE
+  GIT_TAG fa8438ae6b70c57010177de47a9f13d7041a6328
```
mcbarton, line 38: This seemingly random change is a workaround for this bug in Googletest https://github.com/google/googletest/issues/4762 . I got this workaround looking at what others had done to workaround the bug. We are basically fixed to a commit after v1.17 until v1.18 comes out.

mcbarton:
Things to do as part of this PR
- [ ] Change emsdk to 4.0.18 . I didn't notice Emscripten forges emsdk version in its 4x branch had been changed to 4.0.18
- [ ]  Check whether any of the Emscripten tests which are currently skipped pass with this updated emsdk. 
- [ ] Revert https://prefix.dev/emscripten-forge-4x change before PR goes in, as changes should be in Emscripten forges main branch at that point.

unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -220,7 +220,7 @@ TYPED_TEST(CppInterOpTest, ScopeReflectionTestIsBuiltin) {
   //  "int", "unsigned int", "long", "unsigned long", "long long", "unsigned long long",
   //  "float", "double", "long double", "void"}
 
-  std::vector<const char*> interpreter_args = { "-include", "new" };
+  std::vector<const char*> interpreter_args = { "-include", "new", "-Xclang", "-iwithsysroot/include/compat" };
```
mcbarton: Interestingly this wasn't needed for 4.0.9, but was needed for 4.0.18 even though it appears to be fixing the same problem of `xlocale.h` not being found for the other tests which needed it for 4.0.9

PR #759, opened 2025-11-26
change all `enum` to `enum class`

Vipul-Cariappa:


include/CppInterOp/CppInterOp.h
```diff
@@ -93,18 +105,31 @@ enum Operator : unsigned char {
   OP_Coawait,
 };
 
-enum OperatorArity : unsigned char { kUnary = 1, kBinary, kBoth };
+enum class OperatorArity : unsigned char { kUnary = 1, kBinary, kBoth };
+
+constexpr TruthValue<OperatorArity> operator&(OperatorArity l,
+                                              OperatorArity r) {
+  return OperatorArity(underlying(l) & underlying(r));
```
vgvassilev, line 112: I do not think arity could be xor-ed -- we have already `kBoth`.

include/CppInterOp/CppInterOp.h
```diff
@@ -44,7 +44,19 @@ using TCppFuncAddr_t = void*;
 using TInterp_t = void*;
 using TCppObject_t = void*;
 
-enum Operator : unsigned char {
+template <typename T> struct TruthValue {
```
vgvassilev, line 47: I suspect we can forward declare the operators and implement them in the cpp file and then we could hide this class completely from the interface.

PR #760, opened 2025-11-27
Only run Valgrind on ci jobs we choose ci

mcbarton:
With this change we will now only run the tests under Valgrind for ci jobs we choose. In this PR I have set it to only run for the latest llvm for Ubuntu arm and Ubuntu x86.


Vipul-Cariappa (approved):
LGTM!

Hi @mcbarton, in future, make sure that the oop-jit is running with Valgrind, even if it is not the latest LLVM. I believe with the next LLVM release, we can drop the LLVM patch we have. And things will become easy. You are the CI guy, and I may forgot about this 😄.

PR #762, opened 2025-12-01
Drop smallpt notebook reference from Emscripten build instructions

mcbarton:
Now that the smallpt example notebook has been dropped from xeus-cpp, references to it need to be dropped from CppInterOps Emscripten ci and documentation.


mcbarton:
pinging @vgvassilev @anutosh491 @vgvassilev to approve this PR, so we can get the Emscripten ci passing again. Given the smallpt notebook doesn't exist anymore, the ci will fail until this is in.

PR #767, opened 2026-01-07
Add Cling transactions preventing ROOT crashes in GetNamed/GetAllCppNames

aaronj0:
These appeared when building our cppyy forks inside ROOT: https://github.com/aaronj0/root/commits/cppyy-migration/


vgvassilev (approved):
LGTM -- we should have this in an interface wrapping these push transactions and their clang-repl analogs in a single place.

PR #770, opened 2026-01-12
[cmake][tests] Consolidate common logic to `add_cppinterop_unittest`

aaronj0:


unittests/CMakeLists.txt
```diff
@@ -34,19 +34,32 @@ function(add_cppinterop_unittest name)
...
     # Without this cmake will try and get node to run the html file.
     # This guarantees that it runs the js file, and uses emsdks node.
     add_test(NAME cppinterop-${name} COMMAND $ENV{EMSDK_NODE} ${name}.js)
   else()
     add_test(NAME cppinterop-${name} COMMAND ${name})
+    set(EXTRA_PATH_TEST_BINARIES /bin/$<CONFIG>/)
   endif()
+  set_output_directory(${name}
+    BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/${EXTRA_PATH_TEST_BINARIES}
+    LIBRARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/${EXTRA_PATH_TEST_BINARIES}
```
vgvassilev, line 60: We will have to adjust this to the new path, too:

https://github.com/compiler-research/CppInterOp/blob/c0c0bb34de13ce15dcccb271f1cb9a7475582242/unittests/CppInterOp/TestSharedLib/CMakeLists.txt#L7-L9

mcbarton:
It shouldn't be a time out issue

You'll see the line

DynamicLibraryManagerTests in Firefox

but the test never run. That test should finish almost immediately, so what tells me the path to html file is almost certainly wrong now. If you run a find command in the ci, you should be able to work out what the new path is to DynamicLibraryManagerTests.html

anutosh491:
cc @mcbarton

vgvassilev:
Please squash on merge. That change is good by itself no matter what the other PR has to say.

PR #776, opened 2026-01-14
Update Emscripten jobs to add llvm 21

mcbarton:
During the llvm 21 upgrade PR we couldn't upgrade the Emscripten jobs due to an issue where using wasm exceptions broke between llvm 20 and 21. This PR https://github.com/llvm/llvm-project/issues/175717contains the changes which will enable an llvm 21 Emscripten build with wasm exceptions, once the following llvm issue https://github.com/llvm/llvm-project/issues/175717 opened by Anutosh has been resolved  .


patches/llvm/emscripten-clang21-2-enable_exception_handling.patch
```diff
@@ -0,0 +1,86 @@
...
+                            : -1)
+                    << ": [TU=" << T.TUPart << ", M=" << T.TheModule.get()
+                    << " (" << T.TheModule->getName() << ")]\n");
++#ifdef __EMSCRIPTEN__
++  // Re-apply stored -mllvm options; wasm builds reset LLVM opts in wasm-ld.
++  getCompilerInstance()->parseLLVMArgs();
++#endif
+   if (!IncrExecutor) {
+     auto Err = CreateExecutor();
+     if (Err)
```
anutosh491: i'd say wait it out !

Copy pasting my suggested solution at real time is just too much pressure :laughing: (saying cause that patch still needs work haha)

mcbarton: I plan to. I just thought I'd quickly port your patch to llvm 21, to see if you was going along the right lines, even if the patch needs updating to a refined patch later.

mcbarton: Even though the patch is still to be refined, just thought I'd note that the patch as it current stands doesn't compile when used with llvm 21 https://github.com/compiler-research/CppInterOp/actions/runs/21139130037/job/60793744088?pr=776#step:11:3387 (assumming I didn't make a simple mistake when porting it)

anutosh491: Why would it work ? The patch is applied on latest upstream right ;)

mcbarton: I know your patch is designed to apply cleanly to llvms main, but I tweaked the diff, so that it would apply to llvm 21, so I thought it might at least build.

anutosh491 (changes requested):
Hi @vgvassilev @mcbarton,

All the changes we want have been applied as patches here and I just expect everything to work and the tests to pass here.

But something that concerns me heavily is providing an option to switch between EH model (wasm exceptions vs emscripten exceptions) that is being done through 
`CPPINTEROP_ENABLE_WASM_EXCEPTIONS`

**It is just not worth it**. The reward we get out of it vs the maintanance cost doesn't match at all !

A) Firstly as a product, we shouldn't be giving our users options for anything & everything. It's just an exception handling model, we can just go with 1

B) We've switched to emscripten 4 on cppinterop and using packages from `https://prefix.dev/emscripten-forge-4x` for building the deployment. All the packages on this channel (around 400) have been built using wasm exceptions, just cause we realized that emscripten 4 works well with wasm exceptions (this was mainly done for scipy and other R packages but has its effect on everything on the channel like C++/Rust packages)

C) The reward for all this maintaince is just 1 single cell (cell 3) in our demo notebook.

image

We can obviously solve this cell through different EH models but that should not be the goto. **We should have more confidence in one Exception model we pick and report issues on llvm as and when we find them in the future.** 

Sure, we can do it for both ... but then both model might have different errors in the future that we might have to fix on LLVM's side which will be painful and introduces lot of waiting. 

I suggest let's go ahead with one model and build our confidence in it. That should be enough !

.github/workflows/emscripten.yml
```diff
@@ -481,14 +539,14 @@ jobs:
...
         os="${{ matrix.os }}"
-        if [[ "${os}" != macos* ]] ; then
-          actual_size=$(stat -c%s "./lib/libclangCppInterOp.so")
-          max_size=$((40 * 1024 * 1024))
-          if [[ "$actual_size" -gt "$max_size" ]]; then
-            echo "Error: libclangCppInterOp.so is larger than 40 MB."
-            exit 1
-          fi
-        fi
+        #if [[ "${os}" != macos* ]] ; then
```
vgvassilev: Why did we disable it and not increase the threshold?

mcbarton: Because the LLVM_ABI work caused more symbols to be retained for llvm 21. This is turn caused the size of the shared library size to increase.

mcbarton: The decision to not to retain this section was decided by you in a Discord or Github message a while back when we first encountered the size issue as you didn't want it to be a blocker

vgvassilev: Hm… either I confused myself or something got lost in the translation. I’d rather have that with an increased threshold.

.github/workflows/deploy-pages.yml
```diff
@@ -108,6 +108,7 @@ jobs:
                 -DCMAKE_INSTALL_PREFIX=$PREFIX         \
                 -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON            \
                 -DSYSROOT_PATH=$SYSROOT_PATH               \
+                -DCPPINTEROP_ENABLE_WASM_EXCEPTIONS=ON        \
```
vgvassilev:
```suggestion
                -DCPPINTEROP_ENABLE_WASM_EXCEPTIONS=ON        \
```

We make `CPPINTEROP_ENABLE_WASM_EXCEPTIONS=ON` default and we implement a new option `CPPINTEROP_EXTRA_WASM_FLAGS=<...>` where people can specify `-sSUPPORT_LONGJMP=wasm -fwasm-exceptions` which will be appended and overrule the default wasm exception model...

CMakeLists.txt
```diff
@@ -12,6 +12,14 @@ set(CMAKE_CXX_EXTENSIONS NO)
...
 option(CPPINTEROP_USE_REPL "Use clang-repl as backend" ON)
 option(CPPINTEROP_ENABLE_TESTING "Enable the CppInterOp testing infrastructure." ON)
+if(EMSCRIPTEN)
+  set(
+    CPPINTEROP_EXCEPTION_HANDLING_FLAGS
+    "-sSUPPORT_LONGJMP=wasm -fwasm-exceptions"
+    CACHE STRING
+    "Compiler flags for CPP interop exception handling"
+  )
+endif()
```
vgvassilev, line 17:
```suggestion
option(CPPINTEROP_EXTRA_WASM_FLAGS "Enable custom flags for wasm" "-sSUPPORT_LONGJMP=wasm -fwasm-exceptions")
```

docs/Emscripten-build-instructions.rst
```diff
@@ -121,9 +122,9 @@ and osx
                  -DCMAKE_CXX_FLAGS_RELEASE="-Oz -g0 -DNDEBUG" \
                  -DLLVM_ENABLE_LTO=Full \
                  ../llvm
-   emmake make libclang -j $(nproc --all)
-   emmake make clangInterpreter clangStaticAnalyzerCore -j $(nproc --all)
-   emmake make lldWasm -j $(nproc --all)
+   EMCC_CFLAGS="-sSUPPORT_LONGJMP=wasm -fwasm-exceptions" emmake make libclang -j $(nproc --all)
+   EMCC_CFLAGS="-sSUPPORT_LONGJMP=wasm -fwasm-exceptions" emmake make clangInterpreter clangStaticAnalyzerCore -j $(nproc --all)
+   EMCC_CFLAGS="-sSUPPORT_LONGJMP=wasm -fwasm-exceptions" emmake make lldWasm -j $(nproc --all)
```
vgvassilev: We can revert back these things as now we have them in the variables by default.

CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ set(CMAKE_CXX_EXTENSIONS NO)
 option(CPPINTEROP_USE_CLING "Use Cling as backend" OFF)
 option(CPPINTEROP_USE_REPL "Use clang-repl as backend" ON)
 option(CPPINTEROP_ENABLE_TESTING "Enable the CppInterOp testing infrastructure." ON)
+if(EMSCRIPTEN)
+  set(CPPINTEROP_EXTRA_WASM_FLAGS "-sSUPPORT_LONGJMP=wasm -fwasm-exceptions" CACHE STRING "Extra flags for wasm")
```
mcbarton, line 16: `option(CPPINTEROP_EXTRA_WASM_FLAGS "Enable custom flags for wasm" "-sSUPPORT_LONGJMP=wasm -fwasm-exceptions")` which was suggested didn't work. This works though.

The error you get using option can be found here https://github.com/compiler-research/CppInterOp/actions/runs/21318907457/job/61366057216?pr=776#step:9:324

Emscripten-build-instructions.md
```diff
@@ -101,9 +102,9 @@ emcmake cmake -DCMAKE_BUILD_TYPE=Release \
                         -DCMAKE_CXX_FLAGS_RELEASE="-Oz -g0 -DNDEBUG" \
                         -DLLVM_ENABLE_LTO=Full \
                         ../llvm
-emmake make libclang -j $(nproc --all)
-emmake make clangInterpreter clangStaticAnalyzerCore -j $(nproc --all)
-emmake make lldWasm -j $(nproc --all)
+EMCC_CFLAGS="-sSUPPORT_LONGJMP=wasm -fwasm-exceptions" emmake make libclang -j $(nproc --all)
+EMCC_CFLAGS="-sSUPPORT_LONGJMP=wasm -fwasm-exceptions" emmake make clangInterpreter clangStaticAnalyzerCore -j $(nproc --all)
+EMCC_CFLAGS="-sSUPPORT_LONGJMP=wasm -fwasm-exceptions" emmake make lldWasm -j $(nproc --all)
```
anutosh491: This is probably not needed right ? 

Building clangInterpreter above this line would need building lldwasm anyways.

lib/CppInterOp/CMakeLists.txt
```diff
@@ -131,18 +131,24 @@ if(EMSCRIPTEN)
...
   else()
-    target_link_options(clangCppInterOp
+  target_compile_options(clangCppInterOp
+    PRIVATE "SHELL: ${CPPINTEROP_EXTRA_WASM_FLAGS}"
+  )
+  target_link_options(clangCppInterOp
       PRIVATE "SHELL: -s WASM_BIGINT"
-    )
+      PRIVATE "SHELL: ${CPPINTEROP_EXTRA_WASM_FLAGS}"
+  )
```
anutosh491, line 151: Target compile and link options would need -fwasm-exceptions as it is default.

anutosh491: Ahh sorry my bad. It comes with the extra flags.

unittests/CppInterOp/VariableReflectionTest.cpp
```diff
@@ -384,6 +384,9 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, VariableReflection_VariableOffsetsWithInheritan
 #if CLANG_VERSION_MAJOR == 18 && defined(CPPINTEROP_USE_CLING) &&              \
     defined(_WIN32) && (defined(_M_ARM) || defined(_M_ARM64))
   GTEST_SKIP() << "Test fails with Cling on Windows on ARM";
+#endif
+#ifdef __EMSCRIPTEN__
+  GTEST_SKIP() << "This test crashes for Emscripten builds of CppInterOp";
```
anutosh491, line 389: Hmm, could you paste the crash. Let's see if we can fix !

Emscripten-build-instructions.md
anutosh491: Also once set, there's no need to repeat I'm guessing 
```
export EMCC_CFLAGS="-sSUPPORT_LONGJMP=wasm -fwasm-exceptions"
```
And then we can just have `emmake make ...`

CMakeLists.txt
```diff
@@ -359,6 +362,9 @@ endif()
 if (LLVM_COMPILER_IS_GCC_COMPATIBLE)
   if (APPLE OR EMSCRIPTEN)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-common -Woverloaded-virtual -Wcast-qual -fno-strict-aliasing -Wno-long-long -Wall -W -Wno-unused-parameter -Wwrite-strings")
+    if (EMSCRIPTEN)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fwasm-exceptions")
```
vgvassilev: Shouldn’t this go with in the option default value?

mcbarton: Yes if should, and is what I had originally. The reason it is like this at the moment is because -sSUPPORT_LONGJMP=wasm causes googletest to have a build warning, and we treat building warnings as errors. I planned on fixing is soon after  post release since I don't really have any time go debug the googletest build error in the time we have before release. Would a fixme suffice for now?

unittests/CppInterOp/VariableReflectionTest.cpp
mcbarton: The ci doesn'f give any useful output for the error since we have a minimum size build. I don't have time to debug this locally in the time we have left. If you have time to debug locally though and find a solution i'll try it out in the PR.

Emscripten-build-instructions.md
mcbarton: Yes, lldWasm is not a target we need and was planning to remove it at some point but haven't got around to it. 

I don't plan to apply your export EMCC_CFLAGS suggestion though as that makes the environment variable persistent and if not unset then it would override cppinterops new config flag making the flag pointless. I will put all the make targets on a single line to reduce the diff, but keep the EMCC_CFLAGS to a single command rather than using export

CMakeLists.txt
vgvassilev: Can we filter that one warning to error flag out from gtest? I think we have the logic in clad to copy paste from.

mcbarton: Changed the logic so that -sSUPPORT_LONGJMP=wasm is just filtered out of CMAKE_CXX_FLAGS just for the Emscripten Googletest build. We will see if my quick attempt at a filter works when the ci runs.

vgvassilev (approved):
This looks good to me. Let's wait for @anutosh491 to review it.

anutosh491 (approved):
I think this is good.

Let's see if the deployment notebook works well after we merge this (the raytracer and mamba magic install should be a good test to confirm wasm exceptions)

Also falling short on time to check the error on this
https://github.com/compiler-research/CppInterOp/pull/776#discussion_r2724987330 but going through the test, I can somewhat sense what the error is going to be (and I know off the fix already 😄 ) ... but I don't mind getting it in through another PR if someone can help me with the error log.

PR #777, opened 2026-01-15
add library path based on resource directory

Vipul-Cariappa:
# Description

And
Removing `MakeResourcesPath` because it depends on the build system's configuration (at build time) and not on host system/configuration.
Or in the other case where it computes relative to executable is wrong because CppInterOp is a dynamic library and the final executable can be located anywhere on the system.

## Fixes # (issue)

xeus-cpp `LD_LIBRARY_PATH` thing for OpenMP.

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -65,6 +65,7 @@
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/TargetParser/Host.h"
 
 #include <algorithm>
 #include <cassert>
@@ -3316,11 +3317,12 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
                             const std::vector<const char*>& GpuArgs /*={}*/) {
   std::string MainExecutableName = sys::fs::getMainExecutable(nullptr, nullptr);
   std::string ResourceDir = MakeResourcesPath();
-#if defined(__linux__) || defined(__APPLE__)
+  std::string processTargetTriple = llvm::sys::getProcessTriple();
   namespace fs = std::filesystem;
-  if (!fs::is_directory(ResourceDir))
+  if ((!fs::is_directory(ResourceDir)) &&
+      ((processTargetTriple.find("linux") != std::string::npos) ||
+       (processTargetTriple.find("apple") != std::string::npos)))
     ResourceDir = DetectResourceDir();
-#endif
 
   std::vector<const char*> ClingArgv = {"-resource-dir", ResourceDir.c_str(),
                                         "-std=c++14"};
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -3383,7 +3383,9 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
     llvm::cl::ParseCommandLineOptions(NumArgs + 1, Args.get());
   }
 
+#ifndef EMSCRIPTEN
   AddLibrarySearchPaths(ResourceDir, I);
+#endif
 
   I->declare(R"(
     namespace __internal_CppInterOp {
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -3288,7 +3288,6 @@ static std::string MakeResourcesPath() {
   return std::string(P.str());
 }
 
-#ifndef EMSCRIPTEN
 void AddLibrarySearchPaths(const std::string& ResourceDir,
                            compat::Interpreter* I) {
   // the resource-dir can be of the form
@@ -3311,7 +3310,6 @@ void AddLibrarySearchPaths(const std::string& ResourceDir,
                                                  false, false);
   }
 }
-#endif
 } // namespace
 
 TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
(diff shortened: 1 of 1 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3234,36 +3235,38 @@ bool DefineAbsoluteSymbol(compat::Interpreter& I,
...
-} // namespace
 
 TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
                             const std::vector<const char*>& GpuArgs /*={}*/) {
   std::string MainExecutableName = sys::fs::getMainExecutable(nullptr, nullptr);
-  std::string ResourceDir = MakeResourcesPath();
+  std::string ResourceDir = DetectResourceDir();
   std::vector<const char*> ClingArgv = {"-resource-dir", ResourceDir.c_str(),
                                         "-std=c++14"};
   ClingArgv.insert(ClingArgv.begin(), MainExecutableName.c_str());
```
Vipul-Cariappa: The logic here is a bit flawed.
We prepend the `-resource-dir` even if the user specifies a resource dir to use.

I want a first round of review. Then I can fix it.

vgvassilev: That's fine, clang takes the last value. That saves us on iterating over all arguments I guess. We can replace it with a check if the resource-dir was provided...

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3234,36 +3235,35 @@ bool DefineAbsoluteSymbol(compat::Interpreter& I,
   return false;
 }
 #endif
+} // namespace
 
-static std::string MakeResourcesPath() {
```
vgvassilev, line 3263: That works when CppInterOp is build with defined `LLVM_BINARY_DIR` and also works if it is linked against a binary installed where clang usually is installed. It is not totally useless...

vgvassilev: On a second thought it is probably more reliable to use a callback when creating the interpreter which gives us access to the clang::Driver class like it was done here https://github.com/llvm/llvm-project/blob/6309cd8668fc2ae589f156b23f86821f4ce5b7ea/clang/lib/Interpreter/IncrementalExecutor.cpp#L418

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3234,36 +3235,35 @@ bool DefineAbsoluteSymbol(compat::Interpreter& I,
...
+        ResourceDir.substr(0, pos).c_str(), false, false);
+  }
 }
-} // namespace
 
 TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
                             const std::vector<const char*>& GpuArgs /*={}*/) {
   std::string MainExecutableName = sys::fs::getMainExecutable(nullptr, nullptr);
-  std::string ResourceDir = MakeResourcesPath();
+  std::string ResourceDir = DetectResourceDir();
```
vgvassilev: That requires packages to install clang as a hard dependency at runtime even when we can deduce the resource path.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3258,12 +3260,43 @@ static std::string MakeResourcesPath() {
...
+  }
+}
+#endif
 } // namespace
 
 TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
                             const std::vector<const char*>& GpuArgs /*={}*/) {
   std::string MainExecutableName = sys::fs::getMainExecutable(nullptr, nullptr);
   std::string ResourceDir = MakeResourcesPath();
+#if defined(__linux__) || defined(__APPLE__)
```
vgvassilev: We should get the target triple's information at runtime, not what we compiled on.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3291,11 +3292,12 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
                             const std::vector<const char*>& GpuArgs /*={}*/) {
   std::string MainExecutableName = sys::fs::getMainExecutable(nullptr, nullptr);
   std::string ResourceDir = MakeResourcesPath();
-#if defined(__linux__) || defined(__APPLE__)
+  std::string processTargetTriple = llvm::sys::getProcessTriple();
```
```suggestion
  Triple T (llvm::sys::getDefaultTargetTriple());
```

Then T.isOSDarwin()...

Vipul-Cariappa: `getCompilerInstance` needs a compiler, but we need the resource dir to create the compiler. We need to pass the `--resource-dir` flag. This is a chicken and egg problem.

vgvassilev: In that case we should use `llvm::sys::getDefaultTargetTriple()` at least...

vgvassilev: Updated my suggestion...

Vipul-Cariappa: `llvm::sys::getProcessTriple` is specialised for JIT use case. Look at the [docs](https://llvm.org/doxygen/namespacellvm_1_1sys.html#aab375071f641a086c0d7067635ccd3dc) and [docs](https://llvm.org/doxygen/namespacellvm_1_1sys.html#a65fe5377618d2fe881dfbf81e403e906) of `llvm::sys::getDefaultTargetTriple`.

I will change it to use the `Target` type instead of string comparisons.

vgvassilev: Ok, makes sense.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3352,6 +3386,10 @@ TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
     llvm::cl::ParseCommandLineOptions(NumArgs + 1, Args.get());
   }
 
+#ifndef EMSCRIPTEN
```
vgvassilev: The same holds for these, right?

vgvassilev: I'd remove these and add an early exit in the `AddLibrarySearchPaths`

vgvassilev (approved):
Lgtm. Consider squash on merge.

PR #781, opened 2026-01-19
Expand Build_and_Test_CppInterOp action to deal with Emscripten case

mcbarton:
Part of the ongoing effort to make reusable actions. This expands the Build_and_Test_CppInterOp action to deal with Emscripten case


vgvassilev (approved):
LGTM!

PR #786, opened 2026-01-26
Add raytracer notebook to deployment

mcbarton:


anutosh491 (commented):
So the notebook itself was added in a separate Pr but it just wasn't deployed is it ? If yes, then this looks correct!

vgvassilev (approved):
LGTM!

PR #789, opened 2026-01-28
Add `GetDoxygenComment` to Dispatch table

aaronj0:
Required for https://github.com/compiler-research/cppyy-backend/pull/183


Vipul-Cariappa (approved):
In future, this should be done when a new API is introduced, instead of a PR later.

vgvassilev:
> In future, this should be done when a new API is introduced, instead of a PR later.

In a future future, that somehow should generated by tablegen, or we should expand that table on both ends the Dispatch.h and CppInterOp.h...

PR #791, opened 2026-01-30
prioritize resource-dir provided by user over auto-detecting

Vipul-Cariappa:


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3310,12 +3310,24 @@ void AddLibrarySearchPaths(const std::string& ResourceDir,
...
+  }
+  return "";
+}
 } // namespace
 
 TInterp_t CreateInterpreter(const std::vector<const char*>& Args /*={}*/,
                             const std::vector<const char*>& GpuArgs /*={}*/) {
   std::string MainExecutableName = sys::fs::getMainExecutable(nullptr, nullptr);
-  std::string ResourceDir = MakeResourcesPath();
+  std::string ResourceDir = parse_for(Args, "-resource-dir");
```
vgvassilev: Passing an empty resource dir should be wrong on the client side. I guess I do not understand this change.

mcbarton (approved):
I have tested this patch locally, and it was able to remove my hack from the openmp kernels PR. After this is in, it should just be a case of fixing the environment.yml file in my xeus-cpp PR to get the ci to pass without my hack. @Vipul-Cariappa Thanks for all your work solving this issue. I'd wait for @vgvassilev to approve this PR before taking it in though.

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: Among the `Args` passed to `CreateInterpreter`, I try to find the `-resource-dir`. If the passed value is valid, then I use it as it is. Otherwise, I try to detect it using existing methods.

Why is this required?
In Line 3400, we set the necessary library search location explicitly (Because by default, linker flags are ignored in repl).

In xeus-cpp setting, we (CppInterOp) cannot manually detect the correct resource. And the `-resource-dir` passed by xeus-cpp kernel is the correct location. Therefore, I am prioritising it over detecting it within CppInterOp.

Let me know if any other clarifications are required.

vgvassilev: Can you rename the function to extract resource dir argument and also put what you just said as a comment.

vgvassilev (approved):
LGTM!

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3310,12 +3310,30 @@ void AddLibrarySearchPaths(const std::string& ResourceDir,
...
   }
 }
+std::string ExtractArgument(const std::vector<const char*>& Args,
+                            const std::string& Arg) {
+  size_t I = 0;
+  for (auto i = Args.begin(); i != Args.end(); i++) {
+    if ((++I < Args.size()) && (*i == Arg)) {
+      return *(++i);
+    }
+  }
```
```suggestion
  size_t I = 0;
  for (auto i = Args.begin(); i != Args.end(); i++)
    if ((++I < Args.size()) && (*i == Arg))
      return *(++i);
```

PR #793, opened 2026-02-03
fix GetFunctionAddress resolution with explicit name change

conrade-ctc:
# Description

We have stricter compiler/flags in our env, and resolution fails due to two `GetFunctionAddress` definitions. This PR moves the internal function to a different name `GetFunctionAddressFromMangledName` to fix the resolution.

## Type of change

- [ x ] Bug fix
- [ ] New feature

## Testing

In our env, stricter flag for compiler now resolve the issue, no tests added to `CppInterOp`, just a rename on the surface.

## Checklist

- [ x ] I have read the contribution guide recently

### DIFF
--- a/include/CppInterOp/CppInterOp.h
+++ b/include/CppInterOp/CppInterOp.h
@@ -560,7 +560,8 @@ CPPINTEROP_API bool IsStaticMethod(TCppConstFunction_t method);
 CPPINTEROP_API bool IsExplicit(TCppConstFunction_t method);
 
 ///\returns the address of the function given its potentially mangled name.
-CPPINTEROP_API TCppFuncAddr_t GetFunctionAddress(const char* mangled_name);
+CPPINTEROP_API TCppFuncAddr_t
+GetFunctionAddressFromMangledName(const char* mangled_name);
 
 ///\returns the address of the function given its function declaration.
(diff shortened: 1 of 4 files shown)


aaronj0 (approved):
LGTM! Would be good if you could run `clang-format` before we land this

conrade-ctc:
> Are you using the Dispatch.h infrastructure? Otherwise why this particular overload does not work for you?

Just running your test suite causes this on our side, but we are compiling with `-std=c++20`, not `-std=c++17`, and because of that have a number of different compiler flags (partially because of our build setup which has e.g. `-Werror` etc, so is stricter). I *think* (but didn't dig deeply) that it generally results from this change to compiling everything with `-std=c++20`, which causes many more failures in your normal build/test cycle (on our side, but using your documented build instructions with this one change to `CMakeLists.txt`). I just tried to reproduce the specific error we were getting on our side but in your build setup, but as mentioned, there a bunch of other tests failing when I start the process of moving the normal `CppInterOp` toolchain towards the one we're using.

This was the specific error: `error: cannot initialize a parameter of type 'CppImpl::TCppFunction_t' (aka 'void *') with an lvalue of type 'const char[9]'`, and that was for the invocations in the test using `GetFunctionAddress("ret_zero")`.

PR #794, opened 2026-02-03
[cmake] explicitly link against stdc++fs if GCC<9

aaronj0:
Commmit 9d2f4c5d3b11c179c23dbdba15fcc404567e3d08 introduced a dependency on `std::filesystem` which requires explicit linker flags for GCC versions below 9 This fixes a build failure on Alma 8 jobs on ROOT's CI, which still uses GCC8.5


vgvassilev (approved):
Lgtm!

PR #795, opened 2026-02-03
[ci] Run Cppyy tests in multithreaded mode

aaronj0:
Should speed up the cppyy jobs


vgvassilev (approved):
I hate these cuts and pastes...

mcbarton (changes requested):
To reduce the repetitive pytest options (and future PRs needing to change these options needed to change the ci for all repos), I think we should be using a pytest.toml is cppyy (see https://docs.pytest.org/en/stable/reference/customize.html) and we pass `-n auto ` in that toml file to stop us having to manually set the number of worker processes

PR #796, opened 2026-02-03
[cmake] add dynamic check for libstdc++fs

aaronj0:
Based on https://github.com/root-project/root/commit/957ac3356492765c14098525840bcd56f120da67 Looking at the compiler version is not enough to know if we really need to link against stdc++fs:

"""
Whether linking with libstdc++fs is needed depends on the version of libstdc++, not on the compiler. For example, it is possible to use an older library with a newer compiler, or with another compiler that follows a different versioning scheme (for example Clang). """

This adds a similar check in CppInterOp


vgvassilev (approved):
Yes, that's the right check! LGTM!

PR #797, opened 2026-02-04
[interop] Avoid usage of `<filesystem>`

hahnjo:
It requires linking `libstdc++fs` for older versions of the standard library, and the same function is provided by LLVM.


aaronj0 (approved):
LGTM, thank you! I'd like to test this patch with the xeus-cpp openmp kernel to ensure that using `llvm::sys::fs` does not regress. @mcbarton can you try this out?

vgvassilev (approved):
Thank you, @hahnjo! LGTM!

PR #799, opened 2026-02-04
enable dispatch load/reload in single binary, add error message with dlerror

conrade-ctc:
# Description

Problem: Load/UnloadDispatchAPI can't but be used multiple time in an application.

Solution: Enable Load/UnloadDispatchAPI by moving data out and allowing controlled reset of static data via UnloadDispatchAPI.

Also: add `dlerror()` when load fails to get diagnostics on reason for failure (e.g., file not found vs symbols not defined)

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/include/CppInterOp/Dispatch.h
+++ b/include/CppInterOp/Dispatch.h
@@ -25,6 +25,7 @@
 
 #include <cstdlib>
 #include <iostream>
+#include <memory>
 #include <mutex>
 
 #ifdef _WIN32
@@ -191,14 +192,18 @@ extern "C" CPPINTEROP_API CppFnPtrTy CppGetProcAddress(const char* procname);
 // TODO: implement overload that takes an existing opened DL handle
 inline void* dlGetProcAddress(const char* name,
                               const char* customLibPath = nullptr) {
-  if (!name)
-    return nullptr;
+  static auto init = std::make_unique<std::once_flag>();
+  static void* (*getProc)(const char*);
 
-  static std::once_flag init;
-  static void* (*getProc)(const char*) = nullptr;
+  // magic reset to keep static init inlined in function
+  if (!name) {
+    init = std::make_unique<std::once_flag>();
+    getProc = nullptr;
+    return nullptr;
+  }
 
   // this is currently not tested in a multiple thread/process setup
-  std::call_once(init, [customLibPath]() {
+  std::call_once(*init, [customLibPath]() {
     const char* path =
         customLibPath ? customLibPath : std::getenv("CPPINTEROP_LIBRARY_PATH");
     if (!path)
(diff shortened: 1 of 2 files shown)


unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -272,7 +272,11 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_DISABLED_DetectResourceDir) {
     GTEST_SKIP() << "Test not run (Clang binary does not exist)";
 
   std::string DetectedPath = Cpp::DetectResourceDir(Clang.str().str().c_str());
-  EXPECT_STREQ(DetectedPath.c_str(), Cpp::GetResourceDir());
+  llvm::SmallString<256> absPath(Cpp::GetResourceDir());
+  llvm::sys::fs::make_absolute(absPath);
```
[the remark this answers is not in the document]
conrade-ctc: moving this into a separate PR

unittests/CMakeLists.txt
```diff
@@ -77,10 +78,7 @@ function(add_cppinterop_unittest name)
   # FIXME: Enable for WASM builds
   if (NOT EMSCRIPTEN AND ARG_ENABLE_DISPATCH AND BUILD_SHARED_LIBS)
     add_cppinterop_unittest(${name}Dispatch ${ARG_UNPARSED_ARGUMENTS} DispatchTest.cpp)
```
mcbarton:
```suggestion
    add_cppinterop_unittest(${name}Dispatch ${ARG_UNPARSED_ARGUMENTS} DispatchTest.cpp DispatchLoadUnloadTest.cpp)
```

I'm pretty sure this suggested change is correct. I don't think we want any dispatch test files should be included at the moment for Windows or Emscripten. @aaronj0 can confirm this suggestion is correct?

vgvassilev:
I wonder what's emscripten's problem...

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ endif()
 add_cppinterop_unittest(CppInterOpTests
   ENABLE_DISPATCH
   CommentReflectionTest.cpp
+  DispatchLoadUnloadTest.cpp
```
mcbarton:
```suggestion
```

mcbarton:
> I wonder what's emscripten's problem...

We don't currently run the dispatch tests we have on Emscripten because it has a broken dlopen (doesn't work the same as native platforms. @anutosh491 has an issue open about this somewhere). The dispatch tests aren't not run on Windows at the moment, since we don't have a shared library build of CppInterOp on Windows.

unittests/CMakeLists.txt
conrade-ctc: Updated with your changes, thanks!

vgvassilev (commented):
Lgtm, however I’d wait for @aaronj0’s review.

aaronj0 (approved):
Thanks for this nice PR! I have just one minor comment: it would be good to append the newly added tests to `unittests/CppInterOp/DispatchTest.cpp` so we can keep all the Dispatch-related tests in a single source file

PR #800, opened 2026-02-04
make path resolved absolute in detection test

conrade-ctc:
# Description

In environments where links and a mix of relative paths are used, finding a fixed point to compare is necessary.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

previous test is still covering, an external env that was failing test before is now passing.

## Checklist


vgvassilev (approved):
Lgtm!

PR #801, opened 2026-02-04
Remove using namespace std; from CppInterOp.cpp

mcbarton:
I was always told to avoid the use of `using namespace std;` , so thought we might not want it in the CppInterOp codebase. That is my only rationale to this PR.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -87,7 +87,7 @@
 #ifndef _WIN32
 #include <unistd.h>
 #endif
-
+#include <utility>
 // Stream redirect.
 #ifdef _WIN32
 #include <io.h>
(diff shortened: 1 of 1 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2008,7 +2007,7 @@ enum EReferenceType { kNotReference, kLValueReference, kRValueReference };
 
 // FIXME: Use that routine throughout CallFunc's port in places such as
 // make_narg_call.
-static inline void indent(ostringstream& buf, int indent_level) {
```
vgvassilev, line 2011: Why the `static` was dropped here?

mcbarton: It was a clang-tidy suggestion https://github.com/compiler-research/CppInterOp/pull/801#discussion_r2765698470

vgvassilev: Ok, good point...

vgvassilev (approved):
LGTM!

PR #802, opened 2026-02-05
Add full member function support in BestOverloadFunctionMatch

### LINKED ISSUE #590: [Feature Request]: Full member function support in `BestOverloadFunctionMatch`

I've found that `BestOverloadFunctionMatch` doesn't select any member function if one is `const`. It also doesn't support overloading based on `&` or `&&`.

```cpp
struct bar
{
  int foo(int i)
  {
    return i;
  }

  int foo(int i) const
  {
    return i + 1;
  }
};
```

If the second `foo` is commented out, I can match the first member fn and call it no problem. But when both are present, neither matches.

This also brings up the important point that `BestOverloadFunctionMatch` doesn't take into account the constness or the value category of the invoking object. So matching `const` member fns or matching `&` or `&&` member fns cannot be done without specifying more information to overload resolution.

I think we should be able to solve this by having the invoking object as the first argument and then resolve on that. Right now, `BestOverloadFunctionMatch` actually expects that it's not present. So, in my example above, the only argument would be one `int`.

If instead we need to provide a `bar const&` and `int` then we can match the overload on constness and value category of the invoking object.

@Vipul-Cariappa for vis.

Priyanshu3820:
# Description

This PR enhances `BestOverloadFunctionMatch` to properly handle const-qualified member function overload resolution by utilizing Clang's `AddMethodCandidate` API instead of `AddOverloadCandidate`.
Fixes #590 

## Type of change

- [✅] Bug fix
- [✅] New feature

## Testing

Tested calling with both const object type and non-const object type

## Checklist

- [✅] I have read the contribution guide recently

### DIFF
--- a/unittests/CppInterOp/InterpreterTest.cpp
+++ b/unittests/CppInterOp/InterpreterTest.cpp
@@ -70,7 +70,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Evaluate) {
   if (TypeParam::isOutOfProcess)
     GTEST_SKIP() << "Test fails for OOP JIT builds";
   //  EXPECT_TRUE(Cpp::Evaluate(I, "") == 0);
-  // EXPECT_TRUE(Cpp::Evaluate(I, "__cplusplus;") == 201402);
+  //EXPECT_TRUE(Cpp::Evaluate(I, "__cplusplus;") == 201402);
   // Due to a deficiency in the clang-repl implementation to get the value we
   // always must omit the ;
   TestFixture::CreateInterpreter();
@@ -156,8 +156,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Process) {
     GTEST_SKIP() << "Test fails for OOP JIT builds";
   if (llvm::sys::RunningOnValgrind())
     GTEST_SKIP() << "XFAIL due to Valgrind report";
-  std::vector<const char*> interpreter_args = {"-include", "new", "-Xclang",
-                                               "-iwithsysroot/include/compat"};
+  std::vector<const char*> interpreter_args = { "-include", "new", "-Xclang", "-iwithsysroot/include/compat" };
   auto* I = TestFixture::CreateInterpreter(interpreter_args);
   EXPECT_TRUE(Cpp::Process("") == 0);
   EXPECT_TRUE(Cpp::Process("int a = 12;") == 0);
@@ -178,11 +177,14 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Process) {
 
 TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_EmscriptenExceptionHandling) {
 #ifndef EMSCRIPTEN
-  GTEST_SKIP() << "This test is intended to check exception handling for "
-                  "Emscripten builds.";
+  GTEST_SKIP() << "This test is intended to check exception handling for Emscripten builds.";
 #endif
-  std::vector<const char*> Args = {"-std=c++20", "-v", "-fwasm-exceptions",
-                                   "-mllvm", "-wasm-enable-sjlj"};
+    std::vector<const char*> Args = {
+    "-std=c++20",
+    "-v",
+    "-fwasm-exceptions",
+    "-mllvm","-wasm-enable-sjlj"
+  };
 
   Cpp::CreateInterpreter(Args, {});
 
@@ -234,7 +236,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_CreateInterpreter) {
 #ifndef CPPINTEROP_USE_CLING
 TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_CreateInterpreterCAPI) {
   const char* argv[] = {"-std=c++17"};
-  auto* CXI = clang_createInterpreter(argv, 1);
+  auto *CXI = clang_createInterpreter(argv, 1);
   auto CLI = clang_Interpreter_getClangInterpreter(CXI);
   EXPECT_TRUE(CLI);
   clang_Interpreter_dispose(CXI);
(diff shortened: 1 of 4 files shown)


Priyanshu3820:
Wasn't sure whether to check the new feature option but did it anyways. Also, a lot of diffs happened after running clang-format

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -70,7 +70,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Evaluate) {
   if (TypeParam::isOutOfProcess)
     GTEST_SKIP() << "Test fails for OOP JIT builds";
   //  EXPECT_TRUE(Cpp::Evaluate(I, "") == 0);
-  //EXPECT_TRUE(Cpp::Evaluate(I, "__cplusplus;") == 201402);
+  // EXPECT_TRUE(Cpp::Evaluate(I, "__cplusplus;") == 201402);
```
mcbarton: Random whitespace change. Revert

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -156,7 +156,8 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Process) {
     GTEST_SKIP() << "Test fails for OOP JIT builds";
   if (llvm::sys::RunningOnValgrind())
     GTEST_SKIP() << "XFAIL due to Valgrind report";
-  std::vector<const char*> interpreter_args = { "-include", "new", "-Xclang", "-iwithsysroot/include/compat" };
+  std::vector<const char*> interpreter_args = {"-include", "new", "-Xclang",
+                                               "-iwithsysroot/include/compat"};
   auto* I = TestFixture::CreateInterpreter(interpreter_args);
```
mcbarton, line 160: Random format change. Revert

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -177,14 +178,11 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Process) {
...
 TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_EmscriptenExceptionHandling) {
 #ifndef EMSCRIPTEN
-  GTEST_SKIP() << "This test is intended to check exception handling for Emscripten builds.";
+  GTEST_SKIP() << "This test is intended to check exception handling for "
+                  "Emscripten builds.";
 #endif
-    std::vector<const char*> Args = {
-    "-std=c++20",
-    "-v",
-    "-fwasm-exceptions",
```
mcbarton, line 185: Random formatting change. Revert.

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -236,7 +234,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_CreateInterpreter) {
 #ifndef CPPINTEROP_USE_CLING
 TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_CreateInterpreterCAPI) {
   const char* argv[] = {"-std=c++17"};
-  auto *CXI = clang_createInterpreter(argv, 1);
+  auto* CXI = clang_createInterpreter(argv, 1);
```
mcbarton: Random change. Revert.

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -247,7 +245,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_CreateInterpreterCAPIFailure) {
   GTEST_SKIP() << "Disabled on Windows. Needs fixing.";
 #endif
   const char* argv[] = {"-fsyntax-only", "-Xclang", "-invalid-plugin"};
-  auto *CXI = clang_createInterpreter(argv, 3);
+  auto* CXI = clang_createInterpreter(argv, 3);
```
mcbarton: Random change. Revert.

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -314,7 +316,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_IncludePaths) {
   Cpp::AddIncludePath("/non/existent/");
   Cpp::GetIncludePaths(includes);
   EXPECT_NE(std::find(includes.begin(), includes.end(), "/non/existent/"),
-             std::end(includes));
+            std::end(includes));
```
mcbarton: Random whitespace change. Revert.

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -358,22 +360,23 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_ExternalInterpreter) {
...
-    auto *ExtInterp = new compat::Interpreter(ClingArgv.size(), &ClingArgv[0]);
+  std::string MainExecutableName = sys::fs::getMainExecutable(nullptr, nullptr);
+  llvm::SmallString<128> P(LLVM_BINARY_DIR);
+  llvm::sys::path::append(P, CLANG_INSTALL_LIBDIR_BASENAME, "clang",
+                          CLANG_VERSION_MAJOR_STRING);
+  std::string ResourceDir = std::string(P.str());
+  std::vector<const char*> ClingArgv = {"-resource-dir", ResourceDir.c_str(),
+                                        "-std=c++14"};
+  ClingArgv.insert(ClingArgv.begin(), MainExecutableName.c_str());
+  auto* ExtInterp = new compat::Interpreter(ClingArgv.size(), &ClingArgv[0]);
```
mcbarton: Random formatting/whitespace change. Revert.

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -358,22 +360,23 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_ExternalInterpreter) {
...
 #endif
 
   EXPECT_NE(ExtInterp, nullptr);
 
 #if !defined(NDEBUG) && GTEST_HAS_DEATH_TEST
 #ifndef _WIN32 // Windows seems to fail to die...
-    EXPECT_DEATH(Cpp::UseExternalInterpreter(ExtInterp), "sInterpreter already in use!");
+  EXPECT_DEATH(Cpp::UseExternalInterpreter(ExtInterp),
+               "sInterpreter already in use!");
```
mcbarton: Random formatting change. Revert.

mcbarton (changes requested):

Priyanshu3820:
@mcbarton, I've fixed the formatting

PR #804, opened 2026-02-09
Reduce Emscripten shared library size back to 38 Mb for llvm 21

mcbarton:
We know the LLVM_ABI annotation being defined as `__attribute__((visibility("default")))` for Emscripten build was having a negative impact on the size of our Emscripten shared library (by making it so a large number of symbols are not eliminated as dead code). This PR patches llvm so that it is blank, and our Emscripten shared library drops back down to its size we had for llvm 20. When this patch is applied locally all Emscripten tests pass for CppInterOp and xeus-cpp, so the ci should pass.

### DIFF
--- a/patches/llvm/emscripten-clang21-reduce-emscripten-shared-library-size.patch
+++ b/patches/llvm/emscripten-clang21-reduce-emscripten-shared-library-size.patch
@@ -1,0 +1,13 @@
+diff --git a/llvm/include/llvm/Support/Compiler.h b/llvm/include/llvm/Support/Compiler.h
+index 297d3e9b0..bf11024cf 100644
+--- a/llvm/include/llvm/Support/Compiler.h
++++ b/llvm/include/llvm/Support/Compiler.h
+@@ -202,7 +202,7 @@
+ #define LLVM_EXPORT_TEMPLATE
+ #define LLVM_ABI_EXPORT LLVM_ABI
+ #elif defined(__MACH__) || defined(__WASM__) || defined(__EMSCRIPTEN__)
+-#define LLVM_ABI __attribute__((visibility("default")))
++#define LLVM_ABI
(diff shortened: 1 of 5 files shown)


patches/llvm/emscripten-clang21-reduce-emscripten-shared-library-size.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/llvm/include/llvm/Support/Compiler.h b/llvm/include/llvm/Support/Compiler.h
+index 297d3e9b0..bf11024cf 100644
+--- a/llvm/include/llvm/Support/Compiler.h
++++ b/llvm/include/llvm/Support/Compiler.h
+@@ -202,7 +202,7 @@
+ #define LLVM_EXPORT_TEMPLATE
+ #define LLVM_ABI_EXPORT LLVM_ABI
+ #elif defined(__MACH__) || defined(__WASM__) || defined(__EMSCRIPTEN__)
```
mcbarton, line 8: A possible alternative to this patch which has the same effect is to drop |`| defined(__EMSCRIPTEN__)`, and Emscripten will be caught by the logic which looks at the case where `LLVM_ABI` is not defined (which is for LLVM annotations to be blank. see https://github.com/llvm/llvm-project/blob/067f1c95a4520e84216ac66699b6c65bb4b47cd2/llvm/include/llvm/Support/Compiler.h#L212) . 

That new patch we might be able to get upstreamed to llvm, if we make a case that the current definition has the effect of effectively stopping Emscriptens dead code elimination from working properly, increasing our Emscripten shared library size.

vgvassilev: Maybe we should ask Saleem in our discord chat first.

PR #809, opened 2026-02-13
Build and test cppyy in ci for multiple build standards

mcbarton:
@vgvassilev I think this is what you was asking for. Assumming the ci passes with this change, cppyy will now be built and tested with multiple c++ standards for every single job. I'll port this over to the other cppyy repos when I finish the reusable actions.

### DIFF
--- a/.github/actions/Build_and_Test_cppyy/action.yml
+++ b/.github/actions/Build_and_Test_cppyy/action.yml
@@ -1,5 +1,8 @@
 name: 'Builds and test cppyy'
 description: 'This action builds and tests cppyy for native platforms'
+inputs:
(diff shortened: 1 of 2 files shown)


vgvassilev:
I think building for 17 is enough. We should not increase the amounts of builds we have.

mcbarton:
> I think building for 17 is enough. We should not increase the amounts of builds we have.

The best solution is to set the default to 17 then. @aaronj0 already has a PR open up doing that here https://github.com/compiler-research/CPyCppyy/pull/174 . My only suggestion is another PR being open setting the standard in cppyy-backends cmake would be useful too.

PR #810, opened 2026-02-13
Add `RequireCompleteType`, using Sema to force type completion

aaronj0:


vgvassilev (approved):
LGTM!

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -383,6 +383,31 @@ bool IsComplete(TCppScope_t scope) {
   return true;
 }
 
+bool RequireCompleteType(TCppScope_t scope) {
```
vgvassilev: Here Type is misnomer -- this needs to be something like CompleteTag or similar.

aaronj0: `CompleteTagDefinition`?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -659,6 +659,42 @@ bool IsComplete(ConstDeclRef DRef) {
   return INTEROP_RETURN(true);
 }
 
+DeclRef GetOrForceDefinition(ConstDeclRef DRef) {
+  INTEROP_TRACE(DRef);
+  if (!DRef)
+    return INTEROP_RETURN(nullptr);
+
+  auto* D = const_cast<clang::Decl*>(unwrap<clang::Decl>(DRef));
```
vgvassilev: That is wrong here. We should take a DeclRef.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -659,6 +659,42 @@ bool IsComplete(ConstDeclRef DRef) {
...
+      compat::SynthesizingCodeRAII RAII(&getInterp());
+      S.isCompleteType(fakeLoc, QT);
+    }
+    return INTEROP_RETURN(TD->getDefinition());
+  }
+
+  // Function: instantiate the body (template) or resolve the out-of-line def.
+  if (auto* FD = dyn_cast<FunctionDecl>(D)) {
+    if (!FD->getDefinition())
+      InstantiateFunctionDefinition(D);
```
vgvassilev, line 682: What if the function is not a template but a simple forward declaration?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -659,6 +659,42 @@ bool IsComplete(ConstDeclRef DRef) {
...
+    return INTEROP_RETURN(FD->getDefinition());
+  }
+
+  // Variable: resolve to its defining declaration.
+  if (auto* VD = dyn_cast<VarDecl>(D))
+    return INTEROP_RETURN(VD->getDefinition());
+
+  // No getDefinition() concept for this decl kind.
+  return INTEROP_RETURN(nullptr);
```
vgvassilev, line 689: Can you ehck that VD->getDefinition() and friend return nullptr if the definition is not available? I have a feeling they used to return self..

aaronj0: The current behaviour:
 
 ```
 FunctionDecl::getDefinition() → if (isDefined(Definition)) return Definition; return nullptr;
 VarDecl::getDefinition(ASTContext&) → returns the defining redecl, else return nullptr;
```

File: lib/CppInterOp/CppInterOp.cpp (hunk 3 above)
Comment (aaronj0) [reply lib/CppInterOp/CppInterOp.cpp:682]: Good point. I guess we should it guard with `FD->isTemplateInstantiation()`?

File: lib/CppInterOp/CppInterOp.cpp (hunk 4 above)
Comment (vgvassilev) [reply lib/CppInterOp/CppInterOp.cpp:689]: ok, fair enough.

File: lib/CppInterOp/CppInterOp.cpp (hunk 3 above)
Comment (vgvassilev) [reply lib/CppInterOp/CppInterOp.cpp:682]: Yes, we will need some level of guard, not sure if that's the right one.

Comment (aaronj0) [reply lib/CppInterOp/CppInterOp.cpp:682]: It has been added.

File: lib/CppInterOp/CppInterOp.cpp [hunk 5]
```diff
@@ -659,6 +659,44 @@ bool IsComplete(ConstDeclRef DRef) {
...
 }
 
+DeclRef GetOrForceDefinition(DeclRef DRef) {
+  INTEROP_TRACE(DRef);
+  if (!DRef)
+    return INTEROP_RETURN(nullptr);
+
+  auto* D = unwrap<clang::Decl>(DRef);
+
+  // Tag (class/struct/union/enum): complete through Sema, which drives ROOT's
```
vgvassilev: That comment is inaccurate and not necessary. We should drop it.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -659,6 +659,44 @@ bool IsComplete(ConstDeclRef DRef) {
...
+      clang::Sema& S = getSema();
+      QualType QT = QualType::getFromOpaquePtr(GetTypeFromScope(DRef).data);
+      SourceLocation fakeLoc = GetValidSLoc(S);
+      compat::SynthesizingCodeRAII RAII(&getInterp());
+      S.isCompleteType(fakeLoc, QT);
+    }
+    return INTEROP_RETURN(TD->getDefinition());
+  }
+
+  // Function: force the body of an instantiation that has a template pattern
```
vgvassilev: Drop the comments they are too obvious and verbose.

PR #811, opened 2026-02-16
Drop llvm 19 support

mcbarton:
We plan to only support 3 llvms at any one time. This PR is not needed right now, but is preparation for when we get llvm 22 support. Setting to draft until CppInterOp is compatible with llvm 22.


.github/workflows/main.yml
```diff
@@ -1,384 +1,352 @@
-name: Native Builds
```
vgvassilev, line 1: Why did we end up rewriting this entire file?

mcbarton: This commit on main https://github.com/compiler-research/CppInterOp/commit/ab4ca85b318139af27a0551f25b3f33d7c266dd9 has caused nothing but issues with diffs (never understood why it got merged in to begin with, since it had nothing to do with the PR it got merged in with https://github.com/compiler-research/CppInterOp/pull/827/commits). When fixing the merge conflict, I chose main.yml on main, and removed the llvm 19 jobs. No idea why it is showing such a big diff.

aaronj0: image

aaronj0: Not sure if you use Windows for development, but that file was CRLF encoded. The diff is not in the characters themselves but the encoding. Please ensure this PR does not re-introduce that.

aaronj0 (dismissed):
Please revert CRLF encoding on main.yml

.github/workflows/main.yml
mcbarton: The ai may be explaining why you chose  to make that change to the yml file, but what has this change had to do with `Use dated ROOT/llvm tag to build Cling 1.3`? it had nothing to do with that. It was a very disruptive commit which causes all other PRs editing the ci to have a merge conflict.

Again, I don't know why the big diff for that file. I used Github to fix the merge conflict. I used main.yml as it was, and just removed the llvm 19 jobs. That might say Githubs default is CRLF maybe?

mcbarton: Now everytime someone contributes the ci from a Windows platform (or uses Github since that is where I made this change) which still uses /r/n we just supposed to say sorry, we cannot have your change. Seems very strange.

aaronj0 (approved):
LGTM

vgvassilev:
For some reason the CI decided to rebuild caches and I've cancelled the builds as they seem irrelevant for what this PR is doing.

PR #773, opened 2026-01-13
[wip] Add support for llvm 22

mcbarton:
# Description

Please include a summary of changes, motivation and context for this PR.

Now that the release/22.x branch has been created, this PR will get CppInterOp ready for the llvm 22 release. It will concentrate on the native case for now, while the issue with wasm exceptions and llvm 21 is resolved.

Fixes # (issue)

## Type of change

Please tick all options which are relevant.

- [ ] Bug fix
- [x] New feature
- [x] Requires documentation updates

## Testing

Please describe the test(s) that you added and ran to verify your changes.

## Checklist

- [x] I have read the contribution guide recently

PR #812, opened 2026-02-16
Add llvm 22 support (native non oop jobs)

mcbarton:
https://github.com/compiler-research/CppInterOp/pull/773 was previously trying to do too much at once, so this PR limits itself to just changes relating to updating CppInterOp to be compatible with llvm 22. This PR will also not update the oop jobs which will be left to someone else to do.


lib/CppInterOp/Compatibility.h
aaronj0, line 24: We need:

```diff
- #include "clang/Driver/Options.h"
+ #if CLANG_VERSION_MAJOR < 22
+ #include "clang/Driver/Options.h"
+ #else
+ #include "clang/Options/Options.h"
+ #endif
```

aaronj0:
> It was wishful thinking on my part that we only had 1 remaining test to fix on the native side. ScopeReflection_GetCompleteName reflection fails with (look here for ci showing the failure https://github.com/compiler-research/CppInterOp/actions/runs/23660708129/job/68929913544?pr=812#step:13:585)
> 
> ```
> 1: [ RUN      ] CppInterOpTest/InProcessJIT.ScopeReflection_GetCompleteName
> 1: /home/runner/work/CppInterOp/CppInterOp/unittests/CppInterOp/ScopeReflectionTest.cpp:404: Failure
> 1: Expected equality of these values:
> 1:   Cpp::GetCompleteName(Decls[10])
> 1:     Which is: "(unnamed enum)"
> 1:   "(unnamed)"
> ```

@vgvassilev I guess something changed in `TypePrinter::printTag` when printing the QualType? The fact that it's "(unnamed enum)" seems fine to me, perhaps we can make that check

```cpp
#if CLANG_VERSION_MAJOR < 22
  EXPECT_EQ(Cpp::GetCompleteName(Decls[10]), "(unnamed)");
#else
  EXPECT_EQ(Cpp::GetCompleteName(Decls[10]), "(unnamed enum)");
#endif
 ```
unless we havea  better way of dealing with this using the printing policy

On a separate note, @mcbarton is this is the only failing test on native?

**Comment by aaronj0** [comment]
> @vgvassilev I guess something changed in `TypePrinter::printTag` when printing the QualType? The fact that it's "(unnamed enum)" seems fine to me, perhaps we can make that check
> 
> ```c++
> #if CLANG_VERSION_MAJOR < 22
>   EXPECT_EQ(Cpp::GetCompleteName(Decls[10]), "(unnamed)");
> #else
>   EXPECT_EQ(Cpp::GetCompleteName(Decls[10]), "(unnamed enum)");
> #endif
> ```
> 
> unless we havea better way of dealing with this using the printing policy
> 
> On a separate note, is this is the only failing test on native?

To clarify, this Decl is the declaration `enum { enum1 };` which is why it is unnamed

File: .github/actions/.DS_Store
Comment (vgvassilev) [inline .github/actions/.DS_Store:1]: These files seem to not belong to this PR.

Comment (mcbarton) [reply .github/actions/.DS_Store:1]: I noticed these getting addrd the other day. I thought I managed to remove them all but clearly not. I will put in a PR which adds this file to the .gitignore.

Comment (mcbarton) [reply .github/actions/.DS_Store:1]: This must be from an old review, since I cannot find this file anymore when I view the PR.

File: lib/CppInterOp/CppInterOp.cpp [hunk 1]
```diff
@@ -614,6 +614,9 @@ static std::string GetCompleteNameImpl(TCppType_t klass, bool qualified) {
     if (qualified) {
       Policy.FullyQualifiedName = true;
       Policy.Suppress_Elab = true;
+#if CLANG_VERSION_MAJOR > 21
+      Policy.SuppressTagKeywordInAnonNames = true;
```
devajithvs: I’m not familiar with the codebase, but this is in the qualified branch, which isn’t used when GetCompleteName is called. I’m not sure if the flag need to be added in other places, I’ll leave that for others to comment on.

EDIT: Commented in the wrong place initially.

.github/workflows/main.yml
```diff
@@ -104,6 +112,13 @@ jobs:
             llvm_targets_to_build: "host;NVPTX"
             python-version: '3.14'
           # Ubuntu X86 Jobs
+          - name: ubu24-x86-gcc12-clang-repl-22
+            os: ubuntu-24.04
+            compiler: gcc-12
+            clang-runtime: '22'
+            llvm_enable_projects: "clang"
+            llvm_targets_to_build: "host;NVPTX"
+            python-version: '3.14'
```
mcbarton, line 121: The Ubuntu x86 ci job for llvm 22 has a Valgrind issue. I do not have access to a Ubuntu x86, or enough knowledge of Valgrind to debug. Leaving having the tests run under Valgrind on Ubuntu x86 to another PR. 

This change was made just get the ci to pass, so I can get on with some of the other llvm 22 upgrades (Emscripten, self hosted runner, asan job once PR is in, maybe oop)

mcbarton:
@vgvassilev @aaronj0 Although this isn't llvm 22 support for every update (oop, emscripten, cuda, etc). This PR will put us one step closer. The ci for this PR should go green, so I would prefer it gets reviewed asap, so we don't regress again from this point.

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -178,6 +178,9 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Process) {
 }
 
 TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_DeclareSilent) {
+#if CLANG_VERSION_MAJOR == 22
```
vgvassilev:
```suggestion
#if CLANG_VERSION_MAJOR > 21
```

mcbarton: Done.

vgvassilev (approved):
Lgtm!

PR #814, opened 2026-02-17
Fix missing xlocale.h errors in Emscripten tests

mcbarton:
If you look here https://github.com/compiler-research/CppInterOp/actions/runs/22098680580/job/63862337816#step:16:683 and https://github.com/compiler-research/CppInterOp/actions/runs/22098680580/job/63862337816#step:16:748 from CppInterOps ci run on main, you'll see errors about not being able to find the  xlocale.h header. This PR will fix those errors.

It fixes this error in the same way it was done in the PR which update CppInterOp to be compatible with emsdk 4.0.9 (this PR https://github.com/compiler-research/CppInterOp/pull/753/changes#diff-f2a694225b366b51ca1eac35ce0794a0142da89d18fcd20a7f2e7cd2317be848).

### DIFF
--- a/unittests/CppInterOp/ScopeReflectionTest.cpp
+++ b/unittests/CppInterOp/ScopeReflectionTest.cpp
@@ -214,7 +214,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, ScopeReflection_IsBuiltin) {
   //  "int", "unsigned int", "long", "unsigned long", "long long", "unsigned long long",
   //  "float", "double", "long double", "void"}
 
-  std::vector<const char*> interpreter_args = { "-include", "new" };
+  std::vector<const char*> interpreter_args = { "-include", "new", "-Xclang", "-iwithsysroot/include/compat" };
 
   TestFixture::CreateInterpreter(interpreter_args);
   ASTContext &C = Interp->getCI()->getASTContext();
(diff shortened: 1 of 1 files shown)


anutosh491 (commented):
This is fine for now but if I think this is not a pick and choose thingy. Cause if you compile a trivial file you should see this 
```
emcc -v b.cpp -o b.html

// prints
/Users/.../clang -target wasm32-unknown-emscripten \
  -fignore-exceptions \
  -mllvm -combiner-global-alias-analysis=false \
  -mllvm -disable-lsr \
  --sysroot=/Users/.../emscripten/cache/sysroot \
  -DEMSCRIPTEN \
  -Xclang -iwithsysroot/include/fakesdl \
  -Xclang -iwithsysroot/include/compat \
  -v -c b.cpp -o ...
```

So we are lucky its just two places that end up needing this but it's more of a default. So this should be a part while creating the Interpreter itself correct (offcourse under the ifdef emscripten condition) !?

unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -214,7 +214,7 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, ScopeReflection_IsBuiltin) {
   //  "int", "unsigned int", "long", "unsigned long", "long long", "unsigned long long",
   //  "float", "double", "long double", "void"}
 
-  std::vector<const char*> interpreter_args = { "-include", "new" };
+  std::vector<const char*> interpreter_args = { "-include", "new", "-Xclang", "-iwithsysroot/include/compat" };
```
vgvassilev, line 217: Isn't that platform-specific?

mcbarton: Its only needed for Emscripten builds. I could make it a default flag for Emscripten like Anutosh suggest, but its only needed in specific cases (at least that is my understanding since its only needed for certain tests). The alternative to making it a default flag, is to just add it to interpreter aguments for these specific tests, if building with Emscripten.

I'll let you and Anutosh decide.

PR #815, opened 2026-02-17
Remove ci matrix options which are off

mcbarton:
Off is the default for a matrix option, so these options being set to off is redundant I believe.


.github/workflows/main.yml
```diff
@@ -319,3 +299,4 @@ jobs:
 
 
 
+
```
vgvassilev: Can we remove that line?

vgvassilev (approved):
LGTM!

PR #816, opened 2026-02-17
Remove no soname from cmake

mcbarton:
```
  set_target_properties(clangCppInterOp
    PROPERTIES NO_SONAME 1
  )
```

was added to be warning free when doing an Emscripten build of CppInterOp for emsdk 3.1.73 . Now that we are using emsdk 4.0.9, this can be removed from CppInterOps cmake, assuming we don't care about introducing a warning for those still building with emsdk 3.1.73 .


mcbarton:
This PR is ready for review

aaronj0 (approved):
LGTM!

PR #818, opened 2026-02-17
Create CPPINTEROP_COMMON_WASM_TEST_COMPILE_FLAGS and CPPINTEROP_COMMON_WASM_TEST_LINK_FLAGS cmake variables to avoid repetition when creating tests

mcbarton:
We have multiple test suites, which have common compile and link options. Therefore this PR creates common compile and link flags to avoid repetition of these options in future when creating new test suites, and the need for changes in multiple places when changing these defaults in the future.


vgvassilev (approved):
LGTM!

PR #787, opened 2026-01-27
ROOT-Cling specific changes required for Cppyy in ROOT

Vipul-Cariappa:
# Description

## Fixes # (issue)

Everything ROOT specific. Eliminated most, if not all, of the crashes in CppInterOp+Cppyy while using it in ROOT.

## Type of change

Please tick all options which are relevant.

- [x] Bug fix
- [ ] New feature
- [ ] Requires documentation updates

## Testing

Not sure how I would test it here.

## Checklist

- [x] I have read the contribution guide recently

PR #819, opened 2026-02-18
fix SynthesizingCodeRAII for clang-repl

Vipul-Cariappa:
# Description

This was a problem with the state of the IntermentalParser in clang.
If the `TranslationUnit` has some problems, then it is not passed to `Execute`, and I believe it is not pushed into `PTUs`.
This is the minimum change that was required to get the tests to pass, in CppInterOp as well as cppyy.
I don't know if we will have problems later when using PCHs and modules.

Maybe we should consider upstreaming this?

## Fixes # (issue)

Regression introduced by #787, that cases `test_leakcheck.py` cppyy test to fail after running for an hour.

- [x] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -308,6 +308,13 @@ static void InstantiateFunctionDefinition(Decl* D) {
     getSema().InstantiateFunctionDefinition(SourceLocation(), FD,
                                             /*Recursive=*/true,
                                             /*DefinitionRequired=*/true);
+    // FIXME: this can go into a RAII object
+    clang::DiagnosticsEngine& Diags = getSema().getDiagnostics();
+    if (!FD->isDefined() && Diags.hasErrorOccurred()) {
+      // instantiation failed, need to reset DiagnosticsEngine
+      Diags.Reset(/*soft=*/true);
+      Diags.getClient()->clear();
+    }
   }
 }
 
--- a/lib/CppInterOp/Compatibility.h
+++ b/lib/CppInterOp/Compatibility.h
@@ -429,10 +429,15 @@ private:
 public:
   SynthesizingCodeRAII(Interpreter* i) : m_Interpreter(i) {}
   ~SynthesizingCodeRAII() {
-    auto GeneratedPTU = m_Interpreter->Parse("");
-    if (!GeneratedPTU)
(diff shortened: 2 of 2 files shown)


Vipul-Cariappa:
> If there is a syntax error, the AST is broken and what code we are expected to jit?

Broken AST is not sent to JIT. So, in theory, the JIT is not in a broken state. Only good ASTs (TranslationUnit) are JITed.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -308,6 +308,12 @@ static void InstantiateFunctionDefinition(Decl* D) {
     getSema().InstantiateFunctionDefinition(SourceLocation(), FD,
                                             /*Recursive=*/true,
                                             /*DefinitionRequired=*/true);
+    clang::DiagnosticsEngine& Diags = getSema().getDiagnostics();
```
vgvassilev, line 312: We need a fixme to move this in a raii.

Vipul-Cariappa: This is the only code path that requires this. Why create a RAII object for a one-off case?

Also, from the last meeting, what I understood is: The `SynthesizingCodeRAII` class should tackle failures at the JIT level. And to create a new RAII object for this (errors at the parser level). If the "diagnostics reset" is required in other code paths at that point, we should add a new RAII object.

vgvassilev: I think such a raii already exists in Sema.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3672,6 +3678,7 @@ bool InsertOrReplaceJitSymbol(compat::Interpreter& I,
   auto Symbol = compat::getSymbolAddress(I, linker_mangled_name);
   llvm::orc::LLJIT& Jit = *compat::getExecutionEngine(I);
   llvm::orc::ExecutionSession& ES = Jit.getExecutionSession();
+  // FIXME: for getMainJITDylib
```
Vipul-Cariappa: This should not be here. Did not plan to commit this. Note to self: remove this.

vgvassilev (approved):
Lgtm! Consider squash in merge.

PR #824, opened 2026-02-25
Fix read the docs developer documentation

mcbarton:
If you visit the developer documentation link in the readme you'll notice it is currently broken. This is because the read the docs config is currently set to use llvm 16 and for some reason llvm 18 was being installed despite trying to build with llvm 16 (given we haven't supported llvm 16 for a while means the read the docs developer documentation has been broken for a long time). This PR should fix everything by updating the docs build to reference an llvm 20 install.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,8 +9,8 @@ build:
   tools:
     python: "3.11"
   apt_packages:
-    - clang-18
+    - clang-20
     - cmake
-    - libclang-18-dev
-    - llvm-18-dev
-    - llvm-18-tools
+    - libclang-20-dev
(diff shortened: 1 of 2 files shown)


.readthedocs.yaml
```diff
@@ -9,8 +9,8 @@ build:
   tools:
     python: "3.11"
   apt_packages:
-    - clang-18
```
mcbarton, line 12: I initially tried clang/llvm 21, but that wasn't available via apt on Ubuntu 24.04 . That is why I choose clang/llvm 20.

aaronj0 (approved):
LGTM, Thanks for catching this

PR #825, opened 2026-02-25
Fix llvm, lld and clang max supported versions in cmake

mcbarton:
The max llvm, lld and clang versions in the cmakelists.txt allow you to try and build with llvm 22, despite CppInterOp not currently supporting it yet. This change accidentally got included in my PR which dropped llvm 18 support. This PR reverts them to the correct values until we actually have llvm 22 support.


vgvassilev (approved):
LGTM!

PR #826, opened 2026-02-27
fix error prone failure path in `GetTypeFromScope`

aaronj0:
The earlier path forced a cast to TypeDecl, which will assert if that's not a valid cast, eg. if you have a NamespaceDecl.

```
llvm/include/llvm/Support/Casting.h:578: decltype(auto) llvm::cast(From*) [with To = clang::TypeDecl; From = clang::Decl]: Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
```

Allows for more flexible calls from cppyy, being less crash-prone. The old test has been extended.


vgvassilev (approved):
LGTM!

Issue #808, opened 2026-02-13
[Bug][ci] : Cling 1.3 ci difference between root and CppInterOps ci


mcbarton:
### Description of bug

In the PR which dropped llvm 18 support (https://github.com/compiler-research/CppInterOp/pull/785) several tests had to be skipped on Windows when updating to cling 1.3 (see comment herehttps://github.com/compiler-research/CppInterOp/pull/785#issuecomment-3840678541). These tests do not fail on Roots ci, so some difference between the ci in Root and CppInterOp is causing them to fail and needs investigating.

### What operating system was you using when the bug occured?

Windows

PR #827, opened 2026-02-27
[ci] Use dated ROOT/llvm tag to build Cling 1.3

aaronj0:
ROOT's development workflow between cling and the LLVM fork are slightly different. We cannot rely on a version of cling that was released at a fixed point in time to reliably work with llvm's cling-llvm20 tag which is a moving target as development will continue after the 1.3 release (1.2 had the same issue, see https://github.com/root-project/cling/issues/556). Based on my understanding, we should probably use the dated tag `cling-llvm20-20260119-01`. I will bump this based on the cppyy migration status (which is the *main* usecase of having cppyy-cling jobs on our CI in the first place)


aaronj0:
@mcbarton I believe these changes won't trigger a rebuild unless the cling cache is deleted.. Can you take a look? Thanks!

aaronj0:
> @aaronj0 I won't have time to review this, as I go on annual leave today. If you want the cache to follow your dated root-llvm-tag, then you'll need to set CLING_HASH in the ci accordingly. You'll also need to update the documentation to use these dated commits. Hope this helps fix the ci. You'll probably want to recheck the tests in this issue as part of this work https://github.com/compiler-research/CppInterOp/issues/808

No worries, thanks alot!

vgvassilev (approved):
LGTM, if the first and last commits are merged.

Issue #744, opened 2025-11-04
[Feature Request]: Add an interface to determine if c or c++ code


mcbarton:
### Description of new feature

In xeus-cpp it was suggested CppInterOp could have an interface to provide what the language is for the code being processed (see https://github.com/compiler-research/xeus-cpp/issues/394#issuecomment-3438379470). This would should allow the c kernels to be testable.

PR #828, opened 2026-02-27
Adding API functions for detecting interpreter language and standard

keremsahn:
Two API functions;  InterpreterLanguage GetLanguage(),  InterpreterLanguageStandard GetLanguageStandard() and enums for interpreter language, standard added.

Fixes https://github.com/compiler-research/CppInterOp/issues/744


lib/CppInterOp/CXCppInterOp.cpp
```diff
@@ -397,6 +397,97 @@ void clang_Interpreter_unloadLibrary(CXInterpreter I, const char* lib_stem) {
   interp->getDynamicLibraryManager()->unloadLibrary(lib_stem);
 }
 
+CXInterpreterLanguage clang_Interpreter_getLanguage(CXInterpreter I) {
```
Vipul-Cariappa, line 401: The C API implementation should just forward the call to the C++ implementation. We don't want to duplicate code.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3494,6 +3487,91 @@ TInterp_t GetInterpreter() {
   return sInterpreters->back().Interpreter;
 }
 
+InterpreterLanguage GetLanguage() {
```
```suggestion
InterpreterLanguage GetLanguage(TInterp_t Interp = nullptr) {
```

I would like it to take an optional argument, `Interp`. If `Interp == nullptr` then get the top of the stack with `getInterp`, otherwise do something similar to the `getInterpreter` definition in `CXCppInterOp.cpp`.

The same, applies to `GetLanguageStandard`.

Vipul-Cariappa (changes requested):
Nice work.
The tests look complete enough to me.
Please take a look at the changes I suggested, and the GitHub Bot (clang-tidy) suggestions.

If possible, could you also look at https://github.com/compiler-research/xeus-cpp/issues/441. Should be an easy fix in `Cpp::CreateInterpreter`.

lib/CppInterOp/CppInterOp.cpp
keremsahn: Done

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3496,6 +3488,103 @@ TInterp_t GetInterpreter() {
...
 
+InterpreterLanguage GetLanguage(TInterp_t I /*=nullptr*/) {
+  compat::Interpreter* interp;
+  if (I) {
+    interp = static_cast<compat::Interpreter*>(I);
+  } else {
+    interp = &getInterp();
+  }
+  const auto& LO = interp->getCI()->getLangOpts();
+  switch (LO.LangStd) {
```
vgvassilev: We should use the x-macro to write the implementation: https://github.com/llvm/llvm-project/blob/c5e5c9735a33909e2268f99e84654e46bd98b109/clang/include/clang/Basic/LangStandard.h#L71-L76

include/CppInterOp/CppInterOp.h
```diff
@@ -102,6 +102,16 @@ enum QualKind : unsigned char {
...
 };
 
+/// Enum modelling programming languages.
+enum class InterpreterLanguage : unsigned char { C, CPlusPlus, Other };
+
+/// Enum modelling language standards.
+enum class InterpreterLanguageStandard {
+#define LANGSTANDARD(id, name, lang, desc, ...) id,
+#define LANGSTANDARD_ALIAS(id, alias)
+#include "clang/Basic/LangStandards.def"
```
vgvassilev: Our public headers can’t depend on llvm headers.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3496,6 +3488,50 @@ TInterp_t GetInterpreter() {
...
+    #define LANGSTANDARD(id, name, lang, desc, ...) \
+      case clang::LangStandard::lang_##id: \
+        return InterpreterLanguageStandard::id;
+    #define LANGSTANDARD_ALIAS(id, alias)
+    #include "clang/Basic/LangStandards.def"
+  default:
+    return InterpreterLanguageStandard::Other;
```
vgvassilev: We should assert here too.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3496,6 +3488,50 @@ TInterp_t GetInterpreter() {
...
+      return InterpreterLanguage::CPlusPlus;
+    case clang::Language::OpenCL:
+      return InterpreterLanguage::OpenCL;
+    case clang::Language::OpenCLCXX:
+      return InterpreterLanguage::OpenCLCXX;
+    case clang::Language::HLSL:
+      return InterpreterLanguage::HLSL;
+    default:
+      return InterpreterLanguage::Other;
+  }
```
Vipul-Cariappa: If the enum `clang::Language` and `InterpreterLanguage` is a simple bijection. Then can't you just
```c++
return (InterpreterLanguage)standard.getLanguage();
```
Maybe more C++ friendly `static_cast` instead of C style cast?

The same goes for `LanguageStandard`.

Vipul-Cariappa: Usually, `if`s are slower than `switch-case`. That was the reason I initially asked if `switch-case` can be used. Given the current implementation, a simple cast without any branches is faster and preferred. When writing code, try to think about such designs. It should be more readable, and less resource hungry (time and memory).

vgvassilev: What happens if a future llvm decides to reorder the values?

keremsahn: Well, I thought that and decided to a casting comparison. I cast the language according to clang::Language::HLSL which is the last value in enum right now. And if the value is bigger than that it asserts. Here it is: 
```c++
  const auto& LO = interp->getCI()->getLangOpts();
  auto standard = clang::LangStandard::getLangStandardForKind(LO.LangStd);
  auto lang = standard.getLanguage();
  assert(static_cast<unsigned char>(lang) <=
        static_cast<unsigned char>(clang::Language::HLSL) &&
        "Unhandled language");
  return static_cast<InterpreterLanguage>(lang);
```

This can handle new added languages.

  But  as you said the enum values of languages may differ in llvm versions so I do not think this is reliable

Vipul-Cariappa: @vgvassilev, we already do this with the enums for overloads. It is just a forwarding to the underlying Clang enums, but redefined in CppInterOp header because we don't want to expose LLVM/Clang to the user. I believe this is fine. We should be careful during the LLVM upgrades.

@keremsahn the assertion you suggest is a good start, but it will not work for the case where things are reordered. But when things are reordered, I believe the tests will catch it.

---

Also, use the following syntax to highlight multiline code in GitHub conversations.
````
```<language>
multi
line
code
```‎
````
Example:
````
```c++
void f() {
  std::cout << "Hello, World!" << std::endl;
}
```
````
Will be rendered as:
```c++
void f() {
  std::cout << "Hello, World!" << std::endl;
}
```

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -337,6 +337,71 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_CodeCompletion) {
...
+  EXPECT_EQ(clang_Interpreter_getLanguageStandard(CXI),
+            CXInterpreterLanguageStandard_cxx14);
+  clang_Interpreter_dispose(CXI);
+}
+#endif
+
+TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_GetLanguageC) {
+  TestFixture::CreateInterpreter({"-xc", "-std=c99"});
+  EXPECT_EQ(Cpp::GetLanguage(), Cpp::InterpreterLanguage::C);
+  EXPECT_EQ(Cpp::GetLanguageStandard(), Cpp::InterpreterLanguageStandard::c99);
```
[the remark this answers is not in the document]
keremsahn: Gave them nullptr as argument. I were not aware that functions do not get default argument in dispatch mode

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3496,6 +3488,38 @@ TInterp_t GetInterpreter() {
   return sInterpreters->back().Interpreter;
 }
 
+InterpreterLanguage GetLanguage(TInterp_t I /*=nullptr*/) {
+  compat::Interpreter* interp;
+  if (I) {
+    interp = static_cast<compat::Interpreter*>(I);
+  } else {
+    interp = &getInterp();
+  }
```
Vipul-Cariappa: Repeated code in both functions. Please update the `getInterp` function directly, to take the argument `Cpp::TInterp_t`.

include/clang-c/CXCppInterOp.h
```diff
@@ -87,6 +87,74 @@ typedef enum {
...
+  CXInterpreterLanguage_LLVM_IR = 3,
+  CXInterpreterLanguage_C = 4,
+  CXInterpreterLanguage_CPlusPlus = 5,
+  CXInterpreterLanguage_ObjC = 6,
+  CXInterpreterLanguage_ObjCPlusPlus = 7,
+  CXInterpreterLanguage_OpenCL = 8,
+  CXInterpreterLanguage_OpenCLCXX = 9,
+  CXInterpreterLanguage_CUDA = 10,
+  CXInterpreterLanguage_HIP = 11,
+  CXInterpreterLanguage_HLSL = 12
```
Vipul-Cariappa:
```suggestion
  CXInterpreterLanguage_Unknown,
  CXInterpreterLanguage_Asm,
  CXInterpreterLanguage_CIR,
  CXInterpreterLanguage_LLVM_IR,
  CXInterpreterLanguage_C,
  CXInterpreterLanguage_CPlusPlus,
  CXInterpreterLanguage_ObjC,
  CXInterpreterLanguage_ObjCPlusPlus,
  CXInterpreterLanguage_OpenCL,
  CXInterpreterLanguage_OpenCLCXX,
  CXInterpreterLanguage_CUDA,
  CXInterpreterLanguage_HIP,
  CXInterpreterLanguage_HLSL
```

You have assigned the default values; I guess that is not necessary.

PR #829, opened 2026-03-02
Fix macos cmake build

tryh4rd-26:
Fixed macOS single-config CMake test invocation by avoiding an empty config value in the check-cppinterop path.
On macOS with Unix Makefiles/Ninja, $<CONFIG> can resolve to empty, which breaks CTest when passed via --build-config; the target now uses the non-config form in that case.
Behavior for multi-config generators remains unchanged.
Validation: cmake --build build --target check-cppinterop --parallel 4 → 100% tests passed (3/3), 0 failed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,13 @@ endif()
 if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
   project(CppInterOp)
 
+  if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+    message(WARNING "No build type selected. Defaulted CMAKE_BUILD_TYPE=Release.\n"
+      "Valid options are: None, Debug, Release, RelWithDebInfo, MinSizeRel.")
+    set(CMAKE_BUILD_TYPE Release CACHE STRING
+      "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel" FORCE)
+  endif()
+
   # LLVM/Clang/Cling default paths
   if (DEFINED LLVM_DIR)
     if (NOT DEFINED Clang_DIR)


tryh4rd-26:
> What was the used cmake configuration command? Can you paste that along with the error from ctest

@aaronj0 



Used CMake configure command:

```bash
cmake -DBUILD_SHARED_LIBS=ON -DCPPINTEROP_USE_CLING=ON -DCPPINTEROP_USE_REPL=Off -DCling_DIR=$LLVM_DIR/build/tools/cling -DLLVM_DIR=$LLVM_DIR/build/lib/cmake/llvm -DClang_DIR=$LLVM_DIR/build/lib/cmake/clang -DCMAKE_INSTALL_PREFIX=$CPPINTEROP_DIR ..
```

CTest/check target error before the fix:

```text
CMake Error: Invalid value used with --build-config
```

Full tail from the failing run:

```text
[100%] Built target CppInterOpUnitTests
CMake Error: Invalid value used with --build-config
gmake[3]: *** [unittests/CMakeFiles/check-cppinterop.dir/build.make:70: unittests/CMakeFiles/check-cppinterop] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:659: unittests/CMakeFiles/check-cppinterop.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:666: unittests/CMakeFiles/check-cppinterop.dir/rule] Error 2
gmake: *** [Makefile:351: check-cppinterop] Error 2
```

unittests/CMakeLists.txt
```diff
@@ -86,7 +86,14 @@ endfunction()
 
 add_subdirectory(CppInterOp)
 
-add_custom_target(check-cppinterop COMMAND ${CMAKE_CTEST_COMMAND} -V --build-config $<CONFIG>
-  DEPENDS CppInterOpUnitTests WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+# macOS single-configuration generators produce an empty $<CONFIG>, so omit
+# --build-config there. Keep existing behavior on other platforms.
```
aaronj0: I don't understand the point of duplicating the command without `$<CONFIG>`, that's a hacky solution.
Can you try patching the CMake with the suggestion above and see if that solves your build?

aaronj0 (changes requested):
What we need is something like https://github.com/llvm/llvm-project/blob/5ff5a1f14761ecc7160125aadb38c59d29f58e5c/llvm/CMakeLists.txt#L81-L87

unittests/CMakeLists.txt
tryh4rd-26: Hi, 
  I’ve updated the PR to follow the LLVM-style approach instead of duplicating the check-cppinterop command

unittests/CMakeLists.txt
```diff
@@ -86,7 +86,15 @@ endfunction()
 
 add_subdirectory(CppInterOp)
 
-add_custom_target(check-cppinterop COMMAND ${CMAKE_CTEST_COMMAND} -V --build-config $<CONFIG>
+if(CMAKE_CONFIGURATION_TYPES)
```
aaronj0: This shouldn't be required

cmake/modules/GoogleTest.cmake
```diff
@@ -31,7 +31,13 @@ if (EMSCRIPTEN)
   endif()
 else()
   set(config_cmd ${CMAKE_COMMAND})
-  set(build_cmd ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR}/unittests/googletest-prefix/src/googletest-build/ --config $<CONFIG>)
+  # Only add --config for multi-configuration generators (Visual Studio, Xcode)
```
aaronj0: Please drop these changes. CMake will propagate the right config since you set the default build type

CMakeLists.txt
```diff
@@ -24,6 +24,13 @@ endif()
 if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
   project(CppInterOp)
 
+  if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+    message(WARNING "No build type selected. Defaulted CMAKE_BUILD_TYPE=Release.\n"
+      "Learn more about the options in our documentation at https://llvm.org/docs/CMake.html#cmake-build-type")
```
aaronj0:
```suggestion
    message(WARNING "No build type selected. Defaulted CMAKE_BUILD_TYPE=Release")
```

aaronj0 (changes requested):
I'd not want to point to LLVM's cmake options for CppInterOp

aaronj0 (approved):

PR #830, opened 2026-03-03
Support Clad as an interpreter external plugin

### LINKED ISSUE #400: [Feature Request]: Enable clad as ExternalProject_Add

We should also provide clad as part of CppInterOp. Perhaps we do not even need to have a shared object but just take the static archive after clad is built and link it to CppInterOp.

aaronj0:
Provides an option to build clad as a plugin to the interpreter. Revived from the EuroAD workshop talk in December.

Currently a draft, this is a WIP that is also verbose. Certain unittests crash in `DiffCollector::TraverseDecl` when `ParseTopLevelDecl` calls `clang::MultiplexConsumer::HandleTranslationUnit`, where consumers introduced by clad are also handled (even though they shouldn't be)


aaronj0:
The test failures are fixed by https://github.com/llvm/llvm-project/pull/182044, and a workaround in Clad is being figured out

CMakeLists.txt
```diff
@@ -526,6 +530,12 @@ if (CPPINTEROP_INCLUDE_DOCS)
   add_subdirectory(docs)
 endif()
 
+if(CPPINTEROP_ENABLE_CLAD)
```
mcbarton, line 626: No real work has been done so far to enable building Clad using Emscripten, so you probably want the configure to error out if CPPINTEROP_ENABLE_CLAD is set to on, and EMSCRIPTEN is true.

aaronj0:
Requires https://github.com/vgvassilev/clad/pull/1944 for LLVM21 jobs to pass

PR #831, opened 2026-03-03
Add containerization

### LINKED ISSUE #746: [feature request] Add the ability to build CppInterOp and its downstream projects in a container

I will begin work soon to add the ability to build CppInterOp and its downstream projects inside a container. This issue is more to detail my initial plan

- [ ] Add a python script which can used to build CppInterOp and its downstream projects for all platforms where they support said platform (linux, macos, windows and emscripten)
- [ ] Add a minimal docker file which simply calls the python script with the correct options, exposing the port needed so that CppInterOp and its downstream projects can be experienced in a browser outside the container
- [ ] Add a minimal apptainer file which simply calls the python script with the correct options, exposing the port needed so that CppInterOp and its downstream projects can be experienced in a browser outside the container (apptainer is to cover hpc systems where docker is not available due to security issues)
- [ ] Updated the docker and apptainer files to expose a users gpu to the container (if the user has one)

Steps for after this work is done

- [ ] Build the docker images locally, and get them uploaded docker hub, so users in future can just download the docker image, and run it locally to experience CppInterOp and its downstream projects

codegallivant:
# Description

I created a Dockerfile setup procedure to allow users to quickly build and use CppInterOp regardless of their environment. The instructions on how to build and use the Dockerfile are in ``docker/README.md``. This PR aims to resolve issue #746 . More changes may be required to create Dockerfiles with different build options.

Includes
Fixes #746 

## Type of change

- [ ] Bug fix
- [X] New feature

## Testing

I ran CppInterOp build tests, which worked - 
```bash
cmake --build . --target check-cppinterop --parallel $(nproc --all)
```

## Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,32 @@
+FROM ubuntu:24.04
+
+SHELL ["/bin/bash", "-c"]
+
+RUN apt update && \
+    apt install -y \
+        git \
+        cmake \
+        build-essential \
+        python3 \
+        python3-dev \
+        python3-pip \
+        python3-virtualenv && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /compiler-research
+
+RUN mkdir -p env-setup
+
+COPY clone.sh env-setup/clone.sh
+COPY setup.sh env-setup/setup.sh
+COPY env.sh env-setup/env.sh
+
+RUN chmod +x env-setup/*.sh
+
+RUN ./env-setup/clone.sh
+RUN ./env-setup/setup.sh
+RUN ./env-setup/env.sh
+
+RUN echo 'source /compiler-research/env-setup/env.sh' >> /etc/bash.bashrc
+
+CMD ["bash"]
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,25 @@
+# Environment Setup using Docker
+Fast automated setup procedure for [CppInterOp](https://github.com/compiler-research/CppInterOp).
+
+## Setup
+
+### Build Dockerfile
+```bash
+docker build -t CppInterOp:v1 . # Build Docker image
+```
+
+## Run & test
+Run the container,
+```bash
+docker run -it CppInterOp:v1 bash
+```
+To test,
+```bash
+cd compiler-research
+source env-setup/env.sh
+cd CppInterOp/build
+# Run tests
+cmake --build . --target check-cppinterop --parallel $(nproc --all)
+```
+You should see this after running the tests - 
+![image](docs/image.png)
--- a/docker/clone.sh
+++ b/docker/clone.sh
@@ -1,0 +1,4 @@
+git clone --depth=1 https://github.com/compiler-research/CppInterOp.git
+git clone --depth=1 https://github.com/compiler-research/cppyy-backend.git
+git clone --depth=1 --branch release/21.x https://github.com/llvm/llvm-project.git
+git clone --depth=1 https://github.com/compiler-research/CPyCppyy.git
--- a/docker/setup.sh
+++ b/docker/setup.sh
@@ -1,0 +1,60 @@
+# Build LLVM
+cd llvm-project
+mkdir build 
+cd build 
+cmake -DLLVM_ENABLE_PROJECTS=clang                                  \
+                -DLLVM_TARGETS_TO_BUILD="host;NVPTX"                \
+                -DCMAKE_BUILD_TYPE=Release                          \
+                -DLLVM_ENABLE_ASSERTIONS=ON                         \
+                -DCLANG_ENABLE_STATIC_ANALYZER=OFF                  \
+                -DCLANG_ENABLE_ARCMT=OFF                            \
+                -DCLANG_ENABLE_FORMAT=OFF                           \
+                -DCLANG_ENABLE_BOOTSTRAP=OFF                        \
+                ../llvm
+cmake --build . --target clang clang-repl --parallel $(nproc --all)
+cmake --build . --target clang clang-repl llvm-jitlink-executor orc_rt-x86_64 --parallel $(nproc --all)
+
+cd ..
+export LLVM_DIR=$PWD
+
+cd ..
+
+export CB_PYTHON_DIR="$PWD/cppyy-backend/python"
+export CPPINTEROP_DIR="$CB_PYTHON_DIR/cppyy_backend"
+export CPLUS_INCLUDE_PATH="${CPLUS_INCLUDE_PATH}:${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_DIR}/build/include:${LLVM_DIR}/build/tools/clang/include"
+
+# Build CppInterOp
+mkdir CppInterOp/build/
+cd CppInterOp/build/
+cmake -DBUILD_SHARED_LIBS=ON -DLLVM_DIR=$LLVM_DIR/build/lib/cmake/llvm -DClang_DIR=$LLVM_DIR/build/lib/cmake/clang -DCMAKE_INSTALL_PREFIX=$CPPINTEROP_DIR -DCMAKE_BUILD_TYPE=Release ..
+cmake --build . --target install --parallel $(nproc --all)
+
+cd ../..
+
+# Build cppyy-backend
+cd cppyy-backend
+mkdir -p python/cppyy_backend/lib build 
+cd build
+cmake -DCppInterOp_DIR=$CPPINTEROP_DIR ..
+cmake --build .
+cp libcppyy-backend.so ../python/cppyy_backend/lib/
+
+# Build CPyCppyy
+virtualenv .venv
+source .venv/bin/activate
+git clone --depth=1 https://github.com/compiler-research/CPyCppyy.git
+mkdir CPyCppyy/build
+cd CPyCppyy/build
+cmake ..
+cmake --build .
+export CPYCPPYY_DIR=$PWD
+
+cd ../..
+
+# Install Cppyy
+export PYTHONPATH=$PYTHONPATH:$CPYCPPYY_DIR:$CB_PYTHON_DIR
+git clone --depth=1 https://github.com/compiler-research/cppyy.git
+cd cppyy
+pip install setuptools
+python3 -m pip install --upgrade . --no-deps --no-build-isolation
+cd ..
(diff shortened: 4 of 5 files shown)


docker/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+                -DLLVM_TARGETS_TO_BUILD="host;NVPTX"                \
+                -DCMAKE_BUILD_TYPE=Release                          \
+                -DLLVM_ENABLE_ASSERTIONS=ON                         \
+                -DCLANG_ENABLE_STATIC_ANALYZER=OFF                  \
+                -DCLANG_ENABLE_ARCMT=OFF                            \
+                -DCLANG_ENABLE_FORMAT=OFF                           \
+                -DCLANG_ENABLE_BOOTSTRAP=OFF                        \
+                ../llvm
+cmake --build . --target clang clang-repl --parallel $(nproc --all)
+cmake --build . --target clang clang-repl llvm-jitlink-executor orc_rt-x86_64 --parallel $(nproc --all)
```
mcbarton, line 15: To biold these targets you need to be on a Ubuntu x86 system with the correct llvm patch applied. You are missing the targets needed to build these targets.

This also don't account for the case where we build on an arm based system, where these targets would be useless.

I also do not understand why you build the targets for out of process, but don't bother having DLLVM_BUILT_WITH_OOP_JIT=ON .

docker/README.md
```diff
@@ -0,0 +1,25 @@
...
+To test,
+```bash
+cd compiler-research
+source env-setup/env.sh
+cd CppInterOp/build
+# Run tests
+cmake --build . --target check-cppinterop --parallel $(nproc --all)
+```
+You should see this after running the tests - 
+![image](docs/image.png)
```
mcbarton, line 25: There is no such image in this PR.

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+COPY clone.sh env-setup/clone.sh
+COPY setup.sh env-setup/setup.sh
+COPY env.sh env-setup/env.sh
+
+RUN chmod +x env-setup/*.sh
+
+RUN ./env-setup/clone.sh
+RUN ./env-setup/setup.sh
+RUN ./env-setup/env.sh
+
```
mcbarton, line 29: I have no understanding why the build process has been broken down into 3 individual scripts which are copied over to the container.

mcbarton (changes requested):
These are my initial comments. I have quite a few more.

Just a comment that this PR is coming accross as purely created by an ai, and I personally believe it is taken an approach which doesn't make sense.

docker/clone.sh
```diff
@@ -0,0 +1,4 @@
+git clone --depth=1 https://github.com/compiler-research/CppInterOp.git
+git clone --depth=1 https://github.com/compiler-research/cppyy-backend.git
+git clone --depth=1 --branch release/21.x https://github.com/llvm/llvm-project.git
```
mcbarton, line 3: The llvm version should be upto the user. Also, this is not covering the case where the user wants to have a cling based build.

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN apt update && \
+    apt install -y \
+        git \
+        cmake \
+        build-essential \
+        python3 \
+        python3-dev \
+        python3-pip \
+        python3-virtualenv && \
```
mcbarton, line 13: Why do we need to install `virtualenv`. Can't we just use venv?

docker/README.md
```diff
@@ -0,0 +1,25 @@
...
+```bash
+docker run -it CppInterOp:v1 bash
+```
+To test,
+```bash
+cd compiler-research
+source env-setup/env.sh
+cd CppInterOp/build
+# Run tests
+cmake --build . --target check-cppinterop --parallel $(nproc --all)
```
mcbarton, line 22: These tests should really be done as part of the process of building the container.

mcbarton (changes requested):

docker/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Build Dockerfile
+```bash
+docker build -t CppInterOp:v1 . # Build Docker image
+```
+
+## Run & test
+Run the container,
+```bash
+docker run -it CppInterOp:v1 bash
```
mcbarton, line 14: Why do we need the `-it option`

docker/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+cmake --build .
+export CPYCPPYY_DIR=$PWD
+
+cd ../..
+
+# Install Cppyy
+export PYTHONPATH=$PYTHONPATH:$CPYCPPYY_DIR:$CB_PYTHON_DIR
+git clone --depth=1 https://github.com/compiler-research/cppyy.git
+cd cppyy
+pip install setuptools
```
mcbarton, line 58: Pretty sure if you have pip, you automatically have setuptools so redundant.

docker/README.md
codegallivant: The image is not there as I did not want to pollute the upstream repository. I had originally been working on this for my personal use, in https://github.com/codegallivant/cppinterop-env-setup/ .  I'll remove the file path from the README in the PR. Thanks for pointing it out.

docker/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+mkdir -p python/cppyy_backend/lib build 
+cd build
+cmake -DCppInterOp_DIR=$CPPINTEROP_DIR ..
+cmake --build .
+cp libcppyy-backend.so ../python/cppyy_backend/lib/
+
+# Build CPyCppyy
+virtualenv .venv
+source .venv/bin/activate
+git clone --depth=1 https://github.com/compiler-research/CPyCppyy.git
```
mcbarton, line 45: So we have a script for cloning the repos, and then the setup script also clones them? Won't that cause an error.

mcbarton (changes requested):
The more I look at this PR, the more I feel the patch suggestion by the ai hasn't been inspected or tested.

docker/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+cmake --build . --target clang clang-repl llvm-jitlink-executor orc_rt-x86_64 --parallel $(nproc --all)
+
+cd ..
+export LLVM_DIR=$PWD
+
+cd ..
+
+export CB_PYTHON_DIR="$PWD/cppyy-backend/python"
+export CPPINTEROP_DIR="$CB_PYTHON_DIR/cppyy_backend"
+export CPLUS_INCLUDE_PATH="${CPLUS_INCLUDE_PATH}:${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_DIR}/build/include:${LLVM_DIR}/build/tools/clang/include"
```
mcbarton, line 24: These are in your script for env.sh. Why duplicate?

docker/setup.sh
codegallivant: Yes, I did not use it with JIT on. I was planning to add build args so that the user can choose whether they want the patch or not. I was going to add those in after receiving confirmation from you, because I was going to use the --build-arg approach, although you had described a python script approach earlier. What approach do you recommend using?

docker/setup.sh
mcbarton, line 45: We clone CPyCppyy inside the `cppyy-backend` build folder. Why?

docker/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+cd CPyCppyy/build
+cmake ..
+cmake --build .
+export CPYCPPYY_DIR=$PWD
+
+cd ../..
+
+# Install Cppyy
+export PYTHONPATH=$PYTHONPATH:$CPYCPPYY_DIR:$CB_PYTHON_DIR
+git clone --depth=1 https://github.com/compiler-research/cppyy.git
```
mcbarton, line 56: This is getting cloned inside the `cppyy-backend` build folder

docker/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+mkdir CPyCppyy/build
+cd CPyCppyy/build
+cmake ..
+cmake --build .
+export CPYCPPYY_DIR=$PWD
+
+cd ../..
+
+# Install Cppyy
+export PYTHONPATH=$PYTHONPATH:$CPYCPPYY_DIR:$CB_PYTHON_DIR
```
mcbarton, line 55: I'm pretty sure we don't need this.

docker/README.md
```diff
@@ -0,0 +1,25 @@
+# Environment Setup using Docker
+Fast automated setup procedure for [CppInterOp](https://github.com/compiler-research/CppInterOp).
```
mcbarton, line 2: Containers don't make things fast. Then help with reproduciblity and isolating the build from your environment.

docker/Dockerfile
codegallivant: I will add them directly to the Dockerfile. I had done that so I could test it quickly. Apologies. Was not sure if you actually needed containerization as there was uncertainty visible in the issue's comments. So I was waiting for confirmation first.

docker/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Setup
+
+### Build Dockerfile
+```bash
+docker build -t CppInterOp:v1 . # Build Docker image
```
mcbarton, line 8: Don't need `# Build Docker image`. This is implied by having `Build Dockerfile` section

mcbarton (changes requested):

docker/README.md
codegallivant: `-it` flag makes it interactive, so you can run commands within the container.

mcbarton:
> @mcbarton I think using a Dockerfile for the lowest level may be better than a Python script, because states will be saved in layers. Please refer to how this example has done it: https://github.com/stereolabs/zed-ros2-wrapper/tree/master/docker . Instead of the Python script building the repository, it should just call the Dockerfile with different arguments. Please let me know if this is alright.
> 
> For example- build.py: calls `docker build --build-arg xyz=ON` And in the Dockerfile: If build-arg xyz exists, run specific command

Docker is just one such container solution. It cannot be used in certain circumstances such as hpc due to security issues. For example an alternative container solution is called Apptainer. Also by having a python script the user can use it to build cppinterop and downstream projects natively, rather than with a container. Its the most versatile starting point. The containers (we plan to support Docker and Apptainer) will effectively call this Python script when get to the final PR

PR #834, opened 2026-03-04
add automated build script

codegallivant:
# Description

Regarding  https://github.com/compiler-research/CppInterOp/pull/831#issuecomment-3992886429 and #746 .
This PR adds an automated build script requiring minimal user interaction. I have structured such that it is easy to maintain and modify.
Further changes will probably be required. Please provide inputs @mcbarton . Thanks.

Example usage:
```bash
python3 build.py -h        
usage: build.py [-h] [--os {Linux,MacOS,Windows}] [--backend {Clang-REPL,Cling}] [--enable-oop-jit]
                [--base-dir BASE_DIR]

Automated Build Script for CppInterOp

options:
  -h, --help            show this help message and exit
  --os {Linux,MacOS,Windows}
                        Target Operating System (default: Linux)
  --backend {Clang-REPL,Cling}
                        Compiler backend to use (default: Clang-REPL)
  --enable-oop-jit      Enable Out-of-Process JIT Execution (Only applies to Clang-REPL on
                        Linux/MacOS)
  --base-dir BASE_DIR   Base working directory for the build process (default: Current working
                        directory)
```

## Type of change

- [ ] Bug fix
- [X] New feature

## Testing

I ran the build script successfully with following conditions- 
1. Ubuntu 24.04 x86 with Clang-REPL backend with OOP JIT.
2. Ubuntu 24.04 x86 Cling backend
 
The build script has not been tested on MacOS and Windows although the command sequences have been written.

## Checklist


build.py
```diff
@@ -0,0 +1,323 @@
...
+        if self.enable_oop_jit:
+            if self.backend != "Clang-REPL":
+                errors.append("Out-of-Process JIT is only supported with the Clang-REPL backend.")
+
+            if self.target_os not in ("Linux", "MacOS"):
+                errors.append("Out-of-Process JIT is only supported on Linux and MacOS.")
+
+        # Windows restrictions
+        if self.target_os == "Windows" and self.backend == "Cling":
+            errors.append("Cling backend is not supported on Windows.")
```
mcbarton: What? I not sure how this restriction was reached.

build.py
```diff
@@ -0,0 +1,323 @@
...
+parser.add_argument(
+    "--base-dir", 
+    type=str, 
+    default=os.getcwd(),
+    help="Base working directory for the build process (default: Current working directory)"
+)
+
+args = parser.parse_args()
+
+class BuildConfig:
```
mcbarton: What was your reasoning for the need for an entire class for building CppInterOp?

build.py
```diff
@@ -0,0 +1,323 @@
...
+    def __init__(self, target_os, backend, base_dir, enable_oop_jit=False):
+        self.target_os = target_os
+        self.backend = backend
+        self.enable_oop_jit = enable_oop_jit
+        self.base_dir = base_dir
+
+    def validate(self):
+        errors = []
```
mcbarton: We collect all the different errors from what options the user chose? Why?

build.py
```diff
@@ -0,0 +1,323 @@
...
+    print("Invalid build configuration:\n")
+    for err in config_errors:
+        print(f" - {err}")
+    sys.exit(1)
+
+###############################
+# Command State
+###############################
+
+class CommandState:
```
mcbarton: We have an entire class for executing a command. That is very strange

build.py
```diff
@@ -0,0 +1,323 @@
...
+        if env is None:
+            env = self.env
+        process = subprocess.run(self.command, shell=True, cwd=cwd, env=env)
+        if process.returncode != 0:
+            print(f"Error: {self.name} failed with return code {process.returncode}")
+            return False
+        return True
+
+
+class CommandStateSequence:
```
mcbarton: A separate class for collecting the commands together. Why?

build.py
```diff
@@ -0,0 +1,323 @@
...
+###############################
+
+parser = argparse.ArgumentParser(description="Automated Build Script for CppInterOp")
+
+parser.add_argument(
+    "--os", 
+    dest="target_os",
+    type=str, 
+    choices=["Linux", "MacOS", "Windows"], 
+    default="Linux",
```
mcbarton: What is the rational with Linux being a default the target os? Surely if there was any default it would be what python detects for the system your running the script on.

build.py
```diff
@@ -0,0 +1,323 @@
...
+    type=str, 
+    choices=["Clang-REPL", "Cling"], 
+    default="Clang-REPL",
+    help="Compiler backend to use (default: Clang-REPL)"
+)
+
+parser.add_argument(
+    "--enable-oop-jit", 
+    action="store_true", 
+    help="Enable Out-of-Process JIT Execution (Only applies to Clang-REPL on Linux/MacOS)"
```
mcbarton: It is more restrictive then that at the moment.

build.py
```diff
@@ -0,0 +1,323 @@
...
+        return "%NUMBER_OF_PROCESSORS%"
+    elif config.target_os == "MacOS":
+        return "$(sysctl -n hw.ncpu)"
+    else:
+        return "$(nproc --all)"
+
+PARALLEL_FLAG = get_parallel_flag()
+
+# Clone Commands
+cmd_clone_cppinterop = CommandState("Clone CppInterOp", "git clone --depth=1 https://github.com/compiler-research/CppInterOp.git")
```
mcbarton: So we are trying to have a Python script which builds the CppInterOp the user has (which may be patched), and we just build CppInterOps main branch. Why?

mcbarton (changes requested):
This Python script just like the previous PR is coming accross as purely the creation of an ai, without any consideration as to whether it makes any sense what the ai suggests. It looks like an llm was given the readme.md file and asked to produce a Python build script. 

I don't agree with the PR description which says ' I have structured such that it is easy to maintain and modify.'. Some bizarre choices have been made in the construction of this script.

build.py
codegallivant: > What is the rational with Linux being a default the target os? Surely if there was any default it would be what python detects for the system your running the script on.

Will fix that.

build.py
codegallivant: @mcbarton Sorry, what did you mean by "entire class for building CppInterOp" ? The class only stores build configuration. It is done so that later on when we run multiple build configurations for testing multiple builds, it is easy to instantiate them.

build.py
codegallivant: > It is more restrictive then that at the moment.

@mcbarton Sorry, i didn't get what you meant here. Please explain.

build.py
mcbarton: This was my mistake. I still don't understand why you need an entire class to store the build configuration.

build.py
mcbarton: It only works for Linux x86 and MacOS arm.

mcbarton (changes requested):
Do not @ me when replying to every comment of my review. People will respond when they have time, which may not be immediately. @ing in reply to every review comment can quickly lead to my email getting cluttered up. It is very irritating and will make less likely to review the PR any further if you do it again.

build.py
codegallivant: > So we are trying to have a Python script which builds the CppInterOp the user has (which may be patched), and we just build CppInterOps main branch. Why?

There is a function called `append_clone_if_missing`. The cppInterOp repo is only cloned into the directory if it is not present there already. Therefore, if the user wants to reuse his own directory, he is still able to do so. The script will auto-detect that the directory is present and won't re-clone.

scripts/builder/build.py
```diff
@@ -0,0 +1,271 @@
...
+parser = argparse.ArgumentParser(description="Automated Build Script for CppInterOp")
+
+parser.add_argument(
+    "--os", 
+    dest="target_os",
+    type=str, 
+    choices=["Linux", "MacOS", "Windows"], 
+    default=utils.detect_os(),
+    help="Target Operating System (default: auto-detect)"
+)
```
mcbarton: Why is this even an option? I don't get how your expecting this script to build for a different os. If your on Linux, your not going to building for Windows.

scripts/builder/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        return "MacOS"
+    elif system == "Windows":
+        return "Windows"
+    else:
+        return "unknown"
+
+
+def get_parallel_flag(target_os):
+    if target_os == "Windows":
+        return "%NUMBER_OF_PROCESSORS%"
```
mcbarton: This doesn't return anything if building on Powershell. It only returns something in the command prompt.

scripts/builder/build.py
```diff
@@ -0,0 +1,271 @@
...
+config_errors = config.validate()
+if len(config_errors) > 0:
+    print("Invalid build configuration:\n")
+    for err in config_errors:
+        print(f" - {err}")
+    sys.exit(1)
+
+variable_flags = {
+    "__VAR_PARALLEL__": utils.get_parallel_flag(config.target_os)
+}
```
mcbarton: Why is the user forced to do a parrallel build? Some machines will have limited amount of ram, where they won't have the resources to have a process for each cpu core.

scripts/builder/build.py
```diff
@@ -0,0 +1,271 @@
...
+print("\n--- Starting Build ---")
+for command_state in build_sequence.command_list:
+    success = command_state.run()
+    if not success:
+        print("Build pipeline aborted due to an error.")
+        break
+
+print("\nYou may copy the following environment variables\n")
+for key, value in build_sequence.current_env.items():
+    if key in ["LLVM_DIR", "CB_PYTHON_DIR", "CPPINTEROP_DIR", "CLING_DIR", "CLING_BUILD_DIR", "CPLUS_INCLUDE_PATH", "SDKROOT"]:
```
mcbarton: Why do they need the CLING_DIR and CLING_BUILD_DIR? Also SDKROOT is only needed on MacOS.

scripts/builder/utils.py
```diff
@@ -0,0 +1,136 @@
...
+def detect_os():
+    system = platform.system()
+    if system == "Linux":
+        return "Linux"
+    elif system == "Darwin":
+        return "MacOS"
+    elif system == "Windows":
+        return "Windows"
+    else:
+        return "unknown"
```
mcbarton: Why does this exist? Your having detect_os call platform.system to set target_os, rather than just using platform.system where needed.

scripts/builder/utils.py
```diff
@@ -0,0 +1,136 @@
...
+    def validate(self):
+        errors = []
+
+        # Unsupported OS
+        if not (self.target_os.lower() in ["linux", "windows", "macos"]):
+            errors.append(f"Unsupported OS:", self.target_os)
+
+        # Base directory must exist
+        if not os.path.isdir(self.base_dir):
+            errors.append(f"Base directory does not exist: {self.base_dir}")
```
mcbarton: Your consider it an error, rather than just creating the directory if it doesn't exist?

scripts/builder/utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # OOP JIT constraints
+        if self.enable_oop_jit:
+            if self.backend != "Clang-REPL":
+                errors.append("Out-of-Process JIT is only supported with the Clang-REPL backend.")
+
+            if self.target_os not in ("Linux", "MacOS"):
+                errors.append("Out-of-Process JIT is only supported on Linux x86 and MacOS arm.")
```
mcbarton: oop is more restricted that just Linux and MacOS. It is restricted in terms of cpu architechture too.

scripts/builder/utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+    return resolve(raw)
+
+
+class BuildConfig:
+    def __init__(self, target_os, backend, base_dir, enable_oop_jit=False):
+        self.target_os = target_os
+        self.backend = backend
+        self.enable_oop_jit = enable_oop_jit
+        self.base_dir = base_dir
```
mcbarton: Given this is just your parsed options from the command line, I still don't understand the need for storing this.

scripts/builder/utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def resolve(obj):
+        if isinstance(obj, str):
+            for key, value  in variables.items():
+                obj = obj.replace(key, value)
+        elif isinstance(obj, dict):
+            return {k: resolve(v) for k, v in obj.items()}
+        return obj
+
+    return resolve(raw)
```
mcbarton: You have put

```
Commands have been moved into a .yaml file. This is safer than directly editing the .py file whenever commands need to be edited
```

I don't understand why this is safer or better than what you had previously (which I admit I didn't like either). Can you explain more why this is better and safer?

scripts/builder/utils.py
```diff
@@ -0,0 +1,136 @@
...
+        if env is None:
+            env = self.env
+        process = subprocess.run(self.command, shell=True, cwd=cwd, env=env)
+        if process.returncode != 0:
+            print(f"Error: {self.name} failed with return code {process.returncode}")
+            return False
+        return True
+    
+    def enable(self):
+        self.disabled = False
```
mcbarton: self.disabled is the default and you don't call this as far as I can see. Why does this exist?

mcbarton (changes requested):
These are my initial comments, but I have more and will add later when I have time.

scripts/builder/utils.py
codegallivant: > self.disabled is the default and you don't call this as far as I can see. Why does this exist?

You are right that enable() is not called anywhere. I added it only to complement the disable() function I am using here:
The disable() function is called when the user wants to skip a build step (for example I skip building LLVM repeatedly if I have already built it)-
https://github.com/compiler-research/CppInterOp/pull/834/changes/BASE..f773a938f1810901976345a6ef8e9aa441c42b4f#diff-18a61aad40f1a06ae700b4dc041ac9ccfa8a42c979195f83f1c7f67b4c7f3b67R251
Should I remove enable() anyways?

scripts/builder/utils.py
codegallivant: > You have put
> 
> ```
> Commands have been moved into a .yaml file. This is safer than directly editing the .py file whenever commands need to be edited
> ```
> 
> I don't understand why this is safer or better than what you had previously (which I admit I didn't like either). Can you explain more why this is better and safer?

I added it so that a contributor does not accidentally pollute the python file with errors when they meant to edit the terminal commands. Since python is not a compiled language, it would cause greater risk because the error may not even be detected from beforehand. So I thought it was better to separate the files to prevent any mishap.

mcbarton: The errors would happen regardless of whether the commands are in a yml file or Python script if they make a mistake.

scripts/builder/build.py
```diff
@@ -0,0 +1,271 @@
...
+
+# Review build sequence
+print(f"--- Build Plan: {config.target_os} | {config.backend} | OOP JIT: {config.enable_oop_jit} | Base directory: {config.base_dir} ---")
+for i, command_state in enumerate(build_sequence.command_list, 1):
+    print(f"Step {i}: {command_state.name}")
+    print(command_state.command)
+    permission = utils.get_user_permission("Keep command in plan? ")
+    if not permission:
+        command_state.disable()
+    print()
```
mcbarton: So you want the user to give permission for every single command it is going to execute? The would be incredibly annoying and mean most people wouldn't bother using the script.

scripts/builder/build.py
```diff
@@ -0,0 +1,271 @@
...
+        print("Build pipeline aborted due to an error.")
+        break
+
+print("\nYou may copy the following environment variables\n")
+for key, value in build_sequence.current_env.items():
+    if key in ["LLVM_DIR", "CB_PYTHON_DIR", "CPPINTEROP_DIR", "CLING_DIR", "CLING_BUILD_DIR", "CPLUS_INCLUDE_PATH", "SDKROOT"]:
+        if config.target_os == "Windows":
+            print(f'$env:{key}="{value}"')
+        else:
+            print(f'export {key}="{value}"')
```
mcbarton: Why is this is printed even if a command fails?

scripts/builder/cmds.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+patch:
+  apply_jit: "cd llvm-project && git apply -v ../CppInterOp/patches/llvm/clang20-1-out-of-process.patch"
+
+clang_repl:
+  cmake:
+    std: 'mkdir -p llvm-project/build && cd llvm-project/build && cmake -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD="host;NVPTX" -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON -DCLANG_ENABLE_STATIC_ANALYZER=OFF -DCLANG_ENABLE_ARCMT=OFF -DCLANG_ENABLE_FORMAT=OFF -DCLANG_ENABLE_BOOTSTRAP=OFF ../llvm'
+    oop: 'mkdir -p llvm-project/build && cd llvm-project/build && cmake -DLLVM_ENABLE_PROJECTS="clang;compiler-rt" -DLLVM_TARGETS_TO_BUILD="host;NVPTX" -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON -DCLANG_ENABLE_STATIC_ANALYZER=OFF -DCLANG_ENABLE_ARCMT=OFF -DCLANG_ENABLE_FORMAT=OFF -DCLANG_ENABLE_BOOTSTRAP=OFF ../llvm'
+    windows: 'mkdir llvm-project\\build 2>nul & cd llvm-project\\build && cmake -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD="host;NVPTX" -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON -DCLANG_ENABLE_STATIC_ANALYZER=OFF -DCLANG_ENABLE_ARCMT=OFF -DCLANG_ENABLE_FORMAT=OFF -DCLANG_ENABLE_BOOTSTRAP=OFF ..\\llvm'
```
mcbarton: `2>nul`. Why? If I understand correctly you just made it do that it suppresses the error message.

mcbarton (changes requested):
I think your making this build script much more complicated than it needs to be with your current approach, and I feel it will be hard to maintain.

scripts/builder/cmds.yaml
codegallivant: > `2>nul`. Why? If I understand correctly you just made it do that it suppresses the error message.

It was to suppress the error from `mkdir llvm-project` if the directory already exists. I'll see if there's a better method

codegallivant: actually this method of creating the directory seems fine to me? is there an issue im not seeing?

scripts/builder/build.py
codegallivant: I will add an argparser option to disable that.
Also, I just wanted to clarify something in case it was not clear-
The script does not execute the commands as soon as it is appended to the command sequence. It first creates a build plan with each <command string,environment variable, working directory> step. It asks the user for permission for each step at the very start of the entire build. After that it requires no manual intervention and can finish all the steps by itself.

scripts/builder/build.py
```diff
@@ -0,0 +1,165 @@
...
+
+system = platform.system()
+if system == "Linux":
+    target_os = "linux"
+elif system == "Darwin":
+    target_os = "macos"
+elif system == "Windows":
+    target_os = "windows"
+else:
+    raise RuntimeError("Unsupported OS")
```
mcbarton: You set previous review comments I make to resolved without actually tackling them (this is one such example https://github.com/compiler-research/CppInterOp/pull/834#discussion_r2905469087). You are still setting an unnecessary variable (target_os). You could just use `platform.system()` everywhere you need to know what plaform you are compiling on.

scripts/builder/build.py
```diff
@@ -0,0 +1,165 @@
...
+
+for key, value in build_sequence.current_env.items():
+    if key in [
+        "LLVM_DIR",
+        "CB_PYTHON_DIR",
+        "CPPINTEROP_DIR",
+        "CLING_DIR",
+        "CLING_BUILD_DIR",
+        "CPLUS_INCLUDE_PATH",
+        "SDKROOT"
```
mcbarton: Still not fixed despite being labelled as resolved https://github.com/compiler-research/CppInterOp/pull/834#discussion_r2905391458

scripts/builder/build.py
```diff
@@ -0,0 +1,165 @@
...
+else:
+    config_variables["sdkroot"] = ""
+
+# Load config.yaml
+
+with open("config.yml") as f:
+    config_text = f.read()
+config_text = re.sub(r"@@([A-Za-z0-9_-]+)@@", lambda match: utils.replace_var(config_variables, match), config_text)
+config_data = yaml.safe_load(config_text)
+
```
mcbarton, line 98: So you open config.yml and replace with `@@example_variable@@` in the config file with the values aas set by the python script. This seems to a very confusing yml file. Although it may give the correct result now, I am no way to easily decipher thing such as if the build script broke in the future
```
env_cplus_include_path_cling:
  type: env
  name: CPLUS_INCLUDE_PATH
  content: "@@cling@@@@DIR_SEP@@tools@@DIR_SEP@@cling@@DIR_SEP@@include@@PATH_SEP@@@@cling@@@@DIR_SEP@@build@@DIR_SEP@@include@@PATH_SEP@@@@llvm-project@@@@DIR_SEP@@llvm@@DIR_SEP@@include@@PATH_SEP@@@@llvm-project@@@@DIR_SEP@@clang@@DIR_SEP@@include@@PATH_SEP@@@@llvm-project@@@@DIR_SEP@@build@@DIR_SEP@@include@@PATH_SEP@@@@llvm-project@@@@DIR_SEP@@build@@DIR_SEP@@tools@@DIR_SEP@@clang@@DIR_SEP@@include"
```

scripts/builder/utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def append_clone(self, dir_path, command):
+        if not os.path.isdir(dir_path):
+            parent_dir = os.path.abspath(dir_path)
+            def new_run():
+                if not os.path.isdir(parent_dir):
+                    os.makedirs(parent_dir)
+                command.run()
+            command.run = new_run
+            command.makedirs = False
```
mcbarton: I think I am missing something `command.run` runs `new_run` which runs `command.run`  which runs `new_run` ... 

Isn't this going to get stuck in an infinite loop?

mcbarton (changes requested):
Since my previous review you have double down on using yml files for the build script, and made the PR even harder to follow. I didn't think it was necessary when you had just one yml file. Everything should be able to be contained purely in the python code. You also do not need the `CommandStateSequence` and `CommandState` classes in my opinion. 

You have also resolved all my previous review comments without actually checking if you have actually tackled them. 

If the above have not been tackled the next time you ping me for review on this PR, I will not review it. You would need to get another maintainer to review it, and get them to agree with you that you are taking the right approach.

scripts/builder/build.py
codegallivant: > Still not fixed despite being labelled as resolved [#834 (comment)](https://github.com/compiler-research/CppInterOp/pull/834#discussion_r2905391458)

Actually, that is fixed. In the for loop that code is nested in, it is iterating through `current_env`. `SDKROOT` is never added to the `current_env` in a non-macOS system.

scripts/builder/build.py
codegallivant: > You set previous review comments I make to resolved without actually tackling them (this is one such example [#834 (comment)](https://github.com/compiler-research/CppInterOp/pull/834#discussion_r2905469087)). You are still setting an unnecessary variable (target_os). You could just use `platform.system()` everywhere you need to know what plaform you are compiling on.

I didn't use `platform.system()` because I didn't want to use the string "Darwin" in place of "MacOS".  I'll shorten it to do something like `system = platform.system().replace("Darwin", "macos").lower()`. Is that fine?

scripts/builder/build.py
mcbarton: This kind of adds to my point about the PR being hard to follow. Its not clear exactly what `build_sequence.current_env.items()` is for each target system. I cannot quickly tell you looking at the current approach what it is, and I should be able to if it was an easily maintainable approach.

scripts/builder/build.py
mcbarton: Why does it need to be `MacOS` instead of `Darwin`? `Darwin` is the core of the most Apple operating systems. There is no need to replace `Darwin` with anything. Just use `Darwin`.

scripts/builder/utils.py
codegallivant: > I think I am missing something `command.run` runs `new_run` which runs `command.run` which runs `new_run` ...
> 
> Isn't this going to get stuck in an infinite loop?

Yes, this wasn't showing up during testing because later on, I would append a different object by calling `append_command`. So one bug concealed the other. Fixed both these bugs now.

scripts/builder/build.py
codegallivant: > So you open config.yml and replace with `@@example_variable@@` in the config file with the values aas set by the python script. This seems to a very confusing yml file. Although it may give the correct result now, I am no way to easily decipher thing such as if the build script broke in the future
> 
> ```
> env_cplus_include_path_cling:
>   type: env
>   name: CPLUS_INCLUDE_PATH
>   content: "@@cling@@@@DIR_SEP@@tools@@DIR_SEP@@cling@@DIR_SEP@@include@@PATH_SEP@@@@cling@@@@DIR_SEP@@build@@DIR_SEP@@include@@PATH_SEP@@@@llvm-project@@@@DIR_SEP@@llvm@@DIR_SEP@@include@@PATH_SEP@@@@llvm-project@@@@DIR_SEP@@clang@@DIR_SEP@@include@@PATH_SEP@@@@llvm-project@@@@DIR_SEP@@build@@DIR_SEP@@include@@PATH_SEP@@@@llvm-project@@@@DIR_SEP@@build@@DIR_SEP@@tools@@DIR_SEP@@clang@@DIR_SEP@@include"
> ```

Just added a feature to read path environment variables in a better way.

PR #842, opened 2026-03-07
Replace string comparison with AST check in IsBuiltin for std::complex

tryh4rd-26:
# Description

Replace the fragile string-based check for `std::complex` in `IsBuiltin()` with a proper Clang AST query.

he old code was checking if a type's name contained the word "complex" as a string, which is pretty hacky cuz any random user class with "complex" in the name would get wrongly treated as a builtin. I swapped it out for a proper AST check that looks at whether the type is actually a `std::complex<T>`specialization by walking the Clang AST. Same approach that's already used in other parts of the codebase, so it fits right in.

Resolves the FIXME at line 411: `// FIXME: Figure out how to avoid the string comparison.`

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

- Ran the existing `ScopeReflection_IsBuiltin` test which covers:
  - Builtin types (`bool`, `char`, `signed char`, `void`)
  - C `_Complex` types (`_Complex float`, `_Complex double`, `_Complex long double`, `_Complex __float128`)
  - `std::complex<T>` template specializations (via `#include <complex>` and iterating over specializations)
- Full test suite: **100% tests passed, 0 tests failed out of 3**

```
$ CppInterOpTests --gtest_filter="*IsBuiltin*"
[ RUN      ] CppInterOpTest/InProcessJIT.ScopeReflection_IsBuiltin
[       OK ] CppInterOpTest/InProcessJIT.ScopeReflection_IsBuiltin (290 ms)
[  PASSED  ] 1 test.
```

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -412,8 +412,16 @@ bool IsBuiltin(TCppType_t type) {
   QualType Ty = QualType::getFromOpaquePtr(type);
   if (Ty->isBuiltinType() || Ty->isAnyComplexType())
     return true;
-  // FIXME: Figure out how to avoid the string comparison.
-  return llvm::StringRef(Ty.getAsString()).contains("complex");
+  // Check for std::complex<T> specializations.
+  if (const auto* RD = Ty->getAsCXXRecordDecl()) {
+    if (const auto* CTSD = dyn_cast<ClassTemplateSpecializationDecl>(RD)) {
+      IdentifierInfo* II = CTSD->getSpecializedTemplate()->getIdentifier();
+      if (II && II->isStr("complex") &&
+          CTSD->getDeclContext()->isStdNamespace())
+        return true;
+    }
+  }
+  return false;
 }
 
 bool IsTemplate(TCppScope_t handle) {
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -408,8 +408,16 @@ bool IsBuiltin(TCppType_t type) {
   QualType Ty = QualType::getFromOpaquePtr(type);
   if (Ty->isBuiltinType() || Ty->isAnyComplexType())
     return true;
-  // FIXME: Figure out how to avoid the string comparison.
-  return llvm::StringRef(Ty.getAsString()).contains("complex");
+  // Check for std::complex<T> specializations.
+  if (const auto* RD = Ty->getAsCXXRecordDecl()) {
+    if (const auto* CTSD = dyn_cast<ClassTemplateSpecializationDecl>(RD)) {
+      IdentifierInfo* II = CTSD->getSpecializedTemplate()->getIdentifier();
+      if (II && II->isStr("complex") &&
```
Vipul-Cariappa, line 419: There is still some string comparison, but I guess this is a better approach.

Vipul-Cariappa (changes requested):
Looks good to me!
Could you add a test case that covers the uncovered line for the test coverage.

Vipul-Cariappa (approved):
LGTM!

I am approving this, but will defer merging for now. As the CI is failing, I believe the CI failures are unrelated to these changes.

lib/CppInterOp/CppInterOp.cpp
vgvassilev, line 419: What about `Type::isAnyComplexType`?

tryh4rd-26: `isAnyComplexType()` is already called on line 409 ... but `std::complex<T>` is a class temp record type and , not a Clang `ComplexType`, so the func should be returning false for it. The AST check below is specifically to handle standard complex<T> typpa specialisations.

PR #843, opened 2026-03-09
Constrain xeus to < 6.0.0 in Emscripten environment

mcbarton:
@Vipul-Cariappa @aaronj0 The Emscripten ci should pass with this change. The restriction on xeus can be removed once https://github.com/compiler-research/xeus-cpp/pull/453 has been approved and merged in (its currently under review by someone from Quant Stack).


aaronj0 (approved):
LGTM, given the jobs pass

PR #844, opened 2026-03-09
Fix FIXME: thread-safe initialization of DispatchInitializer

voyager-jhk:
Fix FIXME regarding thread-safe initialization of DispatchInitializer.

This change replaces the global static instance with a function-local
static to guarantee thread-safe initialization according to C++11.

The initializer is now created via GetDispatchInitializer().

### DIFF
--- a/unittests/CppInterOp/Utils.cpp
+++ b/unittests/CppInterOp/Utils.cpp
@@ -29,14 +29,18 @@ struct DispatchInitializer {
       std::abort();
     }
   }
-  ~DispatchInitializer() { Cpp::UnloadDispatchAPI(); }
+  ~DispatchInitializer() = default;
   DispatchInitializer(const DispatchInitializer&) = delete;
   DispatchInitializer& operator=(const DispatchInitializer&) = delete;
   DispatchInitializer(DispatchInitializer&&) noexcept = default;
   DispatchInitializer& operator=(DispatchInitializer&&) noexcept = default;
 };
-// FIXME: Make this threadsafe by moving it as a function static.
-DispatchInitializer g_dispatch_init;
+// Thread-safe initialization using function-local static
+DispatchInitializer& GetDispatchInitializer() {
+  static DispatchInitializer instance;
+  return instance;
+}
+const DispatchInitializer& g_dispatch_init = GetDispatchInitializer();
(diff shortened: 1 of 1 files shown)


vgvassilev (commented):
Any chance for a test?

unittests/CppInterOp/Utils.cpp
```diff
@@ -35,8 +35,12 @@ struct DispatchInitializer {
...
   DispatchInitializer& operator=(DispatchInitializer&&) noexcept = default;
 };
-// FIXME: Make this threadsafe by moving it as a function static.
-DispatchInitializer g_dispatch_init;
+// Thread-safe initialization using function-local static
+DispatchInitializer& GetDispatchInitializer() {
+  static DispatchInitializer instance;
+  return instance;
+}
+const DispatchInitializer& g_dispatch_init = GetDispatchInitializer();
```
vgvassilev, line 43: Consider `~DispatchInitializer() { Cpp::UnloadDispatchAPI(); }` which also needs to become thread safe in this case.

cc: @aaronj0

voyager-jhk: Agreed. Over-engineered. I'll just drop the FIXME.

vgvassilev: I am not sure I understand your comment.

voyager-jhk: I meant I am abandoning this PR. Since these tests run sequentially, adding thread-safe initialization here is unnecessary and introduces the unsafe destruction you pointed out.

vgvassilev: No, I am suggesting to remove the logic in the destructor as it does not really matter -- should be a no-op

voyager-jhk: Ah, got it. I completely misunderstood earlier. Just dropped the UnloadDispatchAPI call so the destructor is a no-op now.

aaronj0 (approved):
LGMT, thanks!

aaronj0:
@voyager-jhk Can you rebase on main and squash into a single commit? I'd like the Native jobs on the CI to run before taking this in

PR #849, opened 2026-03-10
Add DiagnosticsEngineRAII

codegallivant:
# Description

Fixes ``FIXME: this can go into a RAII object``. Reference:
https://github.com/compiler-research/CppInterOp/blob/852f6d43ea38c3066351c0a4d014ac0d862fa67d/lib/CppInterOp/CppInterOp.cpp#L311

Implementation details-
I added a RAII class (`compat::DiagnosticsEngineRAII`) that automatically checks if it should reset the DiagnosticsEngine before the RAII object goes out of scope. It checks for `diags.hasErrorOccurred()` and a user-provided boolean `reset_condition`. 

## Type of change

- [ ] Bug fix
- [X] New feature

## Testing

```
100% tests passed, 0 tests failed out of 3

Label Time Summary:
DEPENDS    =  12.35 sec*proc (3 tests)

Total Test time (real) =  12.36 sec
[100%] Built target check-cppinterop
```

## Checklist

### DIFF
--- a/lib/CppInterOp/Compatibility.h
+++ b/lib/CppInterOp/Compatibility.h
@@ -430,6 +430,7 @@ public:
   SynthesizingCodeRAII(Interpreter* i) : m_Interpreter(i) {}
   // ~SynthesizingCodeRAII() {} // TODO: implement
 };
+
 } // namespace compat
 
 #endif // CPPINTEROP_USE_REPL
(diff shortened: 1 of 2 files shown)


Vipul-Cariappa (commented):
Can you check if clang's `SFINAETrap` (Ref: https://clang.llvm.org/doxygen/classclang_1_1Sema_1_1SFINAETrap.html#details) does the necessary trick? i.e. can we replace our custom logic with this object instead?

codegallivant:
cling builds are failing. will push a fix soon.

Vipul-Cariappa (commented):
The way this is implemented, it looks more like a function call instead of a RAII object (Look at the way `reset_condition` is used). But I don't know if a simple function call is better or a RAII object is better...

vgvassilev (approved):
Lgtm!

PR #854, opened 2026-03-13
include API after undef windows.h `LoadLibrary`

aaronj0:
We lose `CppImpl::LoadLibrary` otherwise, fixed an error on ROOT's Windows CI when using Dispatch to load the API


vgvassilev (approved):
LGTM!

PR #855, opened 2026-03-13
use a common `compat::Value` to avoid macro guard duplication

aaronj0:
Noticed when condensing duplication in named lookups. This is mainly visible in the C-API, with just one occurrence in the source


mcbarton (approved):
LGTM

vgvassilev (approved):
LGTM!

PR #859, opened 2026-03-13
fix: Move runtime codes to own separate header file

### LINKED ISSUE #697: [Bug]: Move codes needed at runtime to own separate header file
### Description of bug

In https://github.com/compiler-research/CppInterOp/pull/678/files#r2242837167 it was suggested that certain code present in CppInterOps library should be moved to their own header file. See the link in this issue description for one such code. 

The suggested header file name is CppInterOpRuntime.h

### What operating system was you using when the bug occured?

_No response_

### What is the architechture of the cpu on your system?

_No response_

### What did you build CppInterOp against?

_No response_

aniruddhaadak80:
Fixes https://github.com/compiler-research/CppInterOp/issues/697

### Description
This addresses #697, which required extracting the dynamically declared code block inside `CppInterOp.cpp` (the `__internal_CppInterOp` wrapper signatures) out into its own dedicated header file. 

### Changes Made
- Created `include/CppInterOp/CppInterOpRuntime.h` to house the runtime code signatures.
- Replaced the string literal `I->declare` payload in `CppInterOp.cpp` with a standard runtime `#include <CppInterOp/CppInterOpRuntime.h>` preprocessor directive. 

This cleans up the interpreter startup code and makes extending the runtime capabilities much easier.

### DIFF
--- a/include/CppInterOp/CppInterOpRuntime.h
+++ b/include/CppInterOp/CppInterOpRuntime.h
@@ -1,0 +1,22 @@
+//===--- CppInterOpRuntime.h - Runtime codes for language interoperability ---*- C++ -*-===//
+//
+// Part of the compiler-research project, under the Apache License v2.0 with
+// LLVM Exceptions.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CPPINTEROP_CPPINTEROPRUNTIME_H
+#define CPPINTEROP_CPPINTEROPRUNTIME_H
+
+namespace __internal_CppInterOp {
+template <typename Signature>
+struct function;
+
+template <typename Res, typename... ArgTypes>
+struct function<Res(ArgTypes...)> {
+  using result_type = Res;
+};
+}  // namespace __internal_CppInterOp
+
+#endif // CPPINTEROP_CPPINTEROPRUNTIME_H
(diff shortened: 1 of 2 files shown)


include/CppInterOp/CppInterOpRuntime.h
```diff
@@ -0,0 +1,22 @@
+//===--- CppInterOpRuntime.h - Runtime codes for language interoperability ---*- C++ -*-===//
```
Vipul-Cariappa, line 1: Can this file go to `include/CppInterOp/runtime/CppInterOpRuntime.h`.

Vipul-Cariappa (changes requested):
Other than the folder structure change, this is fine, I guess.
But I am now worried about starting the interpreter with the correct include paths.
This is currently running fine, because we are doing a source build with the env var `CPLUS_INCLUDE_PATH` set to point to the necessary things.
But what happens when we do a conda install? I don't think this will work out of the box without modifications to `CreateInterpreter` logic to include this directory.

aniruddhaadak80:
Thanks @Vipul-Cariappa for the review, and @aaronj0 for the pointer - I missed that #761 already covers this. Closing this in favor of it to avoid duplicate work. Note for whoever picks up #761: the conda include-path concern is real - CreateInterpreter will likely need the conda prefix's include dir wired in explicitly rather than relying on CPLUS_INCLUDE_PATH.

PR #864, opened 2026-03-21
Fix self-hosted runner labels for CUDA CI job

aaronj0:
The prepare-dell job needs the 'spotter' label to target the always-on runner, otherwise the wake-onlan packet never reaches the ip. The build matrix entry also needs [self-hosted, cuda, heavy] to route to the correct machine once it's awake. adapted from the working clad CI configuration

### DIFF
--- a/.github/actions/Build_and_Test_CppInterOp/action.yml
+++ b/.github/actions/Build_and_Test_CppInterOp/action.yml
@@ -8,15 +8,21 @@ runs:
       if: ${{ runner.os != 'Windows' && !endsWith(matrix.name, '-emscripten') }}
       shell: bash
       run: |
-        LLVM_DIR="$(pwd)/llvm-project"
-        LLVM_BUILD_DIR="$(pwd)/llvm-project/build"
-        cling_on=$(echo "${{ matrix.cling }}" | tr '[:lower:]' '[:upper:]')
-        if [[ "${cling_on}" == "ON" ]]; then
-          CLING_DIR="$(pwd)/cling"
-          CLING_BUILD_DIR="$(pwd)/cling/build"
-          CPLUS_INCLUDE_PATH="${CLING_DIR}/tools/cling/include:${CLING_BUILD_DIR}/include:${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_BUILD_DIR}/include:${LLVM_BUILD_DIR}/tools/clang/include:$PWD/include"
+        if [[ "${{ runner.environment }}" == "self-hosted" ]]; then
+          LLVM_DIR="/usr/lib/llvm-${{ matrix.clang-runtime }}"
+          LLVM_BUILD_DIR="/usr/lib/llvm-${{ matrix.clang-runtime }}"
+          CPLUS_INCLUDE_PATH="/usr/lib/llvm-${{ matrix.clang-runtime }}/include:$PWD/include"
         else
-          CPLUS_INCLUDE_PATH="${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_BUILD_DIR}/include:${LLVM_BUILD_DIR}/tools/clang/include:$PWD/include"
+          LLVM_DIR="$(pwd)/llvm-project"
+          LLVM_BUILD_DIR="$(pwd)/llvm-project/build"
+          cling_on=$(echo "${{ matrix.cling }}" | tr '[:lower:]' '[:upper:]')
+          if [[ "${cling_on}" == "ON" ]]; then
+            CLING_DIR="$(pwd)/cling"
+            CLING_BUILD_DIR="$(pwd)/cling/build"
+            CPLUS_INCLUDE_PATH="${CLING_DIR}/tools/cling/include:${CLING_BUILD_DIR}/include:${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_BUILD_DIR}/include:${LLVM_BUILD_DIR}/tools/clang/include:$PWD/include"
+          else
+            CPLUS_INCLUDE_PATH="${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_BUILD_DIR}/include:${LLVM_BUILD_DIR}/tools/clang/include:$PWD/include"
+          fi
         fi
 
         export CB_PYTHON_DIR="$PWD/cppyy-backend/python"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ concurrency:
 jobs:
   prepare-dell:
     name: Activate self-host infrastructure
-    runs-on: self-hosted
+    runs-on: [self-hosted, spotter]
     steps:
       - name: Send Magic Packet
         env:
@@ -105,7 +105,7 @@ jobs:
             python-version: '3.14'
           # Ubuntu X86 Jobs
           - name: self-hosted-ubu22-x86-gcc12-clang-repl-21-cuda
-            os: self-hosted
+            os: [self-hosted, cuda, heavy]
             compiler: gcc-12
             clang-runtime: '21'
             llvm_enable_projects: "clang"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -297,6 +297,7 @@ jobs:
       uses: ./.github/actions/Miscellaneous/Save_PR_Info
 
     - name: Restore cached LLVM-${{ matrix.clang-runtime }} and ${{ matrix.cling == 'On' && 'Cling' || 'Clang-REPL' }} build
+      if: ${{ runner.environment != 'self-hosted' }}
       uses: actions/cache/restore@v4
       id: cache
       with:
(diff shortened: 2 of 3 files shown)


vgvassilev (approved):
LGTM!

.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ concurrency:
 jobs:
   prepare-dell:
     name: Activate self-host infrastructure
-    runs-on: self-hosted
+    runs-on: [self-hosted, spotter]
```
mcbarton, line 20: Did we add these labels to any self hosted runner? If not, it will probably wait until one exists. If one doesn't exist, this will probably just timeout after 6 hours.

aaronj0: cc @vgvassilev, the job still can't find the runner

.github/workflows/main.yml
```diff
@@ -105,7 +105,7 @@ jobs:
             python-version: '3.14'
           # Ubuntu X86 Jobs
           - name: self-hosted-ubu22-x86-gcc12-clang-repl-21-cuda
```
vgvassilev, line 107:
```suggestion
          - name: self-hosted-ubu22-x86-gcc11-clang-repl-21-cuda
```

.github/workflows/main.yml
```diff
@@ -105,7 +105,7 @@ jobs:
...
-            os: self-hosted
+            os: [self-hosted, cuda, heavy]
             compiler: gcc-12
```
vgvassilev, line 109:
```suggestion
            compiler: gcc-11
```

.github/workflows/main.yml
vgvassilev: @aaronj0, now seems to be further down the road. We should not really install stuff on these machines. Do we need the proposed apt-get packages?

vgvassilev: Ok, we need to drop the notion of clang-runtime -- why do we have it? We also should not install software on the self-hosted nodes and skip these steps, too.

mcbarton: In the ci the clang-runtime states which llvm/clang your building CppInterop against. Not sure how you get rid of it.

.github/workflows/main.yml
```diff
@@ -104,9 +104,9 @@ jobs:
             llvm_targets_to_build: "host;NVPTX"
             python-version: '3.14'
           # Ubuntu X86 Jobs
-          - name: self-hosted-ubu22-x86-gcc12-clang-repl-21-cuda
-            os: self-hosted
-            compiler: gcc-12
+          - name: self-hosted-ubu22-x86-gcc11-clang-repl-21-cuda
+            os: [self-hosted, cuda, heavy]
+            compiler: gcc-11
```
mcbarton: You can probably drop this (compiler in the matrix) altogether and skip the step which installs the compiler on the self hosted runners. cmake will then just default to using whatever compiler it can find on the runner.  If you know the runner has a particular compiler installer that you want to use, just set the CC and CXX environment variables accordingly on the self hosted runner.

mcbarton:
```
  sudo apt-get update (don't need)
  sudo apt-get install valgrind ninja-build (dont need to run test under Valgrind on self hosted runners)
  sudo apt-get install git g++ debhelper devscripts gnupg python3 doxygen graphviz python3-sphinx (most documentation stuff related. probably have python already installed)
  sudo apt-get install -y libc6-dbg (probably already installed on runner)
  sudo apt-get install valgrind (don't need valgrind. also this is repeated and should be removed for github hosted runners too) 
  sudo apt autoremove (not needed on self hosted runner)
  sudo apt clean (no needed on self hosted runner)
  # Install libraries used by the cppyy test suite
  sudo apt install libeigen3-dev (depends on whether self hosted runners have boost or eigen installed already. also don't need to test these on self hosted runners, since they already tested on github hosted runners)
  sudo apt install libboost-all-dev
```
@aaronj0 @vgvassilev This is what it tries to install when it fails. I have commented next to each one. I'd just skip the step altogether based on what i see.

aaronj0:
@vgvassilev Can we install ninja on the machine?

```
Updating files: 100% (160123/160123), done.
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
``

File: .github/workflows/main.yml [hunk 5]
```diff
@@ -104,12 +104,13 @@ jobs:
...
             clang-runtime: '21'
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
+            Valgrind: Off
```
mcbarton: off is the default so this isn't needed.

.github/workflows/.main.yml.swp
mcbarton, line 1: I'm assuming this added by mistake?

.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -8,15 +8,21 @@ runs:
...
+          LLVM_DIR="$(pwd)/llvm-project"
+          LLVM_BUILD_DIR="$(pwd)/llvm-project/build"
+          cling_on=$(echo "${{ matrix.cling }}" | tr '[:lower:]' '[:upper:]')
+          if [[ "${cling_on}" == "ON" ]]; then
+            CLING_DIR="$(pwd)/cling"
+            CLING_BUILD_DIR="$(pwd)/cling/build"
+            CPLUS_INCLUDE_PATH="${CLING_DIR}/tools/cling/include:${CLING_BUILD_DIR}/include:${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_BUILD_DIR}/include:${LLVM_BUILD_DIR}/tools/clang/include:$PWD/include"
+          else
+            CPLUS_INCLUDE_PATH="${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_BUILD_DIR}/include:${LLVM_BUILD_DIR}/tools/clang/include:$PWD/include"
+          fi
```
mcbarton, line 25: This change doesn't make any sense to me. This was setup so that CppInterOp builds against the llvm which is cached. Especially since we are assuming that we cannot install stuff on the runner as a general rule, I don't think we can asssume llvm-${{ matrix.clang-runtime }} will be installed, and that the system installed llvm was build with cuda support.

mcbarton: I think this change should be reverted.

aaronj0: The system is configured by @vgvassilev so we know that it has llvm21, I also added a step that installs it with apt,  which should be a one time thing

mcbarton: Does the llvm 21 that apt installs have cuda support, or is the package it installs only built to allow targetting host? If Vassil believes it is (I'm not convinced the one which comes with apt can), then we need to skip the step which loads, builds and saves the llvm cache, so the following isn't created

image

aaronj0: I believe the apt.llvm.org packages should be built with all targets including nvptx:
`/usr/lib/llvm-21/lib/cmake/llvm/LLVMConfig.cmake` contains line `LoongArch;Mips;MSP430;**NVPTX**;PowerPC;RISCV;Sparc;SPIRV;SystemZ;VE;We`. it looks like it is compiled for more or less every possible target.

Good catch on the cache step, pushing a fix

.github/workflows/main.yml
```diff
@@ -314,14 +315,24 @@ jobs:
...
       uses: ./.github/actions/Miscellaneous/Install_Dependencies
     
+    - name: Install LLVM-${{ matrix.clang-runtime }} from apt
+      if: ${{ runner.environment == 'self-hosted' }}
+      shell: bash
+      run: |
+        wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
+        echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-${{ matrix.clang-runtime }} main" | sudo tee /etc/apt/sources.list.d/llvm.list
+        sudo apt-get update
+        sudo apt-get install -y llvm-${{ matrix.clang-runtime }}-dev libclang-${{ matrix.clang-runtime }}-dev clang-${{ matrix.clang-runtime }} libpolly-${{ matrix.clang-runtime }}-dev libzstd-dev
```
mcbarton: If the system is configured with llvm 21 as you said here https://github.com/compiler-research/CppInterOp/pull/864#discussion_r2975650292, then this is unneeded. If this is some sort of backup in case the self hosted runner gets misconfigured in the future, and this is supposed to be a one time install for thos situations, then you should have a check at the start of this step to check whether llvm ${{ matrix.clang-runtime }} is installed, and skip over 
```
        wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
        echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-${{ matrix.clang-runtime }} main" | sudo tee /etc/apt/sources.list.d/llvm.list
        sudo apt-get update
        sudo apt-get install -y llvm-${{ matrix.clang-runtime }}-dev libclang-${{ matrix.clang-runtime }}-dev clang-${{ matrix.clang-runtime }} libpolly-${{ matrix.clang-runtime }}-dev libzstd-dev
```
if already installed.

.github/workflows/main.yml
```diff
@@ -104,8 +104,8 @@ jobs:
             llvm_targets_to_build: "host;NVPTX"
             python-version: '3.14'
           # Ubuntu X86 Jobs
-          - name: self-hosted-ubu22-x86-gcc12-clang-repl-21-cuda
-            os: self-hosted
+          - name: self-hosted-ubu22-x86-gcc11-clang-repl-21-cuda
```
mcbarton:
```suggestion
          - name: self-hosted-ubu22-x86-gcc12-clang-repl-21-cuda
```

.github/workflows/main.yml
aaronj0: Dropped adding the sources, and the apt update and install commands will be a no op since it is already present, but a nice to have in case these dependencies go missing.

.github/actions/Build_and_Test_CppInterOp/action.yml
aaronj0: I believe all these steps are skipped and I don't see any new cache showing up from this PR so I assume thats fixed.

PR #866, opened 2026-03-23
Fix CUDA tests

aaronj0:
Drop hardcoded CUDA arch provided to the `IncrementalCompilerBuilder` in `createClangInterpreter`. While debugging issues with the CUDA support that seems broken on master, after fixing the interpreter creation, I found that the second interpreter creation held an error in cudaGetLastError. CUDA and HIP reuse that option so it would not indicate whether the interpreter was CUDA enabled. All 4 CUDA tests pass locally with an `sm_89` device, LLVM 21

### DIFF
--- a/lib/CppInterOp/Compatibility.h
+++ b/lib/CppInterOp/Compatibility.h
@@ -7,11 +7,31 @@
 
 #include "clang/AST/DeclTemplate.h"
 #include "clang/AST/GlobalDecl.h"
+#include "clang/Basic/DiagnosticIDs.h"
+#include "clang/Basic/DiagnosticOptions.h"
+#if CLANG_VERSION_MAJOR < 21
+#include "clang/Basic/Cuda.h"
+#else
+#include "clang/Basic/OffloadArch.h"
+#endif
 #include "clang/Basic/SourceLocation.h"
 #include "clang/Basic/Specifiers.h"
 #include "clang/Basic/Version.h"
 #include "clang/Config/config.h"
+#include "clang/Driver/Compilation.h"
+#include "clang/Driver/Driver.h"
+#include "clang/Driver/Options.h"
+#include "clang/Frontend/TextDiagnosticBuffer.h"
 #include "clang/Sema/Sema.h"
+
+#include "llvm/ADT/IntrusiveRefCntPtr.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/FileSystem.h"
+
+#include <cstdint>
+#include <cstring>
+#include <memory>
+#include <string>
 
 #ifdef _MSC_VER
 #define dup _dup
@@ -189,11 +209,12 @@ inline void codeComplete(std::vector<std::string>& Results,
 #include "clang/Interpreter/Interpreter.h"
 #include "clang/Interpreter/Value.h"
 
+#include "llvm/Support/DynamicLibrary.h"
 #include "llvm/Support/Error.h"
+#include "llvm/TargetParser/Host.h"
 
 #ifdef LLVM_BUILT_WITH_OOP_JIT
 #include "clang/Basic/Version.h"
-#include "llvm/TargetParser/Host.h"
 
 #include "llvm/ExecutionEngine/Orc/Debugging/DebuggerSupport.h"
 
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/Compatibility.h
```diff
@@ -204,6 +205,38 @@ inline void codeComplete(std::vector<std::string>& Results,
...
+  std::string Err;
+  auto Lib = llvm::sys::DynamicLibrary::getPermanentLibrary(
+#ifdef _WIN32
+      "nvcuda.dll",
+#else
+      "libcuda.so.1",
+#endif
+      &Err);
+  if (!Lib.isValid())
+    return "";
```
vgvassilev: We should ask the clang::Driver to find the relevant cuda folder (using the passed createInterpreter GPU arguments).

lib/CppInterOp/Compatibility.h
```diff
@@ -224,9 +257,17 @@ createClangInterpreter(std::vector<const char*>& args, int stdin_fd = -1,
...
-    // FIXME: Parametrize cuda-path and offload-arch.
-    CB.SetOffloadArch("sm_35");
+    for (const auto* arg : args)
+      if (llvm::StringRef(arg).starts_with("--offload-arch="))
+        OffloadArch = llvm::StringRef(arg).substr(strlen("--offload-arch="));
+    if (OffloadArch.empty())
+      OffloadArch = detectNVIDIAGPUArch();
+    if (OffloadArch.empty())
+      OffloadArch = "sm_35";
+
```
mcbarton, line 380: Do we want this fallback, or should we error is such a circumstance? e.g. should error telling the user to provide the gpu architechture via `--offload-arch`?

lib/CppInterOp/Compatibility.h
```diff
@@ -224,9 +257,17 @@ createClangInterpreter(std::vector<const char*>& args, int stdin_fd = -1,
   CB.SetCompilerArgs({args.begin(), it});
 
   std::unique_ptr<clang::CompilerInstance> DeviceCI;
+  std::string OffloadArch;
   if (CudaEnabled) {
-    // FIXME: Parametrize cuda-path and offload-arch.
-    CB.SetOffloadArch("sm_35");
+    for (const auto* arg : args)
+      if (llvm::StringRef(arg).starts_with("--offload-arch="))
+        OffloadArch = llvm::StringRef(arg).substr(strlen("--offload-arch="));
```
mcbarton: This is probably a stupid question. Do we have any logic to check that the user provided gpu architechture is a valid one? I'm saying this assuming the answer is yes.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3497,6 +3497,13 @@ TInterp_t GetInterpreter() {
 InterpreterLanguage GetLanguage(TInterp_t I /*=nullptr*/) {
   compat::Interpreter* interp = &getInterp(I);
   const auto& LO = interp->getCI()->getLangOpts();
+
+  // CUDA and HIP reuse C++ language standards, so LangStd alone reports CXX.
```
mcbarton, line 3501: This comment isn't specific to this PR, but given we are looking at gpu related stuff, can the interpreter handle, c++ + openmp (gpu offloading)?

lib/CppInterOp/Compatibility.h
```diff
@@ -204,6 +205,38 @@ inline void codeComplete(std::vector<std::string>& Results,
...
+  LOAD(cuDeviceGet, int, uint32_t*, int)
+  LOAD(cuDeviceGetAttribute, int, int*, int, uint32_t)
+#undef LOAD
+
+  uint32_t dev;
+  int maj, min;
+  if (cuInit(0) || cuDeviceGet(&dev, 0) ||
+      cuDeviceGetAttribute(&maj, /*MAJOR*/ 75, dev) ||
+      cuDeviceGetAttribute(&min, /*MINOR*/ 76, dev))
+    return "";
```
vgvassilev:
```suggestion
    return "sm_35";
```

lib/CppInterOp/Compatibility.h
```diff
@@ -204,6 +205,38 @@ inline void codeComplete(std::vector<std::string>& Results,
 
 namespace compat {
 
+/// detect the compute capability of the GPU using the cuda driver.
+/// returns: "sm_xx", or "" on failure.
+inline std::string detectNVIDIAGPUArch() {
```
vgvassilev:
```suggestion
inline bool detectNVIDIAGPUArch(std::string &Version) {
```

and it needs to return true/false on error.

lib/CppInterOp/Compatibility.h
```diff
@@ -224,9 +257,17 @@ createClangInterpreter(std::vector<const char*>& args, int stdin_fd = -1,
...
   if (CudaEnabled) {
-    // FIXME: Parametrize cuda-path and offload-arch.
-    CB.SetOffloadArch("sm_35");
+    for (const auto* arg : args)
+      if (llvm::StringRef(arg).starts_with("--offload-arch="))
+        OffloadArch = llvm::StringRef(arg).substr(strlen("--offload-arch="));
+    if (OffloadArch.empty())
+      OffloadArch = detectNVIDIAGPUArch();
+    if (OffloadArch.empty())
+      OffloadArch = "sm_35";
```
vgvassilev:
```suggestion

```

unittests/CppInterOp/CUDATest.cpp
```diff
@@ -10,42 +10,28 @@
 
 using namespace TestUtils;
 
-static bool HasCudaSDK() {
-  auto supportsCudaSDK = []() {
+// perform SDK and runtime checks on a single CUDA interpreter
+static std::pair<bool, bool> CheckCudaSupport() {
```
vgvassilev:
```suggestion
struct CudaInfo {
  bool X;
  bool Y;
};
static CudaInfo CheckCudaSupport() {
```

lib/CppInterOp/Compatibility.h
vgvassilev: I guess we will need some generic way to rely on what's already in clang. Here is what gemini proposes and I think it is worth investigating:
```cpp
#include "clang/Driver/CudaInstallationDetector.h"
#include "clang/Driver/Options.h"
#include "llvm/Option/ArgList.h"
#include "llvm/Option/OptTable.h"
#include "llvm/TargetParser/Host.h"

// ... inside createClangInterpreter ...

// 1. Parse the raw args into a format the Detector understands
const auto &Opts = clang::driver::getDriverOptTable();
unsigned MissingArgIndex, MissingArgCount;
// Use 0 as the included/excluded flags for a general parse
auto InputArgs = Opts.ParseArgs(args, MissingArgIndex, MissingArgCount);

// 2. Identify the Triple (Host is usually the intended target for JIT)
llvm::Triple HostTriple(llvm::sys::getDefaultTargetTriple());

// 3. Initialize the Detector (No CI needed!)
// The detector will look for --cuda-path in InputArgs automatically
clang::driver::CudaInstallationDetector Detector(
    /*Driver=*/ {}, // We can pass an empty driver object or a dummy
    HostTriple, 
    InputArgs);

if (CudaEnabled && Detector.isValid()) {
    // 4. Resolve the Architecture
    std::string SelectedArch;
    if (auto *Arg = InputArgs.getLastArg(clang::driver::options::OPT_offload_arch_EQ)) {
        SelectedArch = Arg->getValue();
    } else {
        SelectedArch = detectNVIDIAGPUArch();
    }

    // 5. Validation Logic
    if (!SelectedArch.empty()) {
        clang::CudaArch CArch = clang::StringToCudaArch(SelectedArch);
        // If the hardware is newer than the SDK, downgrade to the SDK's max supported arch
        if (Detector.version() < clang::MinVersionForCudaArch(CArch)) {
            CArch = clang::MaxVersionForCudaArch(Detector.version());
            SelectedArch = clang::CudaArchToString(CArch);
        }
        CB.SetOffloadArch(SelectedArch);
    }
}
```

The advantage is that now you know where the sdk is and you also can pass the relevant include paths automatically:

```cpp
if (CudaEnabled && Detector.isValid()) {
    // 1. Add the CUDA include directory (e.g., /usr/local/cuda/include)
    // We use -isystem to treat them as system headers (suppressing warnings)
    std::string IncludeFlag = "-isystem" + Detector.getIncludePath();
    
    // Note: Ensure the string's lifetime matches the 'args' usage. 
    // If 'args' is a vector of const char*, you may need to manage the allocation.
    args.push_back(strdup(IncludeFlag.c_str()));

    // 2. Add the path to the CUDA builtins (libdevice)
    // This is required for math functions like __sinf, __expf, etc.
    std::string LibDevicePath = Detector.getLibDevicePath();
    if (!LibDevicePath.empty()) {
        std::string LinkBitcode = "--cuda-device-lib=" + LibDevicePath;
        args.push_back(strdup(LinkBitcode.c_str()));
    }
}
```

lib/CppInterOp/Compatibility.h
aaronj0: The PR now uses the same fallback as the clang driver: `clang::OffloadArch::CudaDefault` which is sm_52

lib/CppInterOp/CppInterOp.cpp
aaronj0: I believe you need to pass `-fopenmp` to the interpreter, but so far only CPU has been tested and not GPU offloading

lib/CppInterOp/Compatibility.h
```diff
@@ -7,10 +7,17 @@
 
 #include "clang/AST/DeclTemplate.h"
 #include "clang/AST/GlobalDecl.h"
+#include "clang/Basic/DiagnosticIDs.h"
+#include "clang/Basic/DiagnosticOptions.h"
+#include "clang/Basic/OffloadArch.h"
```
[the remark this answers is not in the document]
aaronj0: This header was added in llvm21, is version guarded and points to Cuda.h for  < 20

lib/CppInterOp/Compatibility.h
```diff
@@ -7,10 +7,17 @@
...
 #include "clang/Basic/SourceLocation.h"
 #include "clang/Basic/Specifiers.h"
 #include "clang/Basic/Version.h"
 #include "clang/Config/config.h"
+#include "clang/Driver/Compilation.h"
+#include "clang/Driver/Driver.h"
+#include "clang/Driver/Options.h"
+#include "clang/Frontend/TextDiagnosticBuffer.h"
```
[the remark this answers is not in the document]
aaronj0: Removed

lib/CppInterOp/Compatibility.h
aaronj0: Nice! Implemented

lib/CppInterOp/Compatibility.h
aaronj0: We now rely on the default set by clang in OffloadArch: `clang::OffloadArch::CudaDefault`

unittests/CppInterOp/CUDATest.cpp
aaronj0: Addressed, thanks

lib/CppInterOp/Compatibility.h
aaronj0: This is somewhat what the current implementation does, we use clang::Driver to create the `clang::driver::CudaInstallationDetector` and iterate over the paths it inserts based on the CUDA directory

aaronj0: This is done by creating a compilation object with the driver and fetching the arg with this `clang::driver::options::OPT_cuda_path_EQ` enum value

lib/CppInterOp/Compatibility.h
```diff
@@ -7,12 +7,30 @@
 
 #include "clang/AST/DeclTemplate.h"
 #include "clang/AST/GlobalDecl.h"
+#include "clang/Basic/DiagnosticIDs.h"
+#include "clang/Basic/DiagnosticOptions.h"
+#if CLANG_VERSION_MAJOR >= 21
+#include "clang/Basic/OffloadArch.h"
+#else
+#include "clang/Basic/Cuda.h"
```
vgvassilev:
```suggestion
#if CLANG_VERSION_MAJOR < 21
#include "clang/Basic/Cuda.h"
#else
#include "clang/Basic/OffloadArch.h"
```

lib/CppInterOp/Compatibility.h
```diff
@@ -204,29 +223,163 @@ inline void codeComplete(std::vector<std::string>& Results,
...
+/// \returns true on success, false if not found.
+inline bool detectCudaInstallPath(const std::vector<const char*>& args,
+                                  std::string& CudaPath) {
+  // minimal driver that runs CudaInstallationDetector internally
+  std::string TT = llvm::sys::getProcessTriple();
+  llvm::IntrusiveRefCntPtr<clang::DiagnosticIDs> DiagID(
+      new clang::DiagnosticIDs());
+  // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
+  auto* DiagsBuffer = new clang::TextDiagnosticBuffer;
+#if CLANG_VERSION_MAJOR >= 21
```
vgvassilev:
```suggestion
#if CLANG_VERSION_MAJOR < 21
```

unittests/CppInterOp/CUDATest.cpp
```diff
@@ -7,50 +7,41 @@
...
+    return {true, rtOk};
+  }();
+  return result;
 }
 
-#ifdef CPPINTEROP_USE_CLING
-TEST(DISABLED_CUDATest, Sanity) {
-#else
+static bool HasCudaSDK() { return CheckCudaSupport().HasSDK; }
+static bool HasCudaRuntime() { return CheckCudaSupport().HasRuntime; }
```
vgvassilev: We would need a cuda-specific test fixture to avoid re-executing the checks every test.

aaronj0: Is that supposed to prevent running the other tests? Edit: I understand now, adding a fixture

aaronj0: Done

lib/CppInterOp/Compatibility.h
aaronj0: Done

lib/CppInterOp/Compatibility.h
aaronj0: Done

vgvassilev (approved):
LGTM!

unittests/CppInterOp/CUDATest.cpp
```diff
@@ -7,91 +7,116 @@
...
+  std::vector<const char*> args;
+
+  // no args: rely on auto-detect
+  EXPECT_TRUE(compat::detectCudaInstallPath(args, path));
+  EXPECT_TRUE(llvm::StringRef(path).starts_with("/usr/local/cuda"));
+
+  // explicit path: returns cuda-12.3 if it exists, otherwise may auto-detect
+  args = {"--cuda-path=/usr/local/cuda-12.3"};
+  if (compat::detectCudaInstallPath(args, path)) {
+    if (llvm::sys::fs::is_directory("/usr/local/cuda-12.3/include"))
```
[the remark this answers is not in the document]
aaronj0: Resolved

PR #867, opened 2026-03-24
[ci] Rely on setup compiler step for gcc12, add apt-get update

aaronj0:
#864 passed the jobs on the CI but a fresh run on main hit an error with the apt installation of gcc in "Setup Compiler". This indicates that the setup is in fact stateless based on how github configures the runner, and we require the setup steps without assuming dependencies are installed. Which is more ideal.
This PR should resolve the seen issue, and drops redundant steps (update, gcc install) for the self hosted runner block (only checks llvm21) and installs dependencies.


vgvassilev (approved):
LGTM!

aaronj0:
Merging as the Emscripten jobs are unrelated to these changes and the self-hosted, and other native jobs pass

PR #868, opened 2026-03-25
[ci] Move the coverage to the cuda bot.

vgvassilev:
We also do not need to have more than one machine running codecov.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -138,14 +138,6 @@ jobs:
             oop-jit: On
             Valgrind: On
             python-version: '3.13'
-          - name: ubu24-x86-gcc12-clang-repl-20-out-of-process-coverage
-            os: ubuntu-24.04
-            compiler: gcc-12
-            clang-runtime: '20'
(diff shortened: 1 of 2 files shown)


mcbarton (approved):
Its interesting coverage is set twice. I remember it being set to the out of process job. Not sure if there will be, but I suspect there might be a drop in coverage as part of this PR.

Once we have llvm 22 support, we should turn oop on for the self hosted runner and update the llvm version to 22.

.github/workflows/main.yml
```diff
@@ -135,7 +136,6 @@ jobs:
             llvm_enable_projects: "clang;compiler-rt"
             llvm_targets_to_build: "host;NVPTX"
             oop-jit: On
-            coverage: true
             Valgrind: On
             python-version: '3.13'
           - name: ubu24-x86-gcc12-clang-repl-20-out-of-process-coverage
```
mcbarton: This job can ne removed now. Its only purpose was to generate the coverage report

vgvassilev: I assume when we move to 22 we will have in-process and out-of-process combined.

PR #870, opened 2026-03-25
Fix `GetLanguage` for CUDA/HIP where CXX standard was returned

aaronj0:
Originally part of #866, can land as an independent change after it is merged, and should fix the test

Edit: The `GetLanguage` interface failed for CUDA/HIP where the CXX standard was returned instead because we only check `LangStd`, we should check `LangOpts` for this


vgvassilev (approved):
LGTM, assuming we provide tests for this code in the other PR.

PR #873, opened 2026-03-25
[ci] Shallow clone in ci and update job names

mcbarton:
For some reason the checkout action was set to clone the whole cppinterop history when doing the ci, when we only need the latest commit. On Windows the checkout action takes around 16 seconds on main, so changing to the default of doing a shallow clone should provide a speed up. Although small per job, it should all add up.

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -30,8 +30,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v6
-      with:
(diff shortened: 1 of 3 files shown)


Vipul-Cariappa (approved):
LGTM!

Should we also consider using the GitHub action to checkout other repos? I assume we already do a depth 1 clone for them.

vgvassilev (approved):
Lgtm!

PR #874, opened 2026-03-26
[debug] CUB tests on NUC machine

aaronj0:


unittests/CppInterOp/CUDATest.cpp
```diff
@@ -31,8 +32,15 @@ class CUDATest : public ::testing::Test {
...
 
     HasSDK = true;
-    HasRuntime =
-        !Cpp::Declare("int __cuda_err = (int)cudaGetLastError();" DFLT_FALSE);
+
+    Cpp::Declare("cudaDeviceSynchronize();" DFLT_FALSE);
+    bool evalErr = false;
+    int cudaErr = (int)Cpp::Evaluate("(int)cudaGetLastError()", &evalErr);
+    if (evalErr || cudaErr != 0)
+      printf("SetUpTestSuite: CUDA kernel execution failed (cudaError=%d)\n",
```
vgvassilev: This should go via the gtest logging logic.

unittests/CppInterOp/CUDATest.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "gtest/gtest.h"
 
+#include <iostream>
```
vgvassilev: This does not seem needed.

unittests/CppInterOp/CUDATest.cpp
```diff
@@ -31,25 +30,42 @@ class CUDATest : public ::testing::Test {
...
-    HasRuntime =
-        !Cpp::Declare("int __cuda_err = (int)cudaGetLastError();" DFLT_FALSE);
+
+    if (Cpp::Declare("cudaDeviceSynchronize();" DFLT_FALSE) != 0) {
+      Cpp::DeleteInterpreter(I);
+      return;
+    }
+    bool evalErr = false;
+    int cudaErr = (int)Cpp::Evaluate("(int)cudaGetLastError()", &evalErr);
+    if (evalErr || cudaErr != 0)
```
aaronj0, line 40: @vgvassilev FYI, i'm also relying on the failure of evaluating the expression here, instead of assuming that would always succeed

vgvassilev (approved):
Lgtm!

PR #881, opened 2026-03-31
[win] Fix compilation of `clang::driver::options::OPT_` for CUDA

aaronj0:
https://github.com/compiler-research/CppInterOp/commit/6ab18fdf3faa6c70bbe8b34d657d13dda4507b1c introduced the usage of `clang::driver::options::OPT_` to detect GPU compute capability using `clang::Driver` but this does not  compile on windows without `-Zc:preprocessor` see https://github.com/llvm/llvm-project/issues/86028
I'm confused why we didn't catch this on our CI, but this was caught on ROOT


vgvassilev (approved):
LGTM (assuming the ci passes)!

PR #882, opened 2026-03-31
Change clangCppInterOp to CppInterOp and remove old header

mcbarton:
Given the next version has been labelled as version 2, and breaking changes are implied, it might be worth taking this opportunity to delete the deprecated `clang/Interpreter/CppInterOp.h` header.


mcbarton:
@vgvassilev I saw you manually rerun the Emscripten workflow for this PR. To make that ci pass I would a way for xeus-cpp to link and load the CppInterOp library in a way which is backwards compatible. I have a build where I am debugging this locally. This PR also will require a change to cppyy-backend before going in (you will see the ci is currently using a branch of cppyy-backend).

The other option is to deal with this PR just before a CppInterOp release (i.e. take it in, accepting that  the deployment will be temporarily broken, release, then correct the library name is xeus-cpp after release)

PR #883, opened 2026-04-01
Drop the broken suppression mechanism.

vgvassilev:
This suppression file tells the diagnostics engine to discard all diagnostics. Then the incremental infrastructure has no way to understand if something was ill-formed. That leads to even more subtle issues because the broken AST is not reverted but sent to CodeGen.

A proper solution should use a DiagnosticConsumer which discards the errors on the screen but does not keep the underlying infrastructure blind.


vgvassilev:
This landed in a different way in main #901

PR #884, opened 2026-04-02
[ci] Improve cppyy testing with requirements.txt and pytest-xdist

aaronj0:
If the multithreaded pytest fails on OS X, that's a bug in our test setup. Tests should work in a standalone isolated fashion (`-k test_name`)

### DIFF
--- a/.github/actions/Build_and_Test_cppyy/action.yml
+++ b/.github/actions/Build_and_Test_cppyy/action.yml
@@ -52,7 +52,9 @@ runs:
         # Install cppyy
         git clone --depth=1 https://github.com/compiler-research/cppyy.git
         cd cppyy
+        python -m pip install --upgrade pip
         python -m pip install uv
+        python -m pip install -r requirements.txt
         python -m pip install --upgrade . --no-deps
         cd ..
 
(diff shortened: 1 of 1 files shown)


.github/actions/Build_and_Test_cppyy/action.yml
```diff
@@ -74,18 +76,12 @@ runs:
...
-        python -m pip install pytest
-        python -m pip install pytest-xdist
-        if [[ "${{ matrix.python-version }}" != "3.14" ]]; then
-          python -m uv pip install numba==0.61.2
-        fi
         echo ::endgroup::
         echo ::group::Run complete test suite
         set -o pipefail
-        python -m pytest -sv -ra | tee complete_testrun.log 2>&1
+        python -m pytest -n ${{ env.ncpus }} -sv -ra | tee complete_testrun.log 2>&1
```
mcbarton: See here for my I previously reject of this `-n ${{ env.ncpus }}` change https://github.com/compiler-research/CppInterOp/pull/795#pullrequestreview-3745429458 . We really need a `pytest.toml` so all these options are controlled from within the cppyy repo

mcbarton (dismissed):

.github/actions/Build_and_Test_cppyy/action.yml
aaronj0: Number of cpus should not be placed in a toml file.
edit: that doesn't play well with OS X and the cppyy interpreter state. To be debugged at a later stage.

vgvassilev (approved):
Lgtm!

aaronj0:
No longer applies with the move to cppjit

PR #885, opened 2026-04-02
Teach GetType and friends about typedef names.

vgvassilev:


aaronj0 (approved):
LGTM! I would rerun after rebasing on https://github.com/compiler-research/CppInterOp/pull/884, which should then run numba and leakcheck tests

PR #886, opened 2026-04-03
Rework the way we find builtins.

vgvassilev:
The builtin types can be platform- and target-dependent (see ASTContext::InitBuiltinTypes). The new approach iterates over all registered builtin types and adds their string representation to a map which can be searched efficiently.


aaronj0 (approved):
LGTM!

PR #891, opened 2026-04-07
Suppress llvm_shutdown for external-interpreter clients.

vgvassilev:
sInterpreters used to leak across process exit (FIXME-documented). Add an InterpreterShutdown RAII guard that drains it via `llvm_shutdown`, and a `SkipShutDown` flag that `UseExternalInterpreter` sets to suppress the call -- otherwise a CppInterOp-managed exit would tear down the host's LLVM. The guard is declared after `sInterpreters` in `GetInterpreters` so its dtor fires first in the static-dtor phase (LIFO), giving llvm_shutdown live LLVM globals to work against.

The `llvm_shutdown` call itself is gated on LLVM_VERSION_MAJOR > 22. That is the version with `Platform::lookupResolvedInitSymbols` (https://github.com/llvm/llvm-project/pull/196874), which makes ~Interpreter's JIT deinit skip lazy materialization; older LLVM SEGFAULTs in cleanUp against destroyed function-local statics, so the dtor stays a no-op and `sInterpreters` keeps leaking instead.

Also restore the pipe / OOM / crash handlers lost when InitLLVM was dropped (skipped for external interpreters -- those decisions belong to the host). Tests pin JIT static dtors fire on `Cpp::DeleteInterpreter`, and mirror clang's `ShutdownDoesNotMaterializeAgainstDestroyedGlobals` gated on LLVM 23+.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -65,6 +65,7 @@
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/Error.h"
 #include "llvm/Support/FileSystem.h"
+#include "llvm/Support/InitLLVM.h"
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/Process.h"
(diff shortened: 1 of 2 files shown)


vgvassilev:
@aaronj0, can you take a look at this patch and see if it does not break ROOT?

vgvassilev:
The proper fix will depend on landing: https://github.com/llvm/llvm-project/pull/196874 upstream.

PR #893, opened 2026-04-13
Implement basic tracing support

vgvassilev:
Introduce INTEROP_TRACE/INTEROP_RETURN/INTEROP_VOID_RETURN macros that record API call sequences with timing, and INTEROP_OUT for out-parameter handle registration. On crash, if CPPINTEROP_LOG=1, the signal handler writes a standalone .cpp reproducer to a temp file; otherwise it suggests re-running with the env var set.

This PR also annotates the public api.


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -51,6 +52,7 @@
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Demangle/Demangle.h"
+#include <llvm/ExecutionEngine/JITSymbol.h>
```
```suggestion
#include "llvm/ExecutionEngine/JITSymbol.h"
```

For consistency...

lib/CppInterOp/Tracing.h
```diff
@@ -0,0 +1,381 @@
...
+    if (!m_Data->Returned) {
+      llvm::errs() << "ERROR: Function '" << m_Data->Name
+                   << "' exited without calling INTEROP_RETURN!\n";
+      assert(
+          m_Data->Returned &&
+          "Unannotated exit branch detected: use `return INTEROP_RETURN(...)`");
+    }
+
+    auto EndTime = llvm::TimeRecord::getCurrentTime(false).getWallTime();
+    auto Dur = static_cast<long long>((EndTime - m_Data->StartTime) * 1e9);
```
Vipul-Cariappa, line 292: Why not use `uint64_t`?

lib/CppInterOp/Tracing.h
```diff
@@ -0,0 +1,381 @@
...
+  };
+};
+
+} // namespace Tracing
+} // namespace CppInterOp
+
+#ifdef _MSC_VER
+#define INTEROP_FUNC_SIG __FUNCSIG__
+#else
+#define INTEROP_FUNC_SIG __PRETTY_FUNCTION__
```
Vipul-Cariappa, line 429: FYI, `__PRETTY_FUNCTION__` gives different results depending on GCC and Clang.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -379,29 +401,47 @@ bool JitCall::AreArgumentsValid(void* result, ArgList args, void* self,
...
                            /*Qualified=*/true);
   LLVM_DEBUG(dbgs() << "Run '" << Name << "', compiled at: "
                     << (void*)m_GenericCall << " with result at: " << result
                     << " , args at: " << args.m_Args << " , arg count: "
                     << args.m_ArgSize << " , self at: " << self << "\n";);
+
+  if (auto* TI = CppInterOp::Tracing::TraceInfo::TheTraceInfo) {
+    std::string SelfPart = self ? TI->lookupHandle(self) : "";
+    std::string Entry =
+        llvm::formatv("  // JitCall::Invoke {0}(nargs={1}, self={2})", Name,
```
Vipul-Cariappa, line 415: Is `Name` equivalent to `__PRETTY_FUNCTION__`? It would be nice if that were possible.

Vipul-Cariappa (commented):
LGTM!
There is a lot of template magic, which becomes difficult to read/review. But I guess this is necessary.

lib/CppInterOp/Tracing.h
```diff
@@ -0,0 +1,381 @@
...
+    m_HandleMap.clear();
+    m_Log.clear();
+    m_VarCount = 0;
+  }
+
+  static TraceInfo* TheTraceInfo;
+};
+
+/// Activate tracing. Must be called at most once.
+void InitTracing();
```
mcbarton:
```suggestion
CPPINTEROP_API void InitTracing();
```

lib/CppInterOp/Tracing.h
```diff
@@ -0,0 +1,381 @@
...
+
+  void appendToLog(const std::string& line) { m_Log.push_back(line); }
+  const std::vector<std::string>& getLog() const { return m_Log; }
+  std::string getLastLogEntry() const {
+    return m_Log.empty() ? "" : m_Log.back();
+  }
+
+  /// Write the accumulated reproducer log to a file.
+  /// Uses only pre-collected data and simple writes.
+  std::string writeToFile();
```
mcbarton:
```suggestion
  CPPINTEROP_API std::string writeToFile();
```

lib/CppInterOp/Tracing.h
```diff
@@ -0,0 +1,381 @@
...
+    }
+    // Clear timers after clearing the group to suppress the report.
+    m_TG.clear();
+    m_Timers.clear();
+    m_HandleMap.clear();
+    m_Log.clear();
+    m_VarCount = 0;
+  }
+
+  static TraceInfo* TheTraceInfo;
```
mcbarton:
```suggestion
  CPPINTEROP_API static TraceInfo* TheTraceInfo;
```

mcbarton (commented):
These changes should fix the Emscripten build.

mcbarton:
@vgvassilev You will need the Cppinterop_api annotations below this message to fix the Emscripten build https://github.com/compiler-research/CppInterOp/pull/893#pullrequestreview-4107526731

lib/CppInterOp/exports.ld
```diff
@@ -55,3 +55,4 @@
 -Wl,--export=_ZN5clang11Interpreter6createENSt3__210unique_ptrINS_16CompilerInstanceENS1_14default_deleteIS3_EEEENS2_IN4llvm3orc12LLJITBuilderENS4_IS9_EEEE
 -Wl,--export=_ZNK5clang13CXXRecordDecl19isInjectedClassNameEv
 -Wl,--export=_ZNK5clang8QualType11getAsStringERKNS_14PrintingPolicyE
+-Wl,--export=_ZN10CppInterOp7Tracing11InitTracingEv
```
mcbarton: This is unneeded. You need to use the CppInterOp_API to avoid adding this symbol here. This file only currently contains llvm and clang symbols. CppInterOp symbols were all removed by use the the API annotation in the past, so we shouldn't start adding CppInterOp symbols back into this file.

mcbarton (changes requested):

lib/CppInterOp/exports.ld
vgvassilev: Yes, I figured that out -- let me clean it...

unittests/CMakeLists.txt
```diff
@@ -37,6 +37,14 @@ function(add_cppinterop_unittest name)
...
   add_executable(${name} EXCLUDE_FROM_ALL ${ARG_UNPARSED_ARGUMENTS})
+  if(NOT LLVM_ENABLE_RTTI)
+    if(MSVC)
+      target_compile_options(${name} PRIVATE "/GR-")
+    else()
+      target_compile_options(${name} PRIVATE "-fno-rtti")
+    endif()
+  endif()
+  add_dependencies(CppInterOpUnitTests ${name})
   add_dependencies(CppInterOpUnitTests ${name})
```
mcbarton: Duplication of  `add_dependencies(CppInterOpUnitTests ${name})`

unittests/CMakeLists.txt
```diff
@@ -37,6 +37,14 @@ function(add_cppinterop_unittest name)
     ${ARGN}
   )
   add_executable(${name} EXCLUDE_FROM_ALL ${ARG_UNPARSED_ARGUMENTS})
+  if(NOT LLVM_ENABLE_RTTI)
+    if(MSVC)
+      target_compile_options(${name} PRIVATE "/GR-")
+    else()
+      target_compile_options(${name} PRIVATE "-fno-rtti")
+    endif()
+  endif()
```
mcbarton, line 46: Is this for the case where you don't have the `include(HandleLLVMOptions)` here https://github.com/compiler-research/CppInterOp/blob/c7d9447ef4266fece90a5c762003b5c2b233d845/CMakeLists.txt#L289

mcbarton (approved):
LGTM modulo my comments here https://github.com/compiler-research/CppInterOp/pull/893/changes#r3084798893 and here https://github.com/compiler-research/CppInterOp/pull/893/changes#r3082475745

unittests/CMakeLists.txt
vgvassilev: HandleLLVMOptions doesn't propagate RTTI flags to plain add_executable targets however, the next line adds the dependencies twice. Thanks for pointing this out!

PR #896, opened 2026-04-16
Only build Emscripten static library for nightly build (speeds up Emscripten development for PRs)

mcbarton:
If it works, this should mean that the Emscripten static library is only tested for nightly builds. This feels likely to be ok, since we only have one test failing for the static build which passes for the shared library build. The static library build is much more compute intensive, since linking the static library to the test executables takes a long time. This should help further alleviate the bottleneck the Emscripten ci has posed on development.

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -27,7 +27,7 @@ jobs:
             llvm_enable_projects: "clang;lld"
             llvm_targets_to_build: "WebAssembly"
             emsdk_ver: "4.0.9"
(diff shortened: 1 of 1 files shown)


mcbarton:
With this change, for PRs the Emscripten ci will just be one 19 minute job, so comparable to the native ci. The read the docs failure has nothing to do with this PR. That failed as the read the docs runner failed to install the packages it needed, and it ended up timing out.

vgvassilev (approved):
LGTM!

PR #898, opened 2026-04-20
Enable sanitizers

vgvassilev:


.github/actions/Build_LLVM/action.yml
```diff
@@ -54,6 +54,7 @@ runs:
                 -DCLANG_ENABLE_BOOTSTRAP=OFF                        \
                 -G Ninja                                            \
                 -DLLVM_INCLUDE_BENCHMARKS=OFF                   \
+                -DLLVM_USE_SANITIZER="${{ matrix.sanitizer }}"  \
```
mcbarton, line 57: To be able to build a cache with the santizer, you'll need to append to this key here https://github.com/vgvassilev/CppInterOp/blob/d9e1f39cbd031dd43cc7cc8ec9e909d5ac5ebdad/.github/workflows/main.yml#L304 . Probably will be good enough to append `${{ matrix.sanitizer || '' }}`

.github/workflows/main.yml
```diff
@@ -293,7 +301,7 @@ jobs:
         path: |
           llvm-project
           ${{ matrix.cling=='On' && 'cling' || '' }}
-        key: ${{ env.CLING_HASH }}-${{ runner.os }}-${{ matrix.os }}-${{ matrix.compiler }}-clang-${{ matrix.clang-runtime }}-${{ matrix.root-llvm-tag || 'none' }}.x-patch-${{ hashFiles(format('patches/llvm/clang{0}-*.patch', matrix.clang-runtime)) || 'none' }}${{ matrix.oop-jit == 'On' && '-oop' || '' }}
+        key: ${{ env.CLING_HASH }}-${{ runner.os }}-${{ matrix.os }}-${{ matrix.compiler }}-clang-${{ matrix.clang-runtime }}-${{ matrix.root-llvm-tag || 'none' }}-${{ matrix.sanitizer || '' }}.x-patch-${{ hashFiles(format('patches/llvm/clang{0}-*.patch', matrix.clang-runtime)) || 'none' }}${{ matrix.oop-jit == 'On' && '-oop' || '' }}
```
mcbarton: @vgvassilev I cancelled the workflow run after this change. By adding the `-` and placing it in the middle, you had made it so that every ci job was now building a new cache, which would bring everything to a halt. You need to just put `${{ matrix.sanitizer || '' }}` at the end of the key, without any `-`. Otherwise you will bring the ci in every repo to a grinding halt.

mcbarton (dismissed):

.github/workflows/main.yml
vgvassilev: Please suggest as a suggestion the right fix that I can steal :)

mcbarton:
```suggestion
        key: ${{ env.CLING_HASH }}-${{ runner.os }}-${{ matrix.os }}-${{ matrix.compiler }}-clang-${{ matrix.clang-runtime }}-${{ matrix.root-llvm-tag || 'none' }}-${{ matrix.sanitizer || '' }}.x-patch-${{ hashFiles(format('patches/llvm/clang{0}-*.patch', matrix.clang-runtime)) || 'none' }}${{ matrix.oop-jit == 'On' && '-oop' || '' }}${{ matrix.sanitizer || '' }}
```
This should work in making the new cache, without disturbing anything else.

vgvassilev: Committed, thank you!

PR #900, opened 2026-04-20
Move DynamicLibraryManager from function-local static to per-interpreter member.

vgvassilev:
The DLM was a function-local static shared across all interpreter instances, causing search paths added in one interpreter to leak into all others. Move it to a mutable unique_ptr member of CppInternal::Interpreter so each instance owns its DLM and search path state is properly isolated.

Add a test that creates two interpreters, adds a search path to the first, and verifies the second cannot find a library through that path.


aaronj0 (approved):
LGTM, assuming the jobs pass

PR #901, opened 2026-04-20
Fix silencing of diagnostics in Declare() swallowing error status.

vgvassilev:
Before this patch, calling Declare() with silent=true suppressed diagnostic output but also prevented DiagnosticsEngine::ErrorOccurred from being set. This caused IncrementalParser::ParseOrWrapTopLevelDecl() to miss the error check, returning a broken PTU as if parsing succeeded. As a result, Declare() returned 0 (success) for invalid code, and the interpreter state could be left with partially parsed AST nodes.

This patch adds a DiagnosticErrorTrap around the silent declare() call. The trap checks TrapNumErrorsOccurred, which is incremented before the SuppressAllDiagnostics early return in ProcessDiag(), so it correctly detects errors even when output is suppressed. A new test exercises valid code, invalid code, and post-failure interpreter health under silent mode.

### DIFF
--- a/unittests/CppInterOp/InterpreterTest.cpp
+++ b/unittests/CppInterOp/InterpreterTest.cpp
@@ -177,6 +177,19 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Process) {
   clang_Interpreter_dispose(CXI);
 }
 
+TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_DeclareSilent) {
+  TestFixture::CreateInterpreter();
+
+  // Valid code with silent=true should succeed.
+  EXPECT_EQ(0, Cpp::Declare("int x = 42;", /*silent=*/true));
+
+  // Invalid code with silent=true should still report failure.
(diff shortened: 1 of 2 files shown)


Vipul-Cariappa (approved):
LGTM!

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -177,6 +177,19 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Process) {
   clang_Interpreter_dispose(CXI);
 }
 
+TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_DeclareSilent) {
```
mcbarton, line 180: @vgvassilev this test is crashing for my llvm 22 PR with no useful information that I can see (https://github.com/compiler-research/CppInterOp/actions/runs/24721430630/job/72311029306?pr=812#step:13:929 for the error message) . Any idea what is causing this?

vgvassilev: I am not sure what that is but I'd take care of the llvm22 upgrade soon...

PR #904, opened 2026-04-21
[test] Centralize Emscripten flag handling in unittest helper

vgvassilev:
Before this change, unittests/CppInterOp/CMakeLists.txt juggled a variable named EXTRA_TEST_SOURCE_FILES that meant two different things depending on the platform: main.cpp under Emscripten (appended to every test binary) and CUDATest.cpp on native (appended only to CppInterOpTests, then cleared mid-file). The per-target Emscripten compile/link option blocks were also duplicated three times with the same CPPINTEROP_COMMON_WASM_TEST_* inputs.

Move the common Emscripten setup into add_cppinterop_unittest:

- The helper injects main.cpp into the source list on Emscripten, since the Emscripten link set drops gtest_main and every test binary needs its own entry point.

- The helper applies CPPINTEROP_COMMON_WASM_TEST_COMPILE_FLAGS and CPPINTEROP_COMMON_WASM_TEST_LINK_FLAGS unconditionally for Emscripten test binaries.

With that in place, drop EXTRA_TEST_SOURCE_FILES and the three duplicate target_compile_options / target_link_options blocks. CUDATest.cpp is now attached explicitly via target_sources on the NOT EMSCRIPTEN branch, and the DynamicLibraryManagerTests preload-file becomes a single follow-up target_link_options call since the common flags are already in place.

While there, tighten up the (lengthy) Emscripten link-flag rationale comment for readability.

No behavior change: every test binary still receives the same sources, defines, compile options and link options as before, on both native and Emscripten.

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/unittests/CppInterOp/CMakeLists.txt
+++ b/unittests/CppInterOp/CMakeLists.txt
@@ -1,43 +1,35 @@
 if (EMSCRIPTEN)
-  # So we create a html file, as well as the javascript file
+  # Build a .html shell alongside the .js main module.
   set(CMAKE_EXECUTABLE_SUFFIX ".html")
-  # Omitting CUDATest.cpp since Emscripten build currently has no GPU support
-  # For Emscripten builds linking to gtest_main will not suffice for gtest to run
-  # the tests and an explicitly main.cpp is needed
-  set(EXTRA_TEST_SOURCE_FILES main.cpp)
 
   string(REPLACE "@" "@@" ESCAPED_SYSROOT_PATH "${SYSROOT_PATH}")
-  # Explanation of Emscripten-specific link flags for CppInterOpTests:
+  # Explanation of Emscripten-specific link flags for CppInterOp tests:
   #
   # MAIN_MODULE=1:
-  # Enables building CppInterOpTests.js as the main WebAssembly module, allowing dynamic linking of side modules.
+  # Enables the test binary as the main WebAssembly module, allowing dynamic
+  # linking of side modules (TestSharedLib).
   #
   # WASM_BIGINT:
-  # Ensures support for 64-bit integer types by enabling JavaScript BigInt integration in WASM.
+  # Enables JS BigInt integration in WASM so 64-bit integer types round-trip.
   #
   # ALLOW_MEMORY_GROWTH=1:
-  # Allows the WebAssembly memory to grow dynamically at runtime to accommodate increasing memory needs.
-  # Would lead to an abortOnCannotGrowMemory error if memory cannot be grown while running the tests.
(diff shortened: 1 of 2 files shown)


mcbarton (dismissed):
The rewording of the comments doesn't really seems to add anything, but the rest seems ok.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -1,43 +1,35 @@
...
+  # Preloads the system include directory into the Emscripten VFS so headers
+  # are available at runtime.
   #
-  # --emrun
-  # Makes it so that we run the html file created by this target, that we can capture the standard output
-  # and output to the terminal
-
-  # Common compile flags for WASM tests
+  # --emrun:
+  # Makes emrun drive the .html shell and pipe stdout back to the terminal.
```
mcbarton, line 29: I don't think this rewording of these comments `tightens up the (lengthy) Emscripten link-flag rationale comment for readability.`. The rewrite doesn't really add anything that I can see.

vgvassilev: I'd disagree. It makes is more readable, less verbose and follows the 80 column rules mostly.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -1,43 +1,35 @@
...
+  # linking of side modules (TestSharedLib).
   #
   # WASM_BIGINT:
-  # Ensures support for 64-bit integer types by enabling JavaScript BigInt integration in WASM.
+  # Enables JS BigInt integration in WASM so 64-bit integer types round-trip.
   #
   # ALLOW_MEMORY_GROWTH=1:
-  # Allows the WebAssembly memory to grow dynamically at runtime to accommodate increasing memory needs.
-  # Would lead to an abortOnCannotGrowMemory error if memory cannot be grown while running the tests.
+  # Lets the WebAssembly linear memory grow at runtime. Without it we hit
```
aaronj0, line 16: From my perspective, this looks much more straightforward while the previous sections seemed overly verbose and somewhat overexplained for what it is.

aaronj0 (approved):
LGTM

unittests/CppInterOp/CMakeLists.txt
mcbarton: I personally don't, but feel free to dismiss my review if you think otherwise.

vgvassilev: I'd claim that the current wording makes it more useful for non-native speakers at least ;)

PR #905, opened 2026-04-21
Unregister signal handlers for externally managed interpreters

aaronj0:
This crashes exception handling tests in ROOT. Ideally we should be able to configure if we want to exit the process, but for now we don't mandatory require it. Unblocks upgrading CppInterOp in ROOT, along with https://github.com/compiler-research/CppInterOp/pull/903

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -188,7 +188,8 @@ struct InterpreterInfo {
 
 static void DefaultProcessCrashHandler(void*);
 // Function-static storage for interpreters
-static std::deque<InterpreterInfo>& GetInterpreters() {
+static std::deque<InterpreterInfo>&
+GetInterpreters(bool SetCrashHandler = true) {
   // static int FakeArgc = 1;
   // static const std::string VersionStr = GetVersion();
   // static const char* ArgvBuffer[] = {VersionStr.c_str(), nullptr};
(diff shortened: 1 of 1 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -248,9 +248,10 @@ static void DefaultProcessCrashHandler(void*) {
   llvm::errs() << "**************************************************\n";
   llvm::errs().flush();
 
-  // The process must actually terminate for EXPECT_DEATH to pass.
-  // We use _exit to avoid calling atexit() handlers which might be corrupted.
-  llvm::sys::Process::Exit(/*RetCode=*/1, /*NoCleanup=*/false);
+  // TODO: Figure out a way to configure if we want the crash handler to exit.
+  // The process should actually terminate for EXPECT_DEATH to pass.
+  // Use _exit to avoid calling atexit() handlers which might be corrupted.
+  // llvm::sys::Process::Exit(/*RetCode=*/1, /*NoCleanup=*/false);
```
mcbarton: We should probably open up an issue about this. Maybe the issue can be about it being configurable

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -518,7 +520,12 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, SignalHandler_MultipleInterpreters) {
   ExpectedMsg = "";
 #endif //_WIN32
 
-  EXPECT_DEATH({ raise(SIGSEGV); }, ExpectedMsg);
+  EXPECT_DEATH(
```
vgvassilev: If it does not die, that's fine we can rework the tests accordingly.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -255,6 +255,8 @@ static void DefaultProcessCrashHandler(void*) {
 
 static void RegisterInterpreter(compat::Interpreter* I, bool Owned) {
   std::deque<InterpreterInfo>& Interps = GetInterpreters();
+  if (!Owned)
```
```suggestion
  // We let the externally provided interpreter deal with signal handling and teardown.
  if (!Owned)
```

vgvassilev (approved):
LGTM!

PR #906, opened 2026-04-21
Make all public API functions dispatchable from a single .td definition.

vgvassilev:
Default arguments were lost through the dispatch function pointer indirection, requiring ~245 DFLT_* macro workarounds across the test suite. 47 public functions were not dispatchable at all. Function signatures in the dispatch table could drift from CppInterOp.h because the two were maintained independently.


.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -177,6 +177,21 @@ runs:
...
+        mkdir -p native_cppinterop_build && pushd native_cppinterop_build
+        cmake -DCMAKE_BUILD_TYPE=Release \
+              -DLLVM_DIR=$NATIVE_LLVM_BUILD_DIR/lib/cmake/llvm \
+              -DCPPINTEROP_ENABLE_TESTING=OFF \
+              -DCPPINTEROP_BUILD_TABLEGEN_ONLY=ON \
+              ../
+        cmake --build . --target cppinterop-tblgen -j ${{ env.ncpus }}
+        export CPPINTEROP_TBLGEN_EXE=$PWD/bin/cppinterop-tblgen
+        echo "CPPINTEROP_TBLGEN_EXE=$CPPINTEROP_TBLGEN_EXE" >> $GITHUB_ENV
+        popd
```
mcbarton, line 194: Don't these changes require documentation changes, since we are changing the build process?

vgvassilev: If they work, we do need some wording around that we need tablegen to generate the header files. Not sure if that's something worth documenting and if we were to document it where would be that place?

vgvassilev: Added some comments in the docs.

.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -177,6 +177,22 @@ runs:
           CPLUS_INCLUDE_PATH="${LLVM_DIR}/llvm/include:${LLVM_DIR}/clang/include:${LLVM_BUILD_DIR}/include:${LLVM_BUILD_DIR}/tools/clang/include:$PWD/include"
         fi
 
+        # Build native cppinterop-tblgen for .inc file generation.
+        # Use the native LLVM build (not the Emscripten cross-build) since
+        # cppinterop-tblgen only needs libLLVMTableGen and libLLVMSupport.
+        NATIVE_LLVM_BUILD_DIR="$(pwd)/llvm-project/native_build"
```
mcbarton, line 183: I'm pretty sure here https://github.com/vgvassilev/CppInterOp/blob/2c3e3b4390349ee7e4058bd6a8df8d506a1a6018/.github/actions/Build_LLVM/action.yml#L148 and here https://github.com/vgvassilev/CppInterOp/blob/2c3e3b4390349ee7e4058bd6a8df8d506a1a6018/.github/actions/Build_LLVM/action.yml#L76 we delete the native_build directory from the cache to save space, so `$NATIVE_LLVM_BUILD_DIR/lib/cmake/llvm` probably doesn't exist

vgvassilev: I can try reverting this change but without it emscripten failed. I feel in general that our organically grew to a untenable proportions and I will also revisit these workflows as we've been intending to since long time. I do not think it makes sense to go and debug each single line at that point since my plan is probably going to rework a lot of that anyway.

mcbarton: What I mean is that I don't understand why referencing `$NATIVE_LLVM_BUILD_DIR/lib/cmake/llvm` would help get the Emscripten tests to pass in the first place, since the folder shouldn't exist in our llvm cache for any Emscripten job.

vgvassilev: Ah, ok, yes, because we started depending on tablegen which is an executable that gets called during build to produce the list of functions (in the .inc file). In that case we need to cross compile that binary iirc so that emscripten works because it cannot link to libLLVMTablegen.a or similar...

mcbarton: I assume the cross compiled executable tablegen might already exist in the `build` folder. Does it not? It not, I will investigate, and see how we can get what you need.

vgvassilev: This is a CppInterOp binary not an llvm binary. We need to build it but who knows...

.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -177,6 +177,22 @@ runs:
...
+        mkdir -p native_cppinterop_build && pushd native_cppinterop_build
+        cmake -DCMAKE_BUILD_TYPE=Release \
+              -DLLVM_DIR=$NATIVE_LLVM_BUILD_DIR/lib/cmake/llvm \
+              -DCPPINTEROP_ENABLE_TESTING=OFF \
+              -DCPPINTEROP_BUILD_TABLEGEN_ONLY=ON \
+              ../
+        cmake --build . --target cppinterop-tblgen -j ${{ env.ncpus }}
```
mcbarton, line 190: If this need this cross compiled to Emscripten then this is what I suggest
```suggestion
        ecmake ecmake -DCMAKE_BUILD_TYPE=Release \
              -DLLVM_DIR=$LLVM_BUILD_DIR/lib/cmake/llvm       \
              -DLLD_DIR=$LLVM_BUILD_DIR/lib/cmake/lld         \
              -DClang_DIR=$LLVM_BUILD_DIR/lib/cmake/clang     \
              -DCPPINTEROP_ENABLE_TESTING=OFF \
              -DCPPINTEROP_BUILD_TABLEGEN_ONLY=ON \
              ../
        emmake make cppinterop-tblgen -j ${{ env.ncpus }}
```

.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -461,6 +481,21 @@ runs:
...
+        # Build native cppinterop-tblgen for .inc file generation.
+        mkdir native_cppinterop_build
+        pushd native_cppinterop_build
+        $env:NATIVE_LLVM_BUILD_DIR="$env:PWD_DIR\llvm-project\native_build"
+        cmake -DCMAKE_BUILD_TYPE=Release `
+              -DLLVM_DIR="$env:NATIVE_LLVM_BUILD_DIR\lib\cmake\llvm" `
+              -DCPPINTEROP_ENABLE_TESTING=OFF `
+              -DCPPINTEROP_BUILD_TABLEGEN_ONLY=ON `
+              ..\
+        cmake --build . --target cppinterop-tblgen -j ${{ env.ncpus }}
```
mcbarton, line 493: Same change but for Windows
```suggestion
        emcmake cmake -DCMAKE_BUILD_TYPE=Release `
              -DLLVM_DIR="$env:LLVM_BUILD_DIR\lib\cmake\llvm"        `
              -DLLD_DIR="$env:LLVM_BUILD_DIR\lib\cmake\lld"        `
              -DClang_DIR="$env:LLVM_BUILD_DIR\lib\cmake\clang"    `
              -DLLVM_DIR="$env:NATIVE_LLVM_BUILD_DIR\lib\cmake\llvm" `
              -DCPPINTEROP_ENABLE_TESTING=OFF `
              -DCPPINTEROP_BUILD_TABLEGEN_ONLY=ON `
              ..\
        emmake make cppinterop-tblgen -j ${{ env.ncpus }}
```

.github/actions/Build_and_Test_CppInterOp/action.yml
```diff
@@ -461,6 +481,21 @@ runs:
           echo "CPLUS_INCLUDE_PATH=$env:CPLUS_INCLUDE_PATH" >> $env:GITHUB_ENV
         }
 
+        # Build native cppinterop-tblgen for .inc file generation.
+        mkdir native_cppinterop_build
+        pushd native_cppinterop_build
+        $env:NATIVE_LLVM_BUILD_DIR="$env:PWD_DIR\llvm-project\native_build"
```
mcbarton, line 487:
```suggestion
```

.github/actions/Build_and_Test_CppInterOp/action.yml
mcbarton, line 183:
```suggestion
```

mcbarton (commented):
This is what I think your suggesting, but I may be completly wrong.

mcbarton:
If my suggestions don't work, then I will investigate locally, to see if I can get this to work. 

On the llvm side I think to get the Emscripten tblgens, we first build the native tblgens, and then use that in the creation of the cross compiled one. No idea if it would be the same for CppInterOp though without trying it out first.

vgvassilev:
> If my suggestions don't work, then I will investigate locally, to see if I can get this to work.
> 
> On the llvm side I think to get the Emscripten tblgens, we first build the native tblgens, and then use that in the creation of the cross compiled one. No idea if it would be the same for CppInterOp though without trying it out first.

Right now (before the rebase everything worked). I prefer that to not be blocked and I am happy to get in as a separate improvement. Again the CI will be majorly reworked in the next days.

aaronj0 (approved):
LGTM!

PR #908, opened 2026-04-22
Update self hosted job to llvm 22 (updates cuda job to llvm 22)

mcbarton:


mcbarton:
I have no idea why the coverage ci has a failure, saying it dropped 0.49%, but the self hosted ci job sucessfully passes updating to llvm 22.

vgvassilev (approved):
LGTM!

PR #910, opened 2026-04-22
[ci] Move asan to llvm22

vgvassilev:


mcbarton:
If you see here https://github.com/vgvassilev/CppInterOp/blob/5de10abb671f2513436e183e3b251be320e4d6a7/.github/actions/Build_and_Test_CppInterOp/action.yml#L67 we currently only run the CppInterOpTests tests under Valgrind. Given this PR is planning to tackle the many Valgrind issues CppInterOp has, it might be worth taking this opportunity to run the other test suites we have under Valgrind.

.github/workflows/main.yml
```diff
@@ -67,22 +67,13 @@ jobs:
...
         include:
           # Ubuntu Arm Jobs
-          - name: ubu24-arm-gcc12-clang-repl-22
-            os: ubuntu-24.04-arm
-            compiler: gcc-12
-            clang-runtime: '22'
-            llvm_enable_projects: "clang"
-            llvm_targets_to_build: "host;NVPTX"
-            Valgrind: On
-            python-version: '3.14'
```
mcbarton, line 77: Removing this job will mean we no longer run the test under Valgrind on Ubuntu arm, which has shown to have different issues compared to Ubuntu x86 where you are trying to consolidate everything.

.github/workflows/main.yml
```diff
@@ -104,16 +95,18 @@ jobs:
             llvm_targets_to_build: "host;NVPTX"
             python-version: '3.14'
           # Ubuntu X86 Jobs
-          - name: ubu24-x86-gcc12-clang-repl-22
+          - name: ubu24-x86-gcc12-clang-repl-22-vg
```
mcbarton: Maybe we can have this job on the self hosted runner so we can run the cuda tests under Valgrind too?

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -2759,8 +2735,6 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_DestructArray) {
 #ifdef EMSCRIPTEN
   GTEST_SKIP() << "Test fails for Emscipten builds";
 #endif
-  if (llvm::sys::RunningOnValgrind())
-    GTEST_SKIP() << "XFAIL due to Valgrind report";
```
mcbarton, line 2749: You may notice that quite a few jobs that are skipped on Valgrind, are labelled as failed on Emscripten. Emscripten is quite sensitive to memory issues (at least that has been my experience), so its quite possible that when you have fixed all the Valgrind issues, you may have fixed some Emscripten tests.

.github/workflows/main.yml
vgvassilev: Yes, and that is intended.

.github/workflows/main.yml
vgvassilev: Yes, we have that for clad but it is done differently as there is some memchecker for cuda -- once my planned changes to the ci go through we could add it.

PR #785, opened 2026-01-22
Drop llvm 18 support

mcbarton:
This PR will drop llvm 18 support from CppInterOp.

PR #911, opened 2026-04-22
[ci] Improve the developers experience without losing too much of the ci coverage.

vgvassilev:
Each OS+arch had up to three cppyy matrix rows — one per clang-runtime (20, 21, 22). cppyy exercises the same public CppInterOp API on every clang-runtime, so the clang-20 cppyy rows were pure duplication once the clang-21 rows existed.

Cache keys do not include cppyy (by design, cppyy tests share the LLVM build with the equivalent non-cppyy row), but none of the dropped rows shared a cache key with any remaining job: the Linux arm+gcc-12+clang-20 plain cache, the Linux x86+gcc-12+clang-20 plain cache, and the two macOS clang-20 plain caches are unique to the dropped rows. So four clang+LLVM builds come off the critical path.

Drop:
- ubu24-arm-gcc12-clang-repl-20-cppyy
- ubu24-x86-gcc12-clang-repl-20-cppyy
- osx26-arm-clang-clang-repl-20-cppyy
- osx26-x86-clang-clang-repl-20-cppyy

To preserve cppyy coverage on the latest supported LLVM, bump ubu24-x86-gcc12-clang-repl-21-cppyy to clang-runtime 22; the remaining clang-21 cppyy coverage stays on Linux ARM plus both macOS arches. clang-20 stays exercised via the oop-jit and cling rows.

Matrix: 23 -> 19 jobs.

ARM cling was exercised on three platforms: Ubuntu 24.04 ARM, macOS 26 ARM, and Windows 11 ARM. Cling behavior on ARM is platform-agnostic for the CppInterOp API surface these tests cover; keeping three full LLVM+cling builds is over-coverage.

Keep the macOS 26 ARM cling job as the ARM cling representative: it is the freshest ARM OS in the matrix and it bundles the cppyy test surface as well, so losing it would cost more than losing either of the others. Drop ubu24-arm-gcc14-clang20-cling-cppyy and win11-msvc-clang20-cling. x86 cling coverage is unchanged (Linux x86, macOS x86, Windows x86).

Also drop osx26-x86-clang-clang-repl-22, the Intel-Mac plain smoke test. Intel Macs are phasing out of the supported matrix and this job was the only macOS-x86 entry on clang-22; macOS-ARM already carries the clang-22 plain signal plus the cppyy and oop-jit axes.

Matrix: 19 -> 16 jobs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,26 +75,6 @@ jobs:
             llvm_targets_to_build: "host;NVPTX"
             Valgrind: On
             python-version: '3.14'
-          - name: ubu24-arm-gcc12-clang-repl-21-cppyy
-            os: ubuntu-24.04-arm
-            compiler: gcc-12
-            clang-runtime: '21'
-            cppyy: On
-            llvm_enable_projects: "clang"
-            llvm_targets_to_build: "host;NVPTX"
-            Valgrind: On
-            python-version: '3.14'
-          - name: ubu24-arm-gcc14-clang20-cling-cppyy
-            os: ubuntu-24.04-arm
-            compiler: gcc-14
-            clang-runtime: '20'
-            cling: On
-            cppyy: On
-            cling-version: '1.3'
-            root-llvm-tag: 'cling-llvm20-20260119-01'
-            llvm_enable_projects: "clang"
-            llvm_targets_to_build: "host;NVPTX"
-            python-version: '3.14'
           # Ubuntu X86 Jobs
           - name: ubu24-x86-gcc12-clang-repl-22
             os: ubuntu-24.04
@@ -102,6 +82,15 @@ jobs:
             clang-runtime: '22'
             llvm_enable_projects: "clang"
             llvm_targets_to_build: "host;NVPTX"
+          - name: ubu24-x86-gcc12-clang-repl-21-cppyy
+            os: ubuntu-24.04
+            compiler: gcc-12
+            clang-runtime: '21'
+            cppyy: On
+            llvm_enable_projects: "clang"
+            llvm_targets_to_build: "host;NVPTX"
+            Valgrind: On
+            python-version: '3.14'
           - name: ubu24-x86-gcc12-clang-repl-21-asan
             os: ubuntu-24.04
             compiler: gcc-12
@@ -185,13 +174,6 @@ jobs:
             llvm_targets_to_build: "host;NVPTX"
             python-version: '3.14'
           # MacOS X86 Jobs
-          - name: osx26-x86-clang-clang-repl-22
-            os: macos-26-intel
-            compiler: clang
-            clang-runtime: '22'
-            llvm_enable_projects: "clang"
-            llvm_targets_to_build: "host"
-            python-version: '3.14'
           - name: osx26-x86-clang-clang-repl-21-cppyy
             os: macos-26-intel
             compiler: clang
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -249,16 +201,6 @@ jobs:
...
-          - name: win11-msvc-clang20-cling
-            os: windows-11-arm
-            compiler: msvc
-            clang-runtime: '20'
-            cling: On
-            cling-version: '1.3'
-            root-llvm-tag: 'cling-llvm20-20260119-01'
-            llvm_enable_projects: "clang"
-            llvm_targets_to_build: "host;NVPTX"
-            python-version: '3.14'
```
mcbarton, line 261: Why is Windows arm using cling expected to be the same as Windows x86 using cling?

mcbarton: We actually have tests for CppInterOp we pass for Windows x86 but not arm (see https://github.com/compiler-research/CppInterOp/pull/785#issuecomment-3840678541 for one such example)

.github/workflows/main.yml
```diff
@@ -84,25 +84,6 @@ jobs:
...
-            os: ubuntu-24.04-arm
-            compiler: gcc-14
-            clang-runtime: '20'
-            cling: On
-            cppyy: On
-            cling-version: '1.3'
-            root-llvm-tag: 'cling-llvm20-20260119-01'
-            llvm_enable_projects: "clang"
-            llvm_targets_to_build: "host;NVPTX"
-            python-version: '3.14'
```
mcbarton, line 105: Why are we expected Ubuntu x86 and Ubuntu arm to be the same for the cppyy part? We have differing number of passing tests for Ubuntu arm vs x86 for cppyy.

mcbarton (dismissed):
I am strongly against this PR in its current form. If you want to provide a small speedup to the ci then drop the cpppy option for jobs which don't test the latest llvm, so that CppIntrerOp continues to be tested for multiple llvm versions and platforms, checking we have backwards compatibility, and cross platform support.

A PR like this is better placed for the CPyCppyy, cppyy-backend and cppyy repo cis.

.github/workflows/main.yml
vgvassilev: Cling on arm has no user base and our team is not going to support it.

.github/workflows/main.yml
vgvassilev: The same comment holds here -- arm is a non-relevant platform for cling outside of osx. If anything happens there it's not our responsibility to deal with this. Right now we pay a huge development costs in terms of ci to handle these platforms and if something fails as a test we just skip it as something we will likely never fix.

vgvassilev:
> I am strongly against this PR in its current form. If you want to provide a small speedup to the ci then drop the cpppy option for jobs which don't test the latest llvm, so that CppIntrerOp continues to be tested for multiple llvm versions and platforms, checking we have backwards compatibility, and cross platform support.
> 
> A PR like this is better placed for the CPyCppyy, cppyy-backend and cppyy repo cis.

I disagree because it's not the responsibility of the downstream tool to run extensive checks against clients. Imagine if llvm/linux kernel decided to do this? We still have quite reasonable guardrails if we break cppyy -- just not needed to happen on all possible free nodes we have access to. One thing to consider is the overall minutes our ogranization spends on ci and that's something we must be reducing if not for money at least for co2 footprint.

aaronj0:
Is dropping Linux x86 llvm21 cppyy testing intentional? We have the job for ubuntu-arm instead which based on the comments was not the intention. Or maybe cppyy is included in the asan build?

vgvassilev:
> Is dropping Linux x86 llvm21 cppyy testing intentional? We have the job for ubuntu-arm instead which based on the comments was not the intention. Or maybe cppyy is included in the asan build?

If you mean `ubu24-x86-gcc12-clang-repl-21-cppyy` then yes, that moved as 22. It's not worth testing all possible llvm configurations for cppyy.

mcbarton:
I think its fair to say that any review comments I have for ci changes in the next few days is likely to be ignored, and the changes will be pushed through anyway. Therefore I ask that I am not be added as a reviewer on any ci related PRs while your reworking it, since I don't want to waste my time on reviews.

vgvassilev:
> I think its fair to say that any review comments I have for ci changes in the next few days is likely to be ignored, and the changes will be pushed through anyway. Therefore I ask that I am not be added as a reviewer on any ci related PRs while your reworking it, since I don't want to waste my time on reviews.

Apologies -- this is a dev sprint before the integration in ROOT -- will ping you again when we are done.

aaronj0 (approved):
LGTM, Thanks!

PR #913, opened 2026-04-22
Add various type reflection interfaces for cppyy/numba

aaronj0:
Revived https://github.com/compiler-research/CppInterOp/pull/534.
This was earlier motivated by arg match scoring interfaces for numba, and recently a ROOT commit aliasing `size()` to `__len__` for container-like classes: https://github.com/root-project/root/commit/cccbacc64c431ecbd4ea113f247c9aee9145a672, so I opened a fresh PR and also addressed the last comment on making signedness an out parameter

### DIFF
--- a/include/CppInterOp/CppInterOp.h
+++ b/include/CppInterOp/CppInterOp.h
@@ -95,12 +95,14 @@ enum Operator : unsigned char {
 };
 
 enum OperatorArity : unsigned char { kUnary = 1, kBinary, kBoth };
+enum Signedness : unsigned char { kSigned = 1, kUnsigned };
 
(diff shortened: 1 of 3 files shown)


aaronj0:
Seeing we now have `RemoveTypeQualifier` I will drop the implementation of `GetUnqualifiedType` as it is no longer required. However I have extended it with an `All` enum so that would mimic a call to `GetUnqualifiedType`  instead of making a verbose call on the client-side like `Cpp::RemoveTypeQualifier(a, Cpp::QualKind::Const | Cpp::QualKind::Volatile |Cpp::QualKind::Restrict )`

vgvassilev (approved):
LGTM!

PR #916, opened 2026-04-23
[jitcall] Do not require <utility> for std::move.

vgvassilev:
This patch uses the underlying implementation of std::move instead.


Vipul-Cariappa (approved):
LGTM!

PR #918, opened 2026-04-23
Replace duplicated dispatch test suite with focused smoke tests.

vgvassilev:
The dispatch variant re-compiled all 174 tests through dlopen function pointers, doubling CI time without proportional coverage. The inline wrappers are trivial forwards — if one works, they all work.

Replace with 15 targeted tests in a standalone DispatchTests binary: 4 mechanism tests (CppGetProcAddress lookup, missing symbols, load/unload cycle) and 11 smoke tests covering representative API categories.

The ENABLE_DISPATCH_TESTS conditional compilation is removed. Dispatch storage definitions and initialization move to DispatchInit.cpp, linked only by DispatchTests. Utils.h no longer branches on dispatch mode.


lib/CppInterOp/Dispatch.cpp
```diff
@@ -1,10 +1,11 @@
-#include <CppInterOp/Dispatch.h>
+#include <CppInterOp/CppInterOp.h>
 
 #include <iostream>
 #include <string_view>
 #include <unordered_map>
 
 using namespace CppImpl;
+using CppFnPtrTy = void (*)();
```
aaronj0, line 13: Why is this redefined here? Already present in Dispatch.h

unittests/CppInterOp/DispatchTest.cpp
```diff
@@ -1,87 +1,49 @@
...
-            std::string::npos);
+TEST(DispatchTest, DlGetProcAddress_Unimplemented) {
+  // dlGetProcAddress returns nullptr for unknown symbols (no stderr output
+  // since it goes through the local dispatch, not CppGetProcAddress).
+  void* ptr = dlGetProcAddress("NonExistentFunction");
+  EXPECT_EQ(ptr, nullptr);
 }
 
-TEST(CPPINTEROP_TEST_MODE, DispatchAPI_LoadUnloadCycle) {
-  Cpp::UnloadDispatchAPI(); // make sure we're no loaded already...
```
aaronj0, line 74: The comments here seemed useful..

unittests/CppInterOp/DispatchTest.cpp
```diff
@@ -1,87 +1,49 @@
...
+TEST(DispatchTest, DlGetProcAddress_Basic) {
+  Cpp::CreateInterpreter({});
+  Cpp::Declare("namespace N {} class C{}; int I;");
+
   using IsClassFn_t = bool (*)(Cpp::TCppScope_t);
-  auto IsClassFn = reinterpret_cast<IsClassFn_t>(CppGetProcAddress("IsClass"));
+  auto IsClassFn = reinterpret_cast<IsClassFn_t>(dlGetProcAddress("IsClass"));
   EXPECT_NE(IsClassFn, nullptr) << "Failed to obtain API function pointer";
-  std::vector<Decl*> Decls;
-  GetAllTopLevelDecls("namespace N {} class C{}; int I;", Decls);
```
aaronj0, line 23: Why do we want to rely on GetNamed here for this unittest? I believe using the Decls directly keeps the test more self-contained.

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -0,0 +1,323 @@
...
+  void InvokeDestructor(void* object, unsigned long nary = 0,
+                        int withFree = true) const {
+    assert(m_Kind == kDestructorCall && "Wrong overload!");
+#ifndef NDEBUG
+    ReportInvokeStart(object, nary, withFree);
+#endif
+    m_DestructorCall(object, nary, withFree);
+  }
+
+  void InvokeConstructor(void* result, unsigned long nary = 1,
```
aaronj0, line 316: We lost documentation here:

```
  /// Makes a call to a constructor.
  ///\param[in] result - the memory address at which we construct the object
  ///           (placement new).
  ///\param[in] nary - Use array new if we have to construct an array of
  ///           objects (nary > 1).
  ///\param[in] args - a pointer to a argument list and argument size.
  ///\param[in] is_arena - a pointer that indicates if placement new is to be
  /// used
  // FIXME: Change the type of withFree from int to bool in the wrapper code.
```

 I believe the Claude rewrites are updating things that should not have changed... Can you recheck the diffs?

vgvassilev: Good catch.

unittests/CppInterOp/DispatchTest.cpp
vgvassilev: GetAllTopLevelDecls infrastructure requires Utils which requires explicit linking.

unittests/CppInterOp/DispatchTest.cpp
vgvassilev: Not to me, but will bring it back.

lib/CppInterOp/Dispatch.cpp
vgvassilev: We don't include dispatch however, I think I have a better fix now.

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -102,3 +101,42 @@ endif()
...
+    LIBRARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/bin/$<CONFIG>/
+  )
+  add_test(NAME cppinterop-DispatchTests COMMAND DispatchTests)
+  set_tests_properties(cppinterop-DispatchTests PROPERTIES
+    TIMEOUT "${TIMEOUT_VALUE}"
+    ENVIRONMENT "CPLUS_INCLUDE_PATH=${CMAKE_BINARY_DIR}/etc")
+  # Verify the binary is NOT linked against libclangCppInterOp. If this
+  # check fails, the dispatch tests are not truly testing RTLD_LOCAL.
+  if(UNIX AND NOT APPLE)
+    add_test(NAME cppinterop-DispatchTests-no-direct-link
```
aaronj0: do we need the `-no-direct-link` suffix?

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -102,3 +101,42 @@ endif()
 
 add_subdirectory(TestSharedLib)
 add_dependencies(DynamicLibraryManagerTests TestSharedLib)
+
+# Dispatch smoke tests: verify the dlopen dispatch wrappers forward correctly.
```
aaronj0:
```suggestion
# Dispatch Tests:
```

Since this includes more than just the smoke test (i.e this is repeated in the documentation of `DispatchSmokeTest.cpp`)

unittests/CppInterOp/DispatchTest.cpp
```diff
@@ -1,87 +1,50 @@
...
+  EXPECT_FALSE(IsClassFn(var));
+}
 
-  EXPECT_NE(demangled_add_int.find(GetQualifiedCompleteNameFn(Decls[0])),
-            std::string::npos);
-  EXPECT_NE(demangled_add_double.find(GetQualifiedCompleteNameFn(Decls[1])),
-            std::string::npos);
+TEST(DispatchTest, DlGetProcAddress_Unimplemented) {
+  // dlGetProcAddress returns nullptr for unknown symbols (no stderr output
+  // since it goes through the local dispatch, not CppGetProcAddress).
```
aaronj0: Wouldn't this still go through `CppGetProcAddress`? In this case we find the shared lib and dlsym CppGetProcAddress would give us the address. That would then return nullptr

unittests/CppInterOp/CMakeLists.txt
vgvassilev: What’s a better name? I think the Dispatch-Tests is already taken.

aaronj0 (approved):
Apart from a few minor comments above, this looks good to me! Thanks for this great work

unittests/CppInterOp/CMakeLists.txt
aaronj0: `cppinterop-DispatchTests`?

unittests/CppInterOp/DispatchSmokeTest.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+TEST(DispatchSmokeTest, EnumReflection) {
+  Cpp::CreateInterpreter({});
+  Cpp::Declare("enum DispEnum { A, B, C };");
+
+  auto* e = Cpp::GetNamed("DispEnum");
+  EXPECT_TRUE(Cpp::IsEnumScope(e));
+
+  auto constants = Cpp::GetEnumConstants(e);
+  EXPECT_EQ(constants.size(), 3u);
```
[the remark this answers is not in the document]
aaronj0: Perhaps a useful suggestion

unittests/CppInterOp/CMakeLists.txt
vgvassilev: It's already taken.

unittests/CppInterOp/DispatchSmokeTest.cpp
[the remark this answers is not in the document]
vgvassilev: Should be resolved now.

PR #919, opened 2026-04-23
Partially fix Emscripten ci on main (fixes osx job which fixes deployment, but Windows still broken)

mcbarton:
This PR will partially fix the broken Emscripten ci on main. If the Emscripten cache is cleared before this is merged into main, the osx jobs will be fixed, and the Windows jobs will be partially fixed. To fix the Windows job we need to work out how to make these 29 unresolved symbols https://github.com/mcbarton/CppInterOp/actions/runs/24837217861/job/72700808805#step:16:304  available to the cppinterop-tblgen executable. I couldn't work out how to do it.


vgvassilev (approved):
LGTM!

PR #924, opened 2026-04-24
[jit] Move out-of-process JIT cells from clang-20 to clang-22

vgvassilev:
The OOP JIT configuration previously lived in a local clang-20 patch (patches/llvm/clang20-1-out-of-process.patch) which pinned the matrix cells to clang-20. Upstream release/22.x carries the equivalent surface as clang::IncrementalExecutorBuilder; port lib/CppInterOp/Compatibility.h to use it and promote the two OOP cells to clang-22 / python-3.14 so they match the other cells on compiler and Python version.

The builder's fields (IsOutOfProcess, OOPExecutor, CustomizeFork, ...) line up with the old JITConfig. Register the upstream UpdateOrcRuntimePathCB via IncrementalCompilerBuilder, so the orc_rt archive path is resolved from the driver's toolchain during CreateCpp() — no more per-target path hardcoded in CppInterOp. Relax the CMakeLists OOP-JIT guard to accept Linux aarch64 accordingly.

Drop GetExecutorPID() from the public C API and the internal getOutOfProcessExecutorPID() helper: upstream LLVM-22 exposes no executor-PID accessor, the C API had no in-tree callers, and keeping a stub that silently returned 0 would be a trap.

The upstream write(buf) uninit-byte report from
FDSimpleRemoteEPCTransport::sendMessage persists in 22.x; replace etc/clang20-valgrind.supp with etc/clang22-valgrind.supp that anchors on the offending function and uses a "..." tail (the old supp pinned a specific libstdc++ .so path that no longer matches noble). The new supp carries a note that no llvm/llvm-project issue is filed yet.

Verified on aarch64 / Ubuntu 24.04 against release/22.x: CppInterOpTests runs 349 tests (in-process + OOP-JIT variants) with 320 PASSED and 29 SKIPPED without valgrind. Under valgrind with the new suppression the same suite returns exit 0.


CMakeLists.txt
```diff
@@ -344,10 +344,10 @@ add_definitions(-DLLVM_BINARY_LIB_DIR="${LLVM_BINARY_LIB_DIR}")
 
 if(LLVM_BUILT_WITH_OOP_JIT)
   if((CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_SYSTEM_PROCESSOR MATCHES "arm64") OR
-     (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64"))
+     (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|aarch64|arm64"))
```
mcbarton, line 347: Just curious as to when we got oop support for Linux arm? Only asking since it wasn't supported for llvm 20.

vgvassilev: We will need to check but I guess my assumption is that if it works for arm on osx it works on arm for Unix. There is nothing specific to osx we are using.

mcbarton: The PR description in the last paragraph already says `Verified on aarch64 / Ubuntu 24.04` so I would have assumed it was already tested on Ubuntu arm. Maybe these PR descriptions need checking before any merges to check they are accurate since they get put as the commit description.

vgvassilev: Yes, I have a docker arm container indeed.

PR #926, opened 2026-04-25
[ci] Lift the PR-opt-in gate in emscripten.yml to the job level.

vgvassilev:
Every step in the emscripten build job carried the same conditional,

  if: !(github.event_name == 'pull_request' && matrix.run-in-prs != true)

which only the ubu24-arm row opts into. Step-level gating still consumed a runner slot per matrix row on every PR just to skip each step in turn. Lift the gate onto the job itself so the three rows without run-in-prs are never scheduled on PRs and their runners stay free for other jobs.

Matrix context is valid in job-level `if:` for strategy-expanded jobs. Steps that combine the gate with `runner.os` or `steps.cache.outputs.cache-hit` checks keep only those remaining conditions.

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -25,6 +25,8 @@ jobs:
   build:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


vgvassilev:
we lost all wasm jobs...

PR #930, opened 2026-04-25
Delete .github/ISSUE_TEMPLATE directory

mcbarton:
These issue templates are never really updated (shown by virtue of the fact the llvm you built against option includes cling 1.0 and llvm 19). Therefore this PR removes them, and we will go back to just have an issue description box to fill in.


vgvassilev:
These are useful for new contributors outside of the core team.

PR #931, opened 2026-04-25
[ci] Deduplicate the wasm LLVM cmake invocation; switch off Windows

vgvassilev:
The wasm cross-compile of LLVM/Clang/lld set
`-DLLVM_ENABLE_LTO=Full`, which leaves the resulting `.a` archives as full-LTO bitcode rather than emitting wasm objects. The 18-minute black hole between `[51%] Linking CXX shared library libclangCppInterOp.so` and the next test object on Windows is the emcc/wasm-ld whole-program optimization pass over that bitcode -- paid on every cache-hit run, not just on cache-miss, because the cached archives still contain the bitcode. PR runs are therefore not faster than scheduled runs in practice.

Switch to `LTO=Thin`. Thin LTO link is parallel and several times faster (the codegen happens per-translation-unit summary fragments, not over a single monolithic IR), while keeping most of the size benefit Full provides for the deployed Jupyter Lite artifact. On the Windows runner this should cut the 18 min link to roughly 3-4 min.

Because every event now produces the same (Thin-LTO) artifact, the schedule-only cache-save gate added previously is dropped: the ordinary "save when cache miss and the row is not gated out" test returns.

### DIFF
--- a/.github/actions/Build_LLVM_WASM/action.yml
+++ b/.github/actions/Build_LLVM_WASM/action.yml
@@ -13,104 +13,82 @@ runs:
       run: |
         ./emsdk/emsdk activate ${{ env.EMSDK_VER }}
         source ./emsdk/emsdk_env.sh
+
         cling_on=$(echo "${{ matrix.cling }}" | tr '[:lower:]' '[:upper:]')
         if [[ "${cling_on}" == "ON" ]]; then
           git clone https://github.com/root-project/cling.git
-          cd ./cling
-          git checkout tags/v${{ matrix.cling-version }}
-          git apply -v ../patches/llvm/cling1.2-LookupHelper.patch
-          cd ..
+          (cd cling && git checkout tags/v${{ matrix.cling-version }} \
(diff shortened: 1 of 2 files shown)


.github/actions/Build_LLVM_WASM/action.yml
```diff
@@ -59,7 +59,7 @@ runs:
                         -DLLVM_NATIVE_TOOL_DIR=$NATIVE_DIR 		\
                         -DCMAKE_C_FLAGS_RELEASE="-Oz -g0 -DNDEBUG" \
                         -DCMAKE_CXX_FLAGS_RELEASE="-Oz -g0 -DNDEBUG" \
-                        -DLLVM_ENABLE_LTO=Full \
+                        -DLLVM_ENABLE_LTO=Thin \
```
mcbarton: This not being tested as part of the PR, since it uses the Emscripten llvm on main. You need to delete the Emscripten cache on main and rerun the ci to test this works and doesn't trip our size guard rail.

vgvassilev: That part of the commit is gone.

PR #933, opened 2026-04-25
Fix to https://github.com/compiler-research/CppInterOp/pull/931

mcbarton:
https://github.com/compiler-research/CppInterOp/pull/931 didn't keep the change which made sure the native_build directory was kept. This wasn't caught since the Emscripten build on main was not deleted, so the new action was never actually tested. I noticed this was missing in my Emscripten llvm 22 PR, since I could no longer build the cppInterop-tblgen, since the native_build directory no longer existed.


vgvassilev (approved):
LGTM!

PR #934, opened 2026-04-25
Fix Windows Emscripten build instructions to use Ninja generator when building cppinterop-tblgen

mcbarton:
https://github.com/compiler-research/CppInterOp/pull/928/changes didn't include the documentation changes so that users locally could replicate the build.

### DIFF
--- a/Emscripten-build-instructions.md
+++ b/Emscripten-build-instructions.md
@@ -211,7 +211,7 @@ On Windows:
 ```powershell
 mkdir native_cppinterop_build
 cd native_cppinterop_build
-cmake -DCMAKE_BUILD_TYPE=Release `
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Release `
       -DLLVM_DIR="$env:NATIVE_LLVM_BUILD_DIR\lib\cmake\llvm" `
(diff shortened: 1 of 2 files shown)


vgvassilev:
The build instructions are too volatile and too often buggy or out of date. I'd propose to point to the github actions links so that people get the real picture. Now this should be sufficiently cleaned up for users to look at.

mcbarton:
> The build instructions are too volatile and too often buggy or out of date. I'd propose to point to the github actions links so that people get the real picture. Now this should be sufficiently cleaned up for users to look at.

I personally don't think the documentation to often buggy, and I don't know where this assertion has come from. I certaintly can't stop you pointing the user to the github ci for them to reverse engineer the build instructions, but I certaintly don't think its a good idea, since this will hard for beginner looking to contribute (I say this as someone who had to do that when I first contributed to this project given the documentation was very incomplete then). Its been very disheartening for me recently to see the route this project has been taking whatever Gemini, Claude, etc spits out as gospel and more clear. I find it very demotivating feeling that any contributions I make to this project in the future will likely very soon not receive a human review. Apologies for the way this comment may come accross, but I thought it best for myself if I let my feelings be known.

PR #935, opened 2026-04-25
Move CXCppInterOp.h to CppInterOp folder

mcbarton:
As we are removing the `clang/Interpreter/CppInterOp.h` as part of CppInterOp v2.0, it didn't make sense to me to have a `clang-c` folder, so this PR moves the `CXCppInterOp.h` header file under the `CppInterOp` folder


vgvassilev:
Hi @mcbarton, as I've mentioned -- we are in a sprint and moving quickly. This should be handled by https://github.com/compiler-research/CppInterOp/pull/921

PR #936, opened 2026-04-25
[ci] Make the LLVM/Cling cache leaner and the key readable.

vgvassilev:
Five tied-together hygiene fixes on the cache and Build_LLVM action. None of them is individually large enough to ship alone; together they shrink the cached artifact, make the key string greppable, fix a latent WASM trim bug, and unblock the parallel ROOT-integration work which needs `clangInterpreter` and the source-tree `lib/` drop.

Source-tree shrink. `Build_LLVM` and `Build_LLVM_WASM` were keeping the source `lib/` dirs of `llvm-project/{llvm,clang}` in the cache even though nothing on the consumer include path references them (`Build_and_Test_CppInterOp/action.yml` only adds `llvm/include` and `clang/include` to `CPLUS_INCLUDE_PATH`) and ROOT with `builtin_*=OFF` links the pre-built `.a` files instead of recompiling. Drop them from the keep-lists. Estimated savings ~50-100 MB compressed per cache row, ~1-1.5 GB across the pool.

Build a leaner LLVM. The cling cmake call gains
`-DLLVM_INCLUDE_BENCHMARKS=OFF -DLLVM_INCLUDE_EXAMPLES=OFF -DLLVM_INCLUDE_TESTS=OFF`, mirroring the repl branch. These suppress targets that never end up in the cache but cost configure time and incremental dependency tracking.

Add `clangInterpreter` to the cling ninja line. ROOT's bundled `interpreter/cling/lib/Interpreter/CMakeLists.txt:14` lists `clangInterpreter` in its `LIBS`, but cling v1.3 (the version we cache against) does not. Without it, downstream consumers that rebuild ROOT's bundled cling fail at link with `libclangInterpreter.a ... missing and no known rule to make it`: the imported target from `ClangTargets.cmake` points at a file that was never built. The Build_LLVM trim deletes the top-level `llvm/CMakeLists.txt`, so the lib cannot be ninja'd later from the cached build -- it must be built here.

Fix Build_LLVM_WASM source-trim. The Linux block's `find` was missing its closing `)`, leaving the substitution unterminated and swallowing the next half-dozen lines. Cling-on-WASM trims have been broken since the action was last refactored; correct on its own merits and required for the source-`lib/` shrink to actually take effect on the WASM side.

Quote `cling: 'On'` on all matrix rows. YAML 1.1 parsers (notably nektos/act) interpret bare `On` as boolean true; the upper-cased comparison in Save_PR_Info against the literal string `"ON"` then fails, the row gets `CLING_HASH=Repl`, and the cache key drifts off real CI. GitHub Actions own parser preserves bare `On` as a string, so real CI was unaffected; quoting eliminates the act-only divergence.

Cache key cleanup. The previous key embedded git ls-remote internals verbatim (`refs/tags/v1.3`), a literal `-clang-` separator that produced `clang-clang-20` when `matrix.compiler` was clang, and a `.x` suffix appended unconditionally to `matrix.root-llvm-tag` (a leftover from when the formula assumed a `release/<major>.x` branch). Save_PR_Info now extracts the SHA via `awk` and prefixes it with a readable label, truncated to 8 chars. The formula in `main.yml` swaps `-clang-` for `-llvm`, drops the `.x`, and only emits `-<root-llvm-tag>` when the tag is set. The shape change itself invalidates the existing cache pool; no salt is needed.

Save_PR_Info is also converted to LF line endings (it was the only CRLF file in the action set, which made every prior diff to the file noisy with `^M`).

### DIFF
--- a/.github/actions/Build_LLVM/action.yml
+++ b/.github/actions/Build_LLVM/action.yml
@@ -33,10 +33,23 @@ runs:
                 -DCLANG_ENABLE_ARCMT=OFF                           \
                 -DCLANG_ENABLE_FORMAT=OFF                          \
                 -DCLANG_ENABLE_BOOTSTRAP=OFF                       \
+                -DLLVM_INCLUDE_BENCHMARKS=OFF                      \
+                -DLLVM_INCLUDE_EXAMPLES=OFF                        \
+                -DLLVM_INCLUDE_TESTS=OFF                           \
                 -G Ninja                                           \
                 ../llvm
           ninja clang -j ${{ env.ncpus }}
           ninja LLVMOrcDebugging -j ${{ env.ncpus }}
+          # `clangInterpreter` (clang-repl's Interpreter library) is not
+          # a transitive dep of the `clang` driver, and the cling version
+          # we cache (v1.3) doesn't list it in its `LIBS`. ROOT's bundled
+          # `interpreter/cling/lib/Interpreter/CMakeLists.txt:14` does,
+          # so downstream consumers (the ROOT integration job) link
+          # against `libclangInterpreter.a` and fail with `missing and
+          # no known rule to make it` if it isn't in the cached build.
+          # The trim removes top-level `llvm/CMakeLists.txt`, so we can't
+          # ninja it later from the cache -- it has to be built here.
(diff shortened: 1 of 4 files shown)


vgvassilev:
That is now part of #932

PR #938, opened 2026-04-26
[ci] Auto-detect OOP-JIT support from the LLVM version.

vgvassilev:
`LLVM_BUILT_WITH_OOP_JIT` was a user-driven cmake variable that told `lib/CppInterOp/Compatibility.h` whether to compile its out-of-process JIT path. CI threaded it through the matrix as `matrix.oop-jit`, which meant we needed dedicated cells (and a dedicated `-oop` cache slot) to flip it on. With fb595acb (#924) porting the OOP path to upstream `release/22.x`'s
`clang::IncrementalExecutorBuilder`, the criterion is no longer a user toggle -- it's "is the upstream API present", which is "LLVM >= 22 on a supported platform".

Switch to that. CMakeLists.txt now sets
`-DLLVM_BUILT_WITH_OOP_JIT` automatically when
`LLVM_VERSION_MAJOR >= 22` and the platform is Linux x86_64/ aarch64 or macOS arm64 (the set CppInterOp's OOP path has been validated against). For older runtimes (clang-runtime 20/21 in the matrix) and other platforms, the OOP `#ifdef`s in `Compatibility.h`, `CppInterOpInterpreter.h`, and
`unittests/CppInterOp/Utils.h` continue to gate the code out exactly as before -- so cling rows (clang-20) and Windows rows keep their in-process-only behavior.

`Build_and_Test_CppInterOp` no longer passes
`-DLLVM_BUILT_WITH_OOP_JIT=${{ matrix.oop-jit }}` because the cmake side decides for itself. The matrix's `oop-jit` field still exists -- it's now just a Valgrind-mode hint at
`Build_and_Test_CppInterOp/action.yml:68`, controlling whether Valgrind invokes the test binary with OOP-specific suppressions. The rows themselves can collapse into the existing `*-llvm22` rows in a follow-up; that's a coverage decision rather than a mechanical one.

Cache-shape changes that ride along:

  - Drop `${{ matrix.oop-jit == 'On' && '-oop' || '' }}` from the cache key in `main.yml`. Non-OOP rows had this evaluate to empty string already, so their key is bit-identical -- no invalidation. The previously `-oop`-suffixed entries get orphaned (still in storage until eviction, just unreferenced).
  - Drop the `oop-jit`-gated `ninja llvm-jitlink-executor orc_rt-<arch>` block from `Build_LLVM`. The cached LLVM build is now uniform between OOP-capable and in-process-only rows on the same axes.
  - Add an `Install OOP-JIT runtime` step in `main.yml` that runs on `clang-runtime >= 22` non-cling Linux/macOS rows and layers `llvm-jitlink-executor` + `orc_rt-<arch>` from apt.llvm.org / Homebrew on top of the cached LLVM, symlinked into the resource layout the driver's toolchain searches. The cached LLVM keeps `LLVM_ENABLE_ASSERTIONS=ON` (which is what catches LLVM-internal assertions during JIT compilation -- the bit we want assertion coverage for); `orc_rt` and the executor are runtime support code, not assertion-rich, so sourcing them from packages doesn't lose signal.

Restoring a separate OOP build shape (e.g. for a future variant that needs an LLVM-internal patch again) is local: re-add the `matrix.oop-jit == 'On' && '-oop' || ''` segment to the cache key and the matching ninja gate in `Build_LLVM`.

### DIFF
--- a/.github/actions/Build_LLVM/action.yml
+++ b/.github/actions/Build_LLVM/action.yml
@@ -55,14 +55,6 @@ runs:
                 -DLLVM_INCLUDE_TESTS=OFF                        \
                 ../llvm
           ninja clang clangInterpreter clangStaticAnalyzerCore
-          if [[ "${{ matrix.oop-jit }}" == "On" ]]; then
-            if [[ "${{ matrix.os }}" == macos* ]]; then
-              SUFFIX="_osx"
-            elif [[ "${{ matrix.os }}" == ubuntu* ]]; then
-              SUFFIX="-x86_64"
-            fi
-            ninja llvm-jitlink-executor orc_rt${SUFFIX} -j ${{ env.ncpus }}       
-          fi
           cd ./tools/
           rm -rf $(find . -maxdepth 1 ! -name "clang" ! -name ".")
           cd ..
--- a/.github/actions/Build_and_Test_CppInterOp/action.yml
+++ b/.github/actions/Build_and_Test_CppInterOp/action.yml
@@ -54,7 +54,6 @@ runs:
                 -DCODE_COVERAGE=${{ env.CODE_COVERAGE }}     \
                 -DCMAKE_INSTALL_PREFIX=$CPPINTEROP_DIR       \
                 -DLLVM_ENABLE_WERROR=On                      \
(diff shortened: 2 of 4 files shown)


vgvassilev:
This is already implemented in a slightly different way.

PR #941, opened 2026-04-26
Add missing ) so that Emscripten llvm action works

mcbarton:
I have this added in my Emscripten llvm 22 PR so that the Emscripten llvm action can work. I have just opened this separate PR so that CppInterOps ci on main can pass until the Emscripten llvm 22 PR has been approved to keep everything moving. I will cancel the cis so this PR doesn't interfere with anything else.


vgvassilev (approved):
Lgtm! Thank you, @mcbarton!

PR #942, opened 2026-04-27
[jit] Add a layered probe chain for relocatable OOP runtime discovery.

vgvassilev:
The bake-at-build-time install path from `[jit] Self-contain the OOP-JIT runtime in CppInterOp's distribution.` is fragile under relocation: an installed CppInterOp moved to a different prefix (rpm `--prefix`, conda relocation, AppImage) can't find the runtime parts because the macro holds the original
`CMAKE_INSTALL_FULL_LIBDIR`.

`configureBundledOOPRuntime` now probes a layered list, most-reliable first, before falling back to the baked path:

  1. `$CPPINTEROP_RUNTIME_DIR` -- sysadmin override; trumps every other source so packagers can point CppInterOp at a runtime extracted to a separate location.
  2. `<dir of libclangCppInterOp>/cppinterop-rt` -- self-DSO discovery via `dladdr` (POSIX). Works for any non-static link: if the `.so` moves, the runtime moves with it, no env-var needed.
  3. `CPPINTEROP_RUNTIME_BUILD_DIR` -- in-tree test runs.
  4. `CPPINTEROP_RUNTIME_INSTALL_DIR` -- baked install path; last resort when self-DSO discovery isn't available (Windows, or CppInterOp statically linked into a host binary).

Windows has no self-DSO discovery yet -- a `GetModuleHandleEx` port can be added when Windows OOP support arrives. The clang resource directory (`CLANG_RESOURCE_DIR`) suffers the same
bake-at-build-time-vs-relocation problem and would benefit from the same layering; left as follow-up.

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/actions/Build_and_Test_CppInterOp/action.yml
+++ b/.github/actions/Build_and_Test_CppInterOp/action.yml
@@ -100,6 +100,41 @@ runs:
             test -x "${INSTALL_DIR}/lib/cppinterop-rt/llvm-jitlink-executor" || \
               { echo "FAIL: llvm-jitlink-executor missing or not executable"; exit 1; }
             echo "OK: bundled OOP runtime installed under lib/cppinterop-rt/"
+
+            # Probe-chain validation: relocate the install and override
+            # CPPINTEROP_RUNTIME_DIR. Both must keep the bundled runtime
+            # discoverable -- assert no fallback warning fires.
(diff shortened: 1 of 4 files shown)


vgvassilev:
The asan issue is preexisting and I am working on fixing it.

PR #943, opened 2026-04-27
[ci] Verify the OOP runtime probe chain via relocation + env-var.

vgvassilev:
The layered probe chain in `configureBundledOOPRuntime` from
`[jit] Add a layered probe chain for relocatable OOP runtime
discovery.` adds two new lookup layers (`CPPINTEROP_RUNTIME_DIR`
env-var override and `dladdr`-based self-DSO probe) that the
existing in-tree test path doesn't exercise -- in-tree the baked
`CPPINTEROP_RUNTIME_BUILD_DIR` probe wins first, so a regression
in either new layer would be silent.

After the install-layout assertion succeeds, move the install to
a fresh prefix and re-run the OOP-tagged tests with
`DYLD_LIBRARY_PATH`/`LD_LIBRARY_PATH` pointing at the moved tree:
the `dladdr` probe must follow the relocated `.so` and find the
runtime alongside it. Then stage the runtime parts at a third
location, wipe the moved `cppinterop-rt/` subdir, and re-run with
`CPPINTEROP_RUNTIME_DIR` set to the staged copy: only the env-var
probe can win.

Both runs assert that the
`[CreateClangInterpreter]: --use-oop-jit requested but the bundled
OOP runtime ... is missing` fallback warning does NOT appear in
the test output. Negative-control verified locally -- with all
probe locations wiped the warning does fire, confirming the
assertion isn't dead.

Filter is `CppInterOpTest/OutOfProcessJIT.Jit_StreamRedirectJIT`,
the only OOP-tagged test that calls `CreateInterpreter()` before
its skip check; the test itself fails with a preexisting
`stdio.h not found` issue in the OOP child interpreter, which is
orthogonal to the probe chain and doesn't trip the bundling
warning.

### DIFF
--- a/.github/actions/Build_and_Test_CppInterOp/action.yml
+++ b/.github/actions/Build_and_Test_CppInterOp/action.yml
@@ -100,6 +100,41 @@ runs:
             test -x "${INSTALL_DIR}/lib/cppinterop-rt/llvm-jitlink-executor" || \
               { echo "FAIL: llvm-jitlink-executor missing or not executable"; exit 1; }
             echo "OK: bundled OOP runtime installed under lib/cppinterop-rt/"
+
+            # Probe-chain validation: relocate the install and override
+            # CPPINTEROP_RUNTIME_DIR. Both must keep the bundled runtime
+            # discoverable -- assert no fallback warning fires.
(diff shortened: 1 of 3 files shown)


vgvassilev:
Landed in #942

PR #948, opened 2026-04-28
[ci] Source asan LLVM from ci-workflows recipe cache.

vgvassilev:
The ubu24-x86-clang22-llvm22-asan-ubsan matrix row now pulls a prebuilt asan-instrumented LLVM via
compiler-research/ci-workflows' setup-recipe action instead of rebuilding LLVM in this repo's CI. Other rows are unchanged; the existing actions/cache restore + Build_LLVM + cache save trio is gated by !matrix.use-recipe so it doesn't run for the asan row at all — the recipe cache is the only place LLVM comes from.

setup-recipe is invoked with build-on-miss: false. A cache miss here means the publish-recipe pipeline didn't warm the cell we asked for, and rebuilding LLVM inline (~30 min) on every PR run would defeat the point of the cache. The action fails fast with a populate-via-gh-workflow-run hint, letting the operator warm the cell explicitly rather than silently paying the build cost on every consumer.

Downstream Build_and_Test_CppInterOp discovers the tree at $GITHUB_WORKSPACE/llvm-project automatically; no env wiring needed.

setup-recipe is pinned to a ci-workflows commit sha (2959f335445995fff4613bc1fa868171657b7672 — the trim-tightening commit on main). Bump explicitly when consuming new behaviour.

This is the first end-to-end consumer migration. The expectation: the asan row's first PR run hits the published asan asset on compiler-research/ci-workflows' cache release, downloads the ~1-2 GB tree in seconds, and runs Build_and_Test_CppInterOp against it. CI minutes saved per PR: ~30 minutes (the previous in-place asan-LLVM build).

Other rows (cling, Windows, self-hosted CUDA, Linux/macOS non-cling) keep using the local Build_LLVM action for now. Migration to setup-system-llvm for the Linux/macOS non-cling rows depends on the setup-system-llvm action landing in ci-workflows main; tracked as a follow-up branch.

### DIFF
--- a/.github/actions/Build_and_Test_CppInterOp/action.yml
+++ b/.github/actions/Build_and_Test_CppInterOp/action.yml
@@ -14,14 +14,36 @@ runs:
           CPLUS_INCLUDE_PATH="/usr/lib/llvm-${{ matrix.clang-runtime }}/include:$PWD/include"
         else
           LLVM_DIR="$(pwd)/llvm-project"
-          LLVM_BUILD_DIR="$(pwd)/llvm-project/build"
+          # Two layouts coexist on github-hosted runners:
+          #   - rows that build LLVM from source (everything except the
+          #     asan row today) leave a build tree at llvm-project/build/.
+          #   - rows that pull a prebuilt asset via setup-recipe extract
+          #     a cmake --install tree directly under llvm-project/, with
+          #     LLVMConfig.cmake at lib/cmake/llvm/ (no build/ prefix).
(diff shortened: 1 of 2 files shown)


vgvassilev:
The failures are pre-existing and started appearing after #932. That's likely due to some broken patch between cling and root: https://github.com/root-project/llvm-project/compare/cling-llvm20...root-project:llvm-project:ROOT-llvm20

investigating but this should not block merging the pr.

PR #954, opened 2026-05-01
[cmake] Gate install and LLVM BUILDTREE_ONLY  for non-standalone builds

aaronj0:
When CppInterOp is consumed via add_subdirectory() in an external project like ROOT, we previously had to carry two small CMake patches on top of upstream:

 - clangCppInterOp must be BUILDTREE_ONLY so AddLLVM does not register/export it, ROOT folds the symbols into libCling rather than shipping a separate library.
 - The standalone install() rules for the cmake config, headers, tools and build-tree include/ should not be default behaviour, ROOT installs through libCling's own layout


aaronj0:
Closing, landed via #1011

PR #955, opened 2026-05-01
[ci]  Update ROOT workflow to standard upgrade process with cached incremental builds.

aaronj0:
Updates `root.yml` to mirror manual upgrade process we follow when bumping 
the CppInterOp version in ROOT                                           
                                                                                            
1. Clone ROOT master (cached, `git fetch`'d incrementally on cache hit).
2. Replace `interpreter/CppInterOp` with the current PR's CppInterOp HEAD
3. Strip the few files that don't belong in the embedded copy
4. Run a classic ROOT build with LLVM/Cling compiled from source (cached).        
5. Run ROOT's `ctest` suite against that build.                                   

Most platform independant breaking changes in the API or build system CppInterOp
introduces in ROOT will show up as a configure, compile, or test failure.

subset of these changes are from https://github.com/compiler-research/CppInterOp/pull/954 which this workflow currently uses to avoid temporary patches.

### DIFF
--- a/.github/workflows/root.yml
+++ b/.github/workflows/root.yml
@@ -155,7 +155,7 @@ jobs:
 
     - name: Build ROOT
       shell: bash
-      run: ninja -C root-build -j ${{ env.ncpus }} Cling CppInterOpUnitTests
+      run: cmake --build root-build --target Cling --target CppInterOpUnitTests --parallel ${{ env.ncpus }}
 
     - name: Save ROOT source + build cache
       if: ${{ runner.environment != 'self-hosted' && steps.cache.outputs.cache-hit != 'true' }}
--- a/.github/workflows/root.yml
+++ b/.github/workflows/root.yml
@@ -1,16 +1,21 @@
 name: ROOT
 
 # Build ROOT (root-project/root master) against the CppInterOp tree
-# carried by the PR, by overwriting `interpreter/CppInterOp` in ROOT
-# with our checkout. Any breaking change CppInterOp introduces that
-# ROOT relies on shows up here as a configure or compile error.
+# carried by the PR. This mirrors the manual upgrade flow we follow
+# when bumping the CppInterOp commit pinned in ROOT:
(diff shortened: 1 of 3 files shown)


.github/workflows/root.yml
```diff
@@ -127,88 +72,103 @@ jobs:
...
-              -Dtesting=OFF \
-              -Dfail-on-missing=ON \
-              -Dbuiltin_llvm=OFF \
-              -Dbuiltin_clang=OFF \
-              -DCMAKE_PREFIX_PATH="$GITHUB_WORKSPACE/llvm-project" \
-              -DClang_DIR="$GITHUB_WORKSPACE/llvm-project/lib/cmake/clang" \
-              -DCLANG_INSTALL_PREFIX="$GITHUB_WORKSPACE/llvm-project" \
-              -DCMAKE_CXX_FLAGS="-I${CPPINTEROP_GEN_INC}" \
-              ../root
+              -DCMAKE_CXX_FLAGS="-I$PWD/root-build/include"
```
vgvassilev, line 154: I don't think this is what we want. We want to build the Cling target and run the tests of CppInterOp itself.

.github/workflows/root.yml
```diff
@@ -127,88 +72,103 @@ jobs:
...
 
     - name: Build ROOT
       shell: bash
+      run: ninja -C root-build -j ${{ env.ncpus }} Cling CppInterOpUnitTests
```
vgvassilev:
```suggestion
     run: cmake --build root-build --target CppInterOpUnitTests --target Cling --parallel ${{ env.ncpus }}
```

vgvassilev:
@aaronj0, perhaps we should explore this approach on the root side:
```cmake
  set(CMAKE_SKIP_INSTALL_RULES TRUE)   
  add_subdirectory(CppInterOp)                                                                                                                                
  set(CMAKE_SKIP_INSTALL_RULES FALSE)   # restore for ROOT's own installs that follow
```

PR #966, opened 2026-05-03
[ci] Acquire LLVM via setup-llvm; drop Build_LLVM action.

vgvassilev:
Each consumer workflow has its own apt-llvm.org / brew / actions/cache / inline-build plumbing for acquiring an LLVM tree, plus per-row conditionals to pick between vanilla, sanitizer, and cling-bundled builds. The same dispatcher lives in clad and cppyy too, with subtle drift between copies.

compiler-research/ci-workflows now ships setup-llvm, a composite action that takes one `flavor` input and picks the source: '' pulls the latest LLVM from the llvm-release recipe cache (cells warmed across all six supported runner images), 'system' goes to apt-llvm.org / brew llvm@N for older majors the package manager already ships, 'asan' resolves the llvm-asan recipe, 'cling' resolves llvm-root and builds cling on top inside the action. The arch input is derived from the os slug inside the action, so the matrix doesn't need to carry it.

Adopt it across main.yml and root.yml. The matrix axes rename: use-recipe / recipe-version -> flavor / flavor-version. Vanilla LLVM 22 rows omit flavor (default). LLVM 21 rows pick flavor=system. The asan row picks flavor=asan. cling rows pick flavor=cling with the appropriate flavor-version.

The following steps collapse into one Setup LLVM call:
  - Save PR Info (CLING_HASH / LLVM_HASH no longer feed an actions/cache key in main.yml; emscripten.yml and deploy-pages.yml still consume it)
  - Restore cached LLVM (Clang-REPL) build
  - Install LLVM dependencies on self-hosted (folded into setup-llvm flavor=system)
  - Build LLVM (the Build_LLVM action; deleted with this commit)
  - Setup recipe (called setup-recipe directly with use-recipe-driven inputs; setup-llvm wraps it now)
  - Build cling on top of recipe LLVM (Unix + Windows variants; folded into setup-llvm's flavor=cling path)
  - Cache LLVM (actions/cache save)

Net diff: main.yml goes from 304 to 197 lines, root.yml from 217 to 180, and the local Build_LLVM action (113 lines) is deleted.

Self-hosted CUDA row stays under build-selfhost; it picks flavor=system because there's no llvm-release cell for ubuntu-22.04 (jammy) and apt-llvm.org has the packages we need there. Mirrors what the previous in-line apt step did by hand.

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist


vgvassilev:
The emscripten failure does not seem related to this PR. The PR touches explicitly non-emscripten workflows.

PR #969, opened 2026-05-05
[CppInterOp] Tracing: emit reproducer-safe string literals.

vgvassilev:
ReproBuffer's append(const char*) and append(const std::string&) currently wrap the bytes in a plain "..." literal with no escape pass. Any string the traced API receives that contains ", \, or a control byte -- most commonly the newline embedded in Cpp::Process / Cpp::Declare source blocks -- breaks out of the C++ literal mid- content, and the auto-generated reproducer fails to compile before any traced call has a chance to run.

Pick the literal form by content: scan once for ", \, control bytes (< 0x20), and 0x7f; emit "..." when none are present, otherwise R"CPPI(...)CPPI" (which passes bytes through verbatim without an escape pass). The plain form keeps trace lines short for the common case (identifiers, paths, simple expressions); the raw form catches the rest. A traced string containing the literal sequence )CPPI" would still break the raw form -- a vanishingly unlikely collision that is the standard CPPI-delimiter trade-off, not addressed here.

Tests: five new TracingTest cases pin both branches -- empty string takes plain, embedded ", \, and \n each route to the raw form, and a UTF-8 high-bit string stays on the plain path.

### DIFF
--- a/lib/CppInterOp/Tracing.h
+++ b/lib/CppInterOp/Tracing.h
@@ -221,11 +221,30 @@ struct ReproBuffer {
       OS << h;
   }
 
-  // Strings — emit as a raw string literal R"CPPI(...)CPPI" so newlines,
-  // quotes, and backslashes pass through verbatim. The CPPI delimiter
-  // makes accidental in-content collisions vanishingly unlikely (would
-  // need the literal sequence )CPPI" inside a traced string).
-  void appendRaw(std::string_view s) { OS << "R\"CPPI(" << s << ")CPPI\""; }
+  // Strings -- emit a plain `"..."` literal when the content is
+  // printable ASCII without `"`, `\`, or control chars; otherwise
+  // fall back to `R"CPPI(...)CPPI"`, which passes the bytes through
+  // verbatim. The raw form is reserved for the cases that actually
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/Tracing.h
```diff
@@ -221,11 +221,28 @@ struct ReproBuffer {
...
+  // verbatim. The raw form is reserved for the cases that actually
+  // need it (quotes, backslashes, newlines from Cpp::Process /
+  // Cpp::Declare source blocks); plain literals keep the trace lines
+  // short and human-readable for the common case (identifiers,
+  // paths, simple expressions).
+  void appendRaw(std::string_view s) {
+    bool NeedsRaw = false;
+    for (char c : s) {
+      auto u = static_cast<unsigned char>(c);
+      if (c == '"' || c == '\\' || u < 0x20 || u == 0x7f) {
```
Vipul-Cariappa, line 238: I understand the changes, but what magic numbers here: `0x20` & `0x7f`. Can we add a comment?

Vipul-Cariappa (approved):
Can we add a test with newline character (multi-line strings)? I did not see it, or might have missed.

PR #970, opened 2026-05-05
Fix: GetEnumConstantValue when value can't be within int64_t

Vipul-Cariappa:


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -793,7 +793,11 @@ TCppIndex_t GetEnumConstantValue(TCppScope_t handle) {
   auto* D = (clang::Decl*)handle;
   if (auto* ECD = llvm::dyn_cast_or_null<clang::EnumConstantDecl>(D)) {
     const llvm::APSInt& Val = ECD->getInitVal();
-    return INTEROP_RETURN(Val.getExtValue());
+    if (Val.isRepresentableByInt64())
+      return INTEROP_RETURN(Val.getExtValue());
+    llvm::SmallString<40> StrVal;
+    Val.toString(StrVal);
+    return INTEROP_RETURN(std::stoul(StrVal.c_str()));
```
Vipul-Cariappa, line 890: Do we need macro to check if `sizeof(unsigned long) == sizeof(size_t)` then call `stoul`, else if `sizeof(unsigned long long) == sizeof(size_t)`, call `stoull`? For all the systems we are interested (64 bit), I believe `sizeof(unsigned long) == sizeof(unsigned long long) == 64`. Then we will also have to change the return type. What is the enum value is negative...!?

Vipul-Cariappa: what if `int128_t` https://godbolt.org/z/zv758P33a?

vgvassilev (approved):
LGTM!

aaronj0:
Hi, just wanted to see if this can be rebased and landed.

PR #971, opened 2026-05-05
Fix: JitCall codegen for deleted copy ctor in args

Vipul-Cariappa:
# Description

Change to use the move constructor when the copy constructor is deleted among the arguments to a constructor.

Moves closer to what Cling does.

## Fixes # (issue)

ROOT tests.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -2568,6 +2568,33 @@ void collect_type_info(const FunctionDecl* FD, QualType& QT,
   get_type_as_string(QT, type_name, C, Policy);
 }
 
+static bool IsCopyConstructorDeleted(QualType QT) {
+  CXXRecordDecl* RD = QT->getAsCXXRecordDecl();
+  if (!RD) {
+    // For types that are not C++ records (such as PODs), we assume that they
+    // are copyable, ie their copy constructor is not deleted.
+    return false;
+  }
+
+  RD = RD->getDefinition();
+  assert(RD && "expecting a definition");
+
+  if (RD->hasSimpleCopyConstructor())
+    return false;
+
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2568,6 +2568,33 @@ void collect_type_info(const FunctionDecl* FD, QualType& QT,
...
 
+static bool IsCopyConstructorDeleted(QualType QT) {
+  CXXRecordDecl* RD = QT->getAsCXXRecordDecl();
+  if (!RD) {
+    // For types that are not C++ records (such as PODs), we assume that they
+    // are copyable, ie their copy constructor is not deleted.
+    return false;
+  }
+
+  RD = RD->getDefinition();
```
vgvassilev, line 2579: We can add complete type call to the cppinterop interfaces which has probably a raii object as this operation might trigger deserialization.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2568,6 +2568,33 @@ void collect_type_info(const FunctionDecl* FD, QualType& QT,
...
+  assert(RD && "expecting a definition");
+
+  if (RD->hasSimpleCopyConstructor())
+    return false;
+
+  for (auto* Ctor : RD->ctors()) {
+    if (Ctor->isCopyConstructor()) {
+      return Ctor->isDeleted();
+    }
+  }
```
vgvassilev, line 2589:
```suggestion
  for (auto* Ctor : RD->ctors())
    if (Ctor->isCopyConstructor())
      return Ctor->isDeleted();
  }
```

vgvassilev (approved):
LGTM!

PR #972, opened 2026-05-06
Use threadsafe death-test style in TracingTests

aaronj0:
Fixes sporadious failures on macOS26 and mac-beta. On apple systems, gtest's default "fast" death-test style fork()s the child without exec(), so the child inherits parent process state (atfork handlers, dyld closures, stdio locks) which can kill/stall before `EXPECT_DEATH` writes to stderr, so gtest captures an empty buffer and reports "died but not with expected error" with an empty Actualmsg. Seen on ROOT CI.

This is a known issue and the fix as seen in LLVM's compiler-rt sanitizer tests, which set `testing::GTEST_FLAG(death_test_style) = "threadsafe"; `  in their gtest entry points. Also see`gtest-death-test.cc:250` - "Death tests use fork(), which is unsafe particularly in a threaded context", and the project's docs recommend threadsafe for this case.

### DIFF
--- a/unittests/CppInterOp/TracingTests.cpp
+++ b/unittests/CppInterOp/TracingTests.cpp
@@ -32,13 +32,25 @@ static std::string getFullLog() {
 
 class TracingTest : public ::testing::Test {
 protected:
+  static void SetUpTestSuite() {
+    saved_death_test_style_ = ::testing::GTEST_FLAG(death_test_style);
+    GTEST_FLAG_SET(death_test_style, "threadsafe");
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
LGTM!

aaronj0:
@vgvassilev the error on the asan build is solved, moved the setting under the `SetUpTestSuite`  in your added fixture with a static so the string is no longer leaked and the previous gtest `death_test_style` config is restored at teardown

PR #973, opened 2026-05-06
[Tracing] Format std::vector input args as init lists.

vgvassilev:
The reproducer formatter emits `{...}` for any std::vector argument,
which leaks through to generated reproducers and makes them
un-buildable as soon as a non-OutParam vector reaches a Cpp:: call --
e.g. CreateInterpreter's argv/GpuArgs vectors land as
`Cpp::CreateInterpreter({...}, {...})`, which is not valid C++ for
an `std::vector<const char*>` parameter.

Recurse through `append` so each element is formatted by the
existing overload set; the result is a real C++ braced init list
the reproducer can compile (`{1, 2, 3}` for ints,
`{"a", "b"}` for strings, `{{1, 2}, {3}}` for nested vectors). The
traversal is O(N) per vector arg; tracing is a debug path so the
cost is acceptable.

Tests: four new TracingTest cases pin each branch of the formatter
-- empty vector (First-flag never flips), int vector, string vector
(reuses appendRaw via append(const char*)), and nested
vector<vector<int>> (recursive dispatch).

Depends on #969.

### DIFF
--- a/lib/CppInterOp/Tracing.h
+++ b/lib/CppInterOp/Tracing.h
@@ -261,8 +261,21 @@ struct ReproBuffer {
   void append(double d) { OS << llvm::formatv("{0:f}", d); }
   void append(float f) { OS << llvm::formatv("{0:f}", f); }
 
-  // Containers — not meaningfully printable; emit a placeholder.
-  template <typename T> void append(const std::vector<T>&) { OS << "{...}"; }
+  // Vector input args -- emit a braced init list of recursively-formatted
+  // elements so the reproducer's call-site literal compiles. The previous
+  // `{...}` placeholder leaked through to generated reproducers and made
+  // them un-buildable as soon as a non-OutParam vector was passed in.
+  template <typename T> void append(const std::vector<T>& V) {
+    OS << "{";
+    bool First = true;
+    for (const auto& E : V) {
+      if (!First)
+        OS << ", ";
+      First = false;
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/Tracing.h
```diff
@@ -261,8 +261,21 @@ struct ReproBuffer {
   void append(double d) { OS << llvm::formatv("{0:f}", d); }
   void append(float f) { OS << llvm::formatv("{0:f}", f); }
 
-  // Containers — not meaningfully printable; emit a placeholder.
-  template <typename T> void append(const std::vector<T>&) { OS << "{...}"; }
+  // Vector input args -- emit a braced init list of recursively-formatted
+  // elements so the reproducer's call-site literal compiles. The previous
+  // `{...}` placeholder leaked through to generated reproducers and made
+  // them un-buildable as soon as a non-OutParam vector was passed in.
```
Vipul-Cariappa:
```suggestion
  // elements so the reproducer's call-site literal compiles.
```

Unnecessary comments, according to me. I believe the commit message explains the motivation.

Vipul-Cariappa (approved):
LGTM!

What about out-param vectors? Any ideas there?

vgvassilev:
> LGTM!
> 
> What about out-param vectors? Any ideas there?

Coming in one of the next changes.

PR #974, opened 2026-05-07
[Tracing] Track returned std::vector<P*> as named handles.

vgvassilev:
Functions that return std::vector<P*> (e.g. GetFunctionsUsingName, GetClassMethods) hand the caller a list of opaque clang Decl*. The trace records the call but discards the elements, so the reproducer sees `Cpp::GetFunctionsUsingName(...)` with no producing line for any pointer the caller subsequently passes to other Cpp:: APIs -- those sites print as `nullptr`, and the auto-generated reproducer no longer matches the original control flow.

Snapshot the element pointers in `record()` (the source vector is moved-from before `~TraceRegion` runs, so we copy the addresses out). The destructor allocates a `_retN` slot, names the call site `auto _retN = Cpp::Foo(...)`, and emits one `void* vK = _retN[i]` extraction per newly-registered element with a `.size() > i` bounds guard so the reproducer is safe even when the replay yields a shorter vector. Null elements are skipped so the decl list does not shadow the global `nullptr`. The `_retN` index lives in its own counter, not `m_VarCount`, so the `vN` handle namespace stays dense for the single-pointer-return case.

Tests: five new TracingTest cases pin each branch -- happy path (elements registered + decls emitted), downstream consumer (vN by name), empty vector (no decls), duplicate element across calls (no re-decl), and null element in the middle (skipped, surrounding indices preserved).

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/lib/CppInterOp/Tracing.h
+++ b/lib/CppInterOp/Tracing.h
@@ -49,6 +49,10 @@ class TraceInfo {
 
   std::unordered_map<const void*, std::string> m_HandleMap;
   unsigned m_VarCount = 0;
+  /// Monotonic suffix for `_retN` placeholders -- one slot per call
+  /// that returns std::vector<P*>. Separate from m_VarCount so the
+  /// vN handle namespace stays dense.
+  unsigned m_RetCount = 0;
 
   std::vector<std::string> m_Log;
   size_t m_RegionStart = 0; ///< Log index where current region began.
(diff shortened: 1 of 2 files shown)


Vipul-Cariappa (approved):
LGTM!
The `ReturnedPointerVectorDuplicateElementSkipsDecl` is a nice optimisation.

I believe the next step is to get this to work with out-parameters? Like `BestOverloadFunctionMatch(..., std::vector<TCppScope_t>& fns)`. Where we populate the argument `fns`.

vgvassilev:
> LGTM! The `ReturnedPointerVectorDuplicateElementSkipsDecl` is a nice optimisation.
> 
> I believe the next step is to get this to work with out-parameters? Like `BestOverloadFunctionMatch(..., std::vector<TCppScope_t>& fns)`. Where we populate the argument `fns`.

Yes...

PR #975, opened 2026-05-07
[Tracing] Distinguish unregistered pointers from nullptr in arg output.

vgvassilev:
ReproBuffer::append already had a branch meant to flag unregistered non-null pointers as "nullptr /*unknown*/", but the branch was dead. lookupHandle returned the literal string "nullptr" for both actual null and unregistered pointers, so h.empty() never fired and any unrecognised pointer collapsed silently to a bare nullptr -- the reproducer reads as if the original program passed null at that call, masking the missing producing line.

Tighten lookupHandle's contract: "nullptr" only for an actual null pointer, "" for a non-null pointer the tracer never saw, "vN" for a registered handle. The two ~TraceRegion callers that gated their "auto vN = ..." prelude on lookupHandle == "nullptr" now check .empty() instead, matching the "we have no name yet" semantics. The previously-dead "nullptr /*unknown*/" annotation in append now fires for the unregistered case.

Tests: extend HandleRegistry to pin the empty-string return for an unregistered non-null pointer; add ArgFormattingUnregisteredHandle- Annotated (the previously-dead branch is now reached) and ArgFormattingNullHandlePlain (genuine null still renders without the annotation, so the two cases are visibly distinct in the trace).

Depends on #974

### DIFF
--- a/lib/CppInterOp/Tracing.h
+++ b/lib/CppInterOp/Tracing.h
@@ -100,11 +100,18 @@ public:
     return m_HandleMap[p] = "v" + std::to_string(++m_VarCount);
   }
 
+  /// Resolve a pointer to a printable form.
+  /// - "nullptr" for an actual null pointer.
+  /// - "vN" for a pointer registered via getOrRegisterHandle.
+  /// - "" (empty) for a non-null pointer never seen by the tracer --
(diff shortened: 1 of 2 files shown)


Vipul-Cariappa (approved):
LGTM!

PR #978, opened 2026-05-08
[Tracing] Declare _outN buffers and alias them at OUT-param call sites.

vgvassilev:
INTEROP_OUT-marked arguments were dropped from the formatted call line entirely. The reproducer doesn't have the buffer in scope, so "Cpp::FillHandles();" is ill-formed against the real signature "Cpp::FillHandles(std::vector<void*>&)" and the auto-generated reproducer fails to compile every time an OUT pointer-container appears.

Allocate a fresh "_outN" index per OUT pointer-container at captureArg time and emit two reproducer-side artefacts in ~TraceRegion: a "std::vector<void*> _outN;" preamble line before the call, and an alias to that buffer at the OUT argument's position in the call's argument list. The element type is hard-coded to void* because the public API erases pointer-element OUT containers to that type. Non-pointer OUT containers (e.g. vector<string>) keep the legacy skip rendering -- their type is not erasable to void*, so a generic preamble would not type-check; that case waits on type-name capture.

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/unittests/CppInterOp/TracingTests.cpp
+++ b/unittests/CppInterOp/TracingTests.cpp
@@ -168,11 +168,12 @@ TEST_F(TracingTest, TraceWithNoArgs) {
 }
 
 TEST_F(TracingTest, TraceWithOutParamArg) {
+  // OUT pointer-container: preamble decl + alias at the call site.
   std::vector<void*> v;
   WithOutParamTrace(v);
-  auto output = TraceInfo::TheTraceInfo->getLastLogEntry();
-  // OutParam should not appear in the arg list.
-  EXPECT_THAT(output, HasSubstr("Cpp::WithOutParamTrace();"));
+  auto output = getFullLog();
+  EXPECT_THAT(output, HasSubstr("std::vector<void*> _out0;"));
+  EXPECT_THAT(output, HasSubstr("Cpp::WithOutParamTrace(_out0);"));
   EXPECT_EQ(v.size(), 1u);
 }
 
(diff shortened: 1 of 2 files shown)


unittests/CppInterOp/TracingTests.cpp
```diff
@@ -320,15 +321,44 @@ TEST_F(TracingTest, ArgFormattingMixed) {
...
+  // The counter is region-global, not per-call -- two calls each with
+  // one OUT do not collide on _out0 (C++ redefinition error).
+  std::vector<void*> v1, v2;
+  WithOutParamTrace(v1);
+  WithOutParamTrace(v2);
+  auto output = getFullLog();
+  EXPECT_THAT(output, HasSubstr("std::vector<void*> _out0;"));
+  EXPECT_THAT(output, HasSubstr("std::vector<void*> _out1;"));
+  EXPECT_THAT(output, HasSubstr("Cpp::WithOutParamTrace(_out0);"));
+  EXPECT_THAT(output, HasSubstr("Cpp::WithOutParamTrace(_out1);"));
```
Vipul-Cariappa, line 361: What happens if the following is called:
```c++
std::vector<void*> v1;
WithOutParamTrace(v1);
WithOutParamTrace(v1); // same out-param is used
```

Vipul-Cariappa (changes requested):
What happens when the contents of out-param are then passed to some other function? Like:
```c++
void* h = FuncReturningHandle();
std::vector<void*> out;
FuncWithHandleAndOut(h, out);
FuncTakingHandle(out[0]);
```

Don't we need the unwrap as we have for the regular param? Like https://github.com/compiler-research/CppInterOp/blob/dd0474e30e41568ce0b8a47ad93340fdc1e77043/lib/CppInterOp/Tracing.h#L475 ?

Please add a test similar to the above one.

Vipul-Cariappa (approved):
LGTM!

---

To be complete, there was a discussion about this on discord. Pasting the discussion here for reference for the future:

Vassil:
What do we do with your accumulation comment -- do we really use that pattern for something useful idiomatically?

Vipul:
Looking at the API usage in CppInterOp. You can safely consider that each out-param is unique. And the implementation as it is right now (modulo my comment on using the out-param later) should be sufficient. I put the comment as a thing to think about.
For the usage and the current implementation (all `void*` is tracked across calls even when nested with `std::vector`), the accumulation should not be a concern.
FYI, to be complete, A place where out-param is reused: https://github.com/compiler-research/cppyy-backend/blob/597bcf430db3168f1509a5fb56bc9fc1769ce3f6/clingwrapper/src/clingwrapper.cxx#L1766-L1771
But look at the usage, assuming that the out-param is unique is valid here.

Vassil:
Yes, also for base accumulation: 
```c++
std::vector<Cpp::TCppScope_t> datamembers;                                                                                                                
  Cpp::GetDatamembers(myklass, datamembers);                                                                                                                
  for (size_t i = 0; i < num_bases; i++) {                                                                                                                  
    Cpp::TCppScope_t base = Cpp::GetBaseClass(myklass, i);                                                                                                  
    for (size_t i = 0; i < Cpp::GetNumBases(base); i++)                                                                                                     
      Cpp::GetDatamembers(Cpp::GetBaseClass(base, i), datamembers);                                                                                         
    Cpp::GetDatamembers(base, datamembers);                                                                                                                 
  }  
```

vgvassilev:
> LGTM!
> 
> To be complete, there was a discussion about this on discord. Pasting the discussion here for reference for the future:
> 
> Vassil: What do we do with your accumulation comment -- do we really use that pattern for something useful idiomatically?
> 
> Vipul: Looking at the API usage in CppInterOp. You can safely consider that each out-param is unique. And the implementation as it is right now (modulo my comment on using the out-param later) should be sufficient. I put the comment as a thing to think about. For the usage and the current implementation (all `void*` is tracked across calls even when nested with `std::vector`), the accumulation should not be a concern. FYI, to be complete, A place where out-param is reused: https://github.com/compiler-research/cppyy-backend/blob/597bcf430db3168f1509a5fb56bc9fc1769ce3f6/clingwrapper/src/clingwrapper.cxx#L1766-L1771 But look at the usage, assuming that the out-param is unique is valid here.
> 
> Vassil: Yes, also for base accumulation:
> 
> ```c++
> std::vector<Cpp::TCppScope_t> datamembers;                                                                                                                
>   Cpp::GetDatamembers(myklass, datamembers);                                                                                                                
>   for (size_t i = 0; i < num_bases; i++) {                                                                                                                  
>     Cpp::TCppScope_t base = Cpp::GetBaseClass(myklass, i);                                                                                                  
>     for (size_t i = 0; i < Cpp::GetNumBases(base); i++)                                                                                                     
>       Cpp::GetDatamembers(Cpp::GetBaseClass(base, i), datamembers);                                                                                         
>     Cpp::GetDatamembers(base, datamembers);                                                                                                                 
>   }  
> ```

Yes, the now that idiom is supported, too.

PR #979, opened 2026-05-08
[Tracing] Format TemplateArgInfo as a brace-init constructor call. 

vgvassilev:
The catch-all template renders any non-enum, non-pointer arg as `?`, which would silently swallow a TemplateArgInfo passed by value: the m_Type pointer disappears and a reproducer cannot bind the type back to its registered handle.

Add a ReproBuffer::append(const TemplateArgInfo&) overload that emits `Cpp::TemplateArgInfo{<m_Type>, <m_IntegralValue>}`. m_Type goes through the void* path so the registered name (vN) replaces the raw address; nullptr m_IntegralValue renders as the literal `nullptr` (the constructor's default) rather than the empty string the const char* path would otherwise produce. Tests pin both branches of the integral-value path and the registered-handle substitution for m_Type.

Depends on #978

### DIFF
--- a/lib/CppInterOp/Tracing.h
+++ b/lib/CppInterOp/Tracing.h
@@ -23,6 +23,8 @@
 #define CPPINTEROP_TRACE_API
 #endif
 
+#include "CppInterOp/CppInterOpTypes.h"
+
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringMap.h"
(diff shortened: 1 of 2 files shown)


Vipul-Cariappa (approved):
Reviewing the second commit only:
LGTM!
*The First Commit Is Reviewed in #978*
*DO NOT MERGE BEFORE #978*

vgvassilev:
The bot seems stuck -- moving forward.

PR #980, opened 2026-05-08
[Tracing] Deprecate pointer/size InstantiateTemplate, add vector form.

vgvassilev:
The pointer+size InstantiateTemplate API forces every caller to spell two arguments and lose ownership semantics, and the trace records only an opaque pointer to the template-argument array -- the reproducer cannot reconstruct the call.

Add an overload that takes the template arguments by const std::vector<TemplateArgInfo>& and forwards directly to the static helper (so the trace records one call, not nested ones); mark the old overload deprecated with a vendor-portable CPPINTEROP_DEPRECATED macro that older Clang versions (Cling) parse correctly. The TableGen schema gains a Deprecated field consumed by the Decl emitter; the dispatch table suppresses the warning at the cast site since it intentionally exposes every overload. All in-tree callers (31 sites across the C API and unit tests) migrate to the vector form.

Depends on #979

### DIFF
--- a/include/CppInterOp/CppInterOpTypes.h
+++ b/include/CppInterOp/CppInterOpTypes.h
@@ -33,6 +33,17 @@
 #endif
 #endif
 
+// Cross-platform deprecation attribute. Older Clang versions (Cling)
+// mis-parse C++11 `[[deprecated]]` near the return type, so we use
+// the vendor-specific spelling on each compiler.
+#if defined(_MSC_VER)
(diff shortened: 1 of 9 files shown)


Vipul-Cariappa (approved):
Reviewing the 3rd commit only:
LGTM!
*The Previous Commits Are Reviewed in #979*
*DO NOT MERGE BEFORE #979*

PR #982, opened 2026-05-08
[Tracing] Suppress log emission for nested INTEROP_TRACE calls.

vgvassilev:
When a Cpp:: function internally calls another Cpp:: function, both TraceRegions append a call line. The inner line references its caller's parameters and locals, which the reproducer doesn't have in scope -- so the auto-generated reproducer fails to compile every time a public API is layered on another.

Detect nesting at TraceRegion construction by reading the timer stack before pushing the new frame, store the result on TraceData, and gate emission on it: nested calls skip the call line, the `_outN` preamble, the `_retN` decls, and the `_outN` index allocation (so the counter stays dense). Handle-registration callbacks still run for nested frames so the outer call's `auto vN = ...` line can resolve a pointer an inner OUT-param fills.

Depends on #980

### DIFF
--- a/lib/CppInterOp/Tracing.h
+++ b/lib/CppInterOp/Tracing.h
@@ -85,6 +85,9 @@ public:
     return *T;
   }
 
+  /// True when at least one TraceRegion is currently active.
+  bool insideTracedRegion() const { return !m_TimerStack.empty(); }
+
   void pushTimer(llvm::Timer* T) {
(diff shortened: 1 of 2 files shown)


Vipul-Cariappa (approved):
Reviewing the 4th commit only:
LGTM!
*The Previous Commits Are Reviewed in #980*
*DO NOT MERGE BEFORE #980*

PR #987, opened 2026-05-09
[CppInterOp] Add Cpp::GetBuildInfo() with CMake-captured configuration.

vgvassilev:
A trace reproducer or a bug report needs a recorded snapshot of how CppInterOp was built -- LLVM version, sanitizer, compiler, the relevant cmake cache flags. Today reviewers ask for that out of band and the answers drift from what actually shipped.

Mirror Cpp::GetVersion(): a new GetBuildInfo() entry in the .td table, with the function defined in CppInterOp.cpp so the tracing coverage scanner picks up its INTEROP_TRACE annotation alongside the rest of the public API. The body is a raw-string literal expression generated by configure_file from BuildInfo.inc.in into the build tree's include/CppInterOp/BuildInfo.inc, included in-place from GetBuildInfo(). CMake filters CACHE_VARIABLES to LLVM_*, CPPINTEROP_* plus CMAKE_BUILD_TYPE, CMAKE_CXX_STANDARD, CMAKE_PREFIX_PATH; the substituted snapshot is configure-time and travels with the binary that this build produced.

Test: Interpreter_BuildInfo pins the headline labels (build type, compiler, llvm, target, cmake-cache) so a future template edit doesn't silently strip a field, and asserts the foreach captured at least one kept var with its -D prefix.

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist


aaronj0 (approved):
This is great, thank you!

PR #992, opened 2026-05-10
Enable MSan CI coverage on Linux x86_64.

vgvassilev:
Add the ubu24-x86-clang22-llvm22-msan matrix row consuming ci-workflows' new llvm-msan recipe, plus the two integration fixes the row surfaces. The recipe is transparent via setup-llvm's CMAKE_TOOLCHAIN_FILE pickup -- no Build_and_Test plumbing needed.

unittests/googletest gap: the parent's CMAKE_CXX_FLAGS + CMAKE_*_LINKER_FLAGS were never forwarded to ExternalProject_Add (CXX_FLAGS was only set inside `if(EMSCRIPTEN)`, linker flags not at all), so gtest fell back to system libstdc++ + no sanitizer instrumentation and ABI-clashed with the parent at link time. Forward both.

interop gap: when CppInterOp is built against libc++, the host is libc++-linked end to end and JIT'd source emitting libstdc++ `std::__cxx11::*` can't resolve against the host's `std::__1::*`. Push -stdlib=libc++ when `_LIBCPP_VERSION` is at CppInterOp's TU. Spelling out the libc++ include path is load-bearing too: clang derives Driver::Dir from /proc/self/exe (= host binary) not argv[0], so -stdlib=libc++ alone misses <install>/include/c++/v1 and the force-include of `<new>` SIGSEGVs in GenModule -- derive from -resource-dir.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,6 +33,7 @@ jobs:
         #   system   apt-llvm.org / brew llvm@N — for older majors
         #            the package manager already ships
         #   asan     llvm-asan recipe
+        #   msan     llvm-msan recipe (Linux x86_64 only)
         #   cling    llvm-root recipe + cling built on top
         include:
           # Ubuntu Arm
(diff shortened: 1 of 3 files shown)


vgvassilev:
@lhames,  we initially tried the cleaner approach of enabling `-fsanitize=memory` on the embedded interpreter's invocation, gated on __has_feature(memory_sanitizer) at CppInterOp's TU. By construction this would have made every JIT'd write update MSan shadow without per-wrapper plumbing — no `__msan_unpoison` sprinkled through `make_narg_ctor_with_return` / `make_narg_call_with_return,` just a one-line knob that auto-covers every present and future codegen path.            
                                                                                                                                                            
It didn't work because of a JIT-environment / MSan-runtime TLS contract mismatch. ORC defaults to emulated TLS to avoid setting up the TLS templates and DT_TLS blocks that a real loader provides; MSan's runtime, on the other hand, exposes `__msan_va_arg_tls` / `__msan_va_arg_overflow_size_tls` as native TLS. JIT-instrumented code ended up referencing `__emutls_v.__msan_va_arg_tls`, which doesn't exist in any loaded library, and JIT-link failed to materialize the wrapper symbols (Symbols not found: [ __emutls_v.__msan_va_arg_tls, __emutls_v.__msan_va_arg_overflow_size_tls ]). `-fno-emulated-tls` on the interpreter args didn't reach codegen — the ORC IR-transform layer rewrites TLS access regardless. Properly closing this would require ORC-level work: either exposing host-side TLS slots as templates the JIT can resolve, or actually disabling emulated-TLS lowering in the JIT's IRTransformLayer. Both are upstream LLVM changes and out of scope for this patch, so we use the narrower per-wrapper unpoison approach instead.

PR #995, opened 2026-05-11
Preserve using-declaration access for inherited methods

### LINKED ISSUE #220: Regression in using protected base class members as public methods in derived classes
Reproducer:
```python
import cppyy

cppyy.cppdef("""

class MyBase {
protected:
   void foo(int, int) {}
};

class MyDerived : public MyBase {
public:
   using MyBase::foo;
   void foo(int) {}
};

""")

obj = cppyy.gbl.MyDerived()

obj.foo(1, 2)
```
This is valid C++ and used to work with the old cppyy, and now fails like this:
```txt
Traceback (most recent call last):
  File "repro.py", line 20, in <module>
    obj.foo(1, 2)
    ~~~~~~~^^^^^^
TypeError: void MyDerived::foo(int) =>
    TypeError: takes at most 1 arguments (2 given)
```

guitargeek:
When `GetClassDecls<CXXMethodDecl>` walked a class's decls and hit a `UsingShadowDecl` whose target is a method, it pushed the *target* `CXXMethodDecl` from the base class. The target's `getAccess()` returns the access in the base (e.g. `protected`); the elevated access carried by the using-declaration (`public` in the introducing class) was silently lost. Consumers like CPyCppyy filter on `IsPublicMethod`, so a republished overload such as

    class Base   { protected: void foo(int, int); };
    class Derived: public Base {
    public:
        using Base::foo;
        void foo(int);
    };

would never reach the Python proxy - calling `derived.foo(1, 2)` from cppyy raised `TypeError: takes at most 1 arguments (2 given)` (cppyy issue https://github.com/compiler-research/cppyy/issues/220).

Push the `UsingShadowDecl` itself (for the non-constructor case; constructor-using-shadows still go through `findInheritingConstructor`) and teach `CheckMethodAccess` to consult `USD->getAccess()`. A small `UnwrapUsingShadowToFunction` helper unwraps the shadow at the entry of every API that downcasts a `TCppFunction_t` to `FunctionDecl` (return type, arg counts/types/names/defaults, signature, IsConst/Static/Virtual/Constructor/Destructor/Templated/Explicit, function address, operator arity).

`MakeFunctionCallable` keeps the unwrapped target as the wrapped function but threads a `relaxAccessControl` flag into `make_wrapper` when the caller passed a using-shadow: the generated wrapper still references the target by its original qualified name (e.g. `((Base*)obj)->Base::foo(...)`), so access control has to be relaxed during wrapper compilation for it to compile. The runtime call is well-defined because the using-declaration made the member reachable through the derived class.

Adds a `FunctionReflection_GetClassMethods_UsingShadowAccess` regression test covering both the public-promoted case from the issue and a protected-using control case (which must stay hidden).

Fixes: https://github.com/compiler-research/cppyy/issues/220

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1021,6 +1021,27 @@ static const clang::Decl* GetUnderlyingScopeImpl(const clang::Decl* D) {
   return D->getCanonicalDecl();
 }
 
+// If D is a UsingShadowDecl whose target is a function-like decl,
+// return that target; otherwise return D unchanged. The using-shadow
+// is the only carrier of the "effective access" introduced into the
+// derived class, so callers that need access info should consult D
+// before unwrapping.
+static clang::Decl* UnwrapUsingShadowToFunction(clang::Decl* D) {
+  if (auto* USD = dyn_cast_or_null<UsingShadowDecl>(D))
+    if (auto* Target = USD->getTargetDecl())
+      if (isa<FunctionDecl>(Target) || isa<FunctionTemplateDecl>(Target))
+        return Target;
+  return D;
+}
+
+static const clang::Decl* UnwrapUsingShadowToFunction(const clang::Decl* D) {
+  if (const auto* USD = dyn_cast_or_null<UsingShadowDecl>(D))
+    if (const auto* Target = USD->getTargetDecl())
+      if (isa<FunctionDecl>(Target) || isa<FunctionTemplateDecl>(Target))
+        return Target;
+  return D;
+}
+
 DeclRef GetUnderlyingScope(ConstDeclRef DRef) {
   INTEROP_TRACE(DRef);
   if (!DRef)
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1809,6 +1835,13 @@ BestOverloadFunctionMatch(const std::vector<TCppFunction_t>& candidates,
 // the provided AccessSpecifier.
 bool CheckMethodAccess(TCppFunction_t method, AccessSpecifier AS) {
   auto* D = (Decl*)method;
+  // For methods introduced into a derived class via `using Base::name;`,
+  // the effective access is the access of the using-declaration, not of
+  // the underlying target method.
+  if (auto* USD = llvm::dyn_cast_or_null<UsingShadowDecl>(D)) {
+    if (llvm::isa_and_nonnull<CXXMethodDecl>(USD->getTargetDecl()))
+      return USD->getAccess() == AS;
+  }
```
vgvassilev, line 2411: Why not using `UnwrapUsingShadowToFunction`?

guitargeek: This is explained now in the inline comments.

guitargeek:
FYI @vgvassilev : this fix is not strictly needed for the ROOT migration, as the ROOT pythonizations tests have evolved to not hit this edge case anymore. So reviewing this PR is not a priority.

vgvassilev:
> FYI @vgvassilev : this fix is not strictly needed for the ROOT migration, as the ROOT pythonizations tests have evolved to not hit this edge case anymore. So reviewing this PR is not a priority.

Looks quite good already -- can we make sure the codecov reports are addressed with tests -- should be ready to go.

vgvassilev (approved):
Lgtm!

guitargeek:
I'm just fixing the clang tidy warnings

vgvassilev (approved):
Lgtm!

PR #996, opened 2026-05-11
Look up operator overloads by CXXOperatorName in GetFunctionsUsingName

guitargeek:
GetFunctionsUsingName built its lookup key only as an identifier, so queries like "operator==" never matched: operator overloads live in the AST under CXXOperatorName, not under an Identifier. cppyy's namespace getattr fallback (Cppyy::GetMethodsFromName -> Cpp::GetFunctionsUsingName) therefore returned no candidates and getattr(ns, 'operator==') raised AttributeError.

Detect names that start with "operator" followed by a non-identifier character and match the trimmed suffix against OperatorKinds.def. When it matches, build the DeclarationName via getCXXOperatorName(OO_...); otherwise fall back to the identifier lookup. Covers symbolic and multi-token overloads (==, +, (), [], new, delete, new[], delete[], ...). Names that merely start with "operator" (e.g. "operators_count") are left on the identifier path because the next character is an identifier-continue.

Fixes https://github.com/compiler-research/cppyy/issues/219.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -53,6 +53,7 @@
 #include "clang/AST/Stmt.h"
 #include "clang/AST/Type.h"
 #include "clang/AST/VTableBuilder.h"
+#include "clang/Basic/CharInfo.h"
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/DiagnosticSema.h"
 #include "clang/Basic/LangStandard.h"
@@ -1484,6 +1485,29 @@ void DumpScope(ConstDeclRef DRef) {
   return INTEROP_VOID_RETURN();
 }
 
+// Map an operator spelling (e.g. "operator==") to the CXXOperatorName its
+// overloads are stored under, since identifier lookup never matches them.
+// Returns an empty DeclarationName for non-operators (e.g. "operators_count"),
+// leaving the caller on the identifier path. We can't use LookupOperatorName
+// for this: it ignores class members, which this entry point must also find.
+static DeclarationName getCXXOperatorDeclName(ASTContext& Ctx,
+                                              llvm::StringRef name) {
+  static constexpr llvm::StringRef OperatorPrefix("operator");
+  if (!name.consume_front(OperatorPrefix) || name.empty() ||
+      clang::isAsciiIdentifierContinue(
+          static_cast<unsigned char>(name.front())))
+    return DeclarationName();
+
+  llvm::StringRef Spelling = name.trim();
+#define OVERLOADED_OPERATOR(OpName, OpSpelling, Token, Unary, Binary,          \
+                            MemberOnly)                                        \
+  if (Spelling == (OpSpelling))                                                \
+    return Ctx.DeclarationNames.getCXXOperatorName(clang::OO_##OpName);
+#include "clang/Basic/OperatorKinds.def"
+#undef OVERLOADED_OPERATOR
+  return DeclarationName();
+}
+
 std::vector<FuncRef> GetFunctionsUsingName(ConstDeclRef DRef,
                                            const std::string& name) {
   INTEROP_TRACE(DRef, name);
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1468,9 +1468,28 @@ std::vector<TCppFunction_t> GetFunctionsUsingName(TCppScope_t scope,
...
+    llvm::StringRef Spelling = NameRef.trim();
+#define OVERLOADED_OPERATOR(OpName, OpSpelling, Token, Unary, Binary,          \
+                            MemberOnly)                                        \
+  if (Spelling == (OpSpelling))                                                \
+    DName = Ctx.DeclarationNames.getCXXOperatorName(clang::OO_##OpName);
+#include "clang/Basic/OperatorKinds.def"
+  }
+
+  if (!DName)
+    DName = &Ctx.Idents.get(name);
```
vgvassilev, line 1526: Can we use [LookupOperatorName](https://clang.llvm.org/doxygen/classclang_1_1Sema.html#a0c6e994c44767e74e2892aa0235efb1aa6253ec1084459081e77ca3a11d3de567) instead of the string comparisons?

Something similar to:
```
clang/lib/Sema/SemaCoroutine.cpp-829-ExprResult Sema::BuildOperatorCoawaitLookupExpr(Scope *S, SourceLocation Loc) {
clang/lib/Sema/SemaCoroutine.cpp-830-  DeclarationName OpName =
clang/lib/Sema/SemaCoroutine.cpp-831-      Context.DeclarationNames.getCXXOperatorName(OO_Coawait);
clang/lib/Sema/SemaCoroutine.cpp-832-  LookupResult Operators(*this, OpName, SourceLocation(),
clang/lib/Sema/SemaCoroutine.cpp:833:                         Sema::LookupOperatorName);
clang/lib/Sema/SemaCoroutine.cpp-834-  LookupName(Operators, S);

```

guitargeek: Answer from Claude:

Good idea, but `LookupOperatorName` doesn't quite fit here, for two reasons:

1. **It's non-member-only.** That lookup kind resolves to Decl::IDNS_NonMemberOperator (SemaLookup.cpp:233) and explicitly bails out of class-member lookup (`SemaLookup.cpp:2486`); Sema::LookupOverloadedOperatorName documents this as "all member functions are ignored." Member operators (CXXMethodDecl) only carry IDNS_Ordinary — setNonMemberOperator() is applied solely to operators in a non-record context. This entry point needs to find member operators too (e.g. WithOps::operator==, operator[], operator(), which the test covers), so it has to stay on LookupOrdinaryName.
2. **It wouldn't remove the string mapping anyway.** LookupOperatorName is a lookup kind, not a name parser — like the SemaCoroutine example, you still build the DeclarationName from a known OO_* via getCXXOperatorName. We only have the textual spelling (\"operator==\"), and clang exposes no spelling→OverloadedOperatorKind helper (the parser maps operator tokens, not strings). So OperatorKinds.def is the canonical reverse map.

I've pulled the mapping into a small getCXXOperatorDeclName helper with a comment explaining the above, which should make the intent clearer at the call site.

Confirmed empirically: swapping to LookupOperatorName keeps the namespace (free-operator) cases passing but drops all four member-operator assertions (WithOps::operator==, operator+=, operator[], operator()) to 0 matches.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1484,6 +1484,28 @@ void DumpScope(ConstDeclRef DRef) {
...
+          static_cast<unsigned char>(name.front())))
+    return DeclarationName();
+
+  llvm::StringRef Spelling = name.trim();
+#define OVERLOADED_OPERATOR(OpName, OpSpelling, Token, Unary, Binary,          \
+                            MemberOnly)                                        \
+  if (Spelling == (OpSpelling))                                                \
+    return Ctx.DeclarationNames.getCXXOperatorName(clang::OO_##OpName);
+#include "clang/Basic/OperatorKinds.def"
+  return DeclarationName();
```
vgvassilev, line 1508: We probably need an undef here.

vgvassilev (approved):
Lgtm!

lib/CppInterOp/CppInterOp.cpp
vgvassilev: Makes sense. Consider fixing the comments and clang-tidy. In the ci summary you can find the full clang-tidy json and give it to claude for a one shot fix.

lib/CppInterOp/CppInterOp.cpp
vgvassilev: Perhaps to appease codecov this should be some level of llvm_unreachable.

guitargeek: I managed to extend the test to cover all lines!

PR #997, opened 2026-05-12
Desugar template type alias specialisations in wrapper return types

guitargeek:
When generating a JitCall wrapper for a function whose return type is a
template type alias specialisation -- e.g.
`std::enable_if_t<is_constructible_v<...>, std::any>` as on
`std::make_any` -- get_type_as_string left the alias sugared. The
printer's FullyQualifiedName policy applies to types but does not
propagate into expression-level non-type template arguments, so
identifiers like `is_constructible_v` were emitted without their `std::`
qualifier and the wrapper failed to compile with `use of undeclared
identifier`.

Strip TemplateSpecializationType sugar where isTypeAlias() is true
before printing. The desugared form is the alias body with explicit
arguments substituted (e.g. std::any), which compiles cleanly. Regular
typedefs (std::string and friends) are not TemplateSpecializationType
sugar and keep their preferred name.

Reported as cppyy issue
https://github.com/compiler-research/cppyy/issues/218; reproducer is
`cppyy.gbl.std.make_any["MyClass*", "MyClass*"](my_inst)` after
`cppyy.include("any")`.

Adds a regression test built from a stdlib-free distillation rather than
<any>, so it does not depend on libstdc++ internals. The bug needs three
ingredients, all reproduced by hand: (1) a type-alias-template return
type whose sugar carries a non-type argument that prints as an
expression (a variable-template reference), (2) that predicate living in
a namespace so the unqualified print does not resolve in the global-scope
wrapper, and (3) a class (non-builtin) canonical type returned by value,
so get_type_as_string keeps the sugar instead of taking its existing
isBuiltinType() desugaring branch.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -2975,6 +2975,18 @@ void get_type_as_string(QualType QT, std::string& type_name, ASTContext& C,
                         PrintingPolicy Policy) {
   // TODO: Implement cling desugaring from utils::AST
   //       cling::utils::Transform::GetPartiallyDesugaredType()
+  // Desugar template type alias specializations (e.g. std::enable_if_t,
+  // std::remove_cvref_t). Their printed form can carry expression-level
+  // template arguments (variable-template references, SFINAE predicates)
+  // that PrintingPolicy::FullyQualifiedName does not propagate into, so the
+  // emitted text may reference identifiers like `is_constructible_v` without
+  // the `std::` qualifier and fail to compile in the wrapper. Regular
+  // typedefs (e.g. std::string) keep their sugared name.
+  while (const auto* TST = QT->getAs<TemplateSpecializationType>()) {
+    if (!TST->isTypeAlias())
+      break;
+    QT = TST->desugar();
+  }
   if (!QT->isTypedefNameType() || QT->isBuiltinType())
     QT = QT.getDesugaredType(C);
   Policy.Suppress_Elab = true;
--- a/unittests/CppInterOp/FunctionReflectionTest.cpp
+++ b/unittests/CppInterOp/FunctionReflectionTest.cpp
@@ -2431,6 +2431,56 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetFunctionCallWrapper) {
   EXPECT_EQ(f3.getKind(), Cpp::JitCall::Kind::kGenericCall);
 }
 
+TYPED_TEST(CPPINTEROP_TEST_MODE,
+           FunctionReflection_WrapAliasTemplateReturnType) {
+  // Regression test for cppyy issue
+  // https://github.com/compiler-research/cppyy/issues/218 (original reproducer:
+  // `std::make_any<...>`, return type `std::enable_if_t<is_constructible_v<...>,
+  // std::any>`). Building a wrapper for a function template whose return type is
+  // a type-alias-template specialisation used to fail to compile; the snippet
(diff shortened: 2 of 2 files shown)


unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -2431,6 +2431,48 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetFunctionCallWrapper) {
...
+  // SFINAE shape used by libstdc++'s std::any ctors. Hand-rolled trait classes
+  // get folded eagerly, so the test reaches the bug via std::make_any rather
+  // than a stdlib-free construction.
+  if (TypeParam::isOutOfProcess)
+    GTEST_SKIP() << "Test fails for OOP JIT builds";
+
+  std::vector<const char*> interpreter_args = {"-std=c++17", "-include", "new"};
+  TestFixture::CreateInterpreter(interpreter_args);
+  Interp->process(R"(
+    #include <any>
```
vgvassilev: Can we not rely on header files but instead add the smaller underlying reproducer?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2975,6 +2975,18 @@ void get_type_as_string(QualType QT, std::string& type_name, ASTContext& C,
                         PrintingPolicy Policy) {
   // TODO: Implement cling desugaring from utils::AST
   //       cling::utils::Transform::GetPartiallyDesugaredType()
+  // Desugar template type alias specializations (e.g. std::enable_if_t,
+  // std::remove_cvref_t). Their printed form can carry expression-level
+  // template arguments (variable-template references, SFINAE predicates)
+  // that PrintingPolicy::FullyQualifiedName does not propagate into, so the
```
vgvassilev, line 2981: Is that something we should fix upstream too?

unittests/CppInterOp/FunctionReflectionTest.cpp
guitargeek: Yep! the AI did it!

lib/CppInterOp/CppInterOp.cpp
guitargeek: Agreed there's an underlying clang issue - `PrintingPolicy::FullyQualifiedName` isn't honored by `StmtPrinter::VisitDeclRefExpr`, so expression-form non-type template arguments print unqualified. Worth filing upstream (and I can draft the issue/patch). But I'd keep the desugaring here regardless: it's strictly more robust than relying on printing SFINAE predicates into wrapper source, and we need it to work on current released clang.

vgvassilev: I am fine with that but maybe open the issue in clang and tag it with a good first issue label. New contributors might fix it.

vgvassilev (approved):
Lgtm!

lib/CppInterOp/CppInterOp.cpp
guitargeek: Upstream issue opened: https://github.com/llvm/llvm-project/issues/205925

guitargeek: I can't add the label though. Maybe you could do that?

vgvassilev: Done.

vgvassilev:
The wasm failures seem real.

guitargeek:
> The wasm failures seem real.

I have now disabled the test for emscripten, as we did already for many other tests in that file.

PR #1000, opened 2026-05-12
[ci] Read install root from setup-llvm's $LLVM_DIR env.

vgvassilev:
Build_and_Test_CppInterOp's github-hosted branch hardcoded LLVM_DIR="$(pwd)/llvm-project" -- a path the ci-workflows install-tree rename moved to $WS/install/. setup-llvm now exports $LLVM_DIR / $Clang_DIR pointing at the cmake-config dirs (compiler-research/ci-workflows#68). Derive the install root three levels up from $LLVM_DIR when it's set; fall back to the legacy clone path for source-build rows that don't go through setup-llvm. Same fix applied to the Windows native step and to root.yml's cell-layout sanity-check + cmake invocation. The bash local that meant "install root" was also named LLVM_DIR -- it shadowed the env var on second-read paths; renamed to LLVM_PREFIX throughout. self-hosted's branch tracks the same rename for consistency.

Drop CPLUS_INCLUDE_PATH from the native bash and Windows steps. The FIXME documented this as a workaround for missing target-based include propagation; CppInterOp's CMakeLists.txt already calls `include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})` and `${CLANG_INCLUDE_DIRS}` (lines 337-338), making the env-var union redundant. Wasm steps keep their CPLUS_INCLUDE_PATH for now (separate emscripten-side story). ROOT keeps it too: ROOT's own cmake doesn't propagate CLANG_INCLUDE_DIRS, and that's an upstream ROOT bug rather than a workaround we should chase here.

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/.github/actions/Build_and_Test_CppInterOp/action.yml
+++ b/.github/actions/Build_and_Test_CppInterOp/action.yml
@@ -33,51 +33,38 @@ runs:
       shell: bash
       run: |
         if [[ "${{ runner.environment }}" == "self-hosted" ]]; then
-          LLVM_DIR="/usr/lib/llvm-${{ inputs.clang_runtime }}"
-          LLVM_BUILD_DIR="/usr/lib/llvm-${{ inputs.clang_runtime }}"
-          CPLUS_INCLUDE_PATH="/usr/lib/llvm-${{ inputs.clang_runtime }}/include:$PWD/include"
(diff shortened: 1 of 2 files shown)


vgvassilev:
That is superseded by the other PRs.

PR #1002, opened 2026-05-12
add API to work with FunctionProtoType

Vipul-Cariappa:
These functions take `QualType`, when a `Decl` is not available.
For example:
```c++
void f2(int (&f)(int, double)) {}
```
The first argument is a `QualType`, and there is no associated `Decl`.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -125,6 +125,7 @@
 #include <dlfcn.h>
 #include <unistd.h>
 #endif // WIN32
+#include <vector>
 
 //  Runtime symbols required if the library using JIT (Cpp::Evaluate) does
 //  not link to llvm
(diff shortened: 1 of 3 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1556,6 +1556,33 @@ TCppType_t GetFunctionReturnType(TCppFunction_t func) {
...
+TCppIndex_t GetFnTypeNumArgs(TCppType_t fn_type) {
+  INTEROP_TRACE(fn_type);
+  QualType QT = QualType::getFromOpaquePtr(fn_type);
+  const auto* FPT = QT->getAs<clang::FunctionProtoType>();
+  if (!FPT)
+    return INTEROP_RETURN(-1);
+  return INTEROP_RETURN(FPT->getNumParams());
+}
+
+TCppType_t GetFnTypeArgsType(TCppType_t fn_type, TCppIndex_t i) {
```
vgvassilev: Should we squash that into a single interface that refills a std::vector<TCppType_t> where the return type is the first argument?

Vipul-Cariappa: Done.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1555,6 +1557,43 @@ TCppType_t GetFunctionReturnType(TCppFunction_t func) {
...
+}
+
+TCppIndex_t GetFnTypeNumArgs(TCppType_t fn_type) {
+  INTEROP_TRACE(fn_type);
+  QualType QT = QualType::getFromOpaquePtr(fn_type);
+  const auto* FPT = QT->getAs<clang::FunctionProtoType>();
+  if (!FPT)
+    return INTEROP_RETURN(-1);
+  return INTEROP_RETURN(FPT->getNumParams());
+}
```
vgvassilev: These can go now on the client side as convenience functions, right?

Vipul-Cariappa: You still need `GetFnTypeReturnType`. For the `GetFnTypeNumArgs`, yes, the client can handle it. But I believe we can keep it.

vgvassilev (approved):
LGTM!

PR #1004, opened 2026-05-12
add: GetClassTemplateArgs

Vipul-Cariappa:
For the cases where Template Instantiation Arguments are not the same as Template Arguments.

For Example:
```c++
template<typename _Signature>
class function;

template<typename _Res, typename... _ArgTypes>
class function<_Res(_ArgTypes...)> {};

function<int(int)> f;
```

`f`'s Template Arguments: int(int)
`f`'s Template Instantiation Arguments: int, int


vgvassilev (approved):
LGTM!

PR #1013, opened 2026-05-17
[test] Use cmake compatible GTEST_FLAG to set threadsafe death-test

aaronj0:
The previous approach used GTEST_FLAG_SET(death_test_style, ...), a macro introduced in googletest 1.13.0. Some distros (ubuntu-22, alma9) ship gtest 1.11.0, where the macro is undefined and compilation fails. Switch to `::testing::GTEST_FLAG(death_test_style) = value`, available for gtest>1.8, same form used by compiler-rt


vgvassilev (approved):
Lgtm!

PR #1014, opened 2026-05-17
[types] Replace void* opaque handles with distinct struct types.

vgvassilev:
CppInterOp's public handles for Decls, Types, Functions, Objects, and the interpreter were all aliases for void*, so the compiler silently accepted any cross-kind mistake. The IsAbstract(type) crash from the c-layer-rework work was exactly this: a QualType pointer flowed into a function expecting a Decl pointer, and the mismatch only surfaced as a segfault under ARM. Every API taking an opaque handle was one typo away from an unrelated kind, with production crashes as the only test.

Replace each handle with a distinct struct wrapping a single pointer. C++ consumers see namespaced types (Cpp::DeclRef, Cpp::TypeRef, ... plus Const variants for read-only access); C consumers see prefixed global structs (CppDeclRef etc.) -- same layout, so the C-linkage boundary is byte-identical regardless of which view names the type. Cross-kind passes fail at compile time; const widens to the read-only form implicitly, narrowing is rejected. The TableGen-driven C wrapper generation gets typed records too, so cppyy, xeus-cpp and other downstream consumers gain compile-time discipline in both languages.


include/CppInterOp/CppInterOpTypes.h
```diff
@@ -42,6 +42,120 @@
...
+  friend bool operator!=(CppInterp a, CppInterp b) { return !(a == b); }
+#endif
+} CppInterp;
+
+/// Const handle types — read-only access to the underlying AST node.
+/// Implicit widening from mutable to const is allowed; narrowing is not.
+
+typedef struct CppConstDecl {
+  const void* data;
+#ifdef __cplusplus
```
Vipul-Cariappa: Why do we need a const version of the struct? Can the function be marked `const CppDecl` instead of introducing `CppConstDecl`?
Also, cannot this be an anonymous struct? Due to the typedef?

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -100,17 +213,66 @@ typedef struct TemplateArgInfo {
...
+              "Const/mutable handle ABI mismatch");
+static_assert(sizeof(CppFunc) == sizeof(CppConstFunc),
+              "Const/mutable handle ABI mismatch");
+
+// Mutable handles — unwrap to non-const pointer.
+template <typename T> T* unwrap(CppDecl H) { return static_cast<T*>(H.data); }
+template <typename T> T* unwrap(CppType H) { return static_cast<T*>(H.data); }
+template <typename T> T* unwrap(CppFunc H) { return static_cast<T*>(H.data); }
+template <typename T> T* unwrap(CppObject H) { return static_cast<T*>(H.data); }
+template <typename T> T* unwrap(CppInterp H) { return static_cast<T*>(H.data); }
```
Vipul-Cariappa: Should `wrap` and `unwarp` be in the public header? Shouldn't this be in the internal header?

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -251,8 +413,8 @@ enum class ValueKind : std::uint8_t {
 /// function, constructor or destructor.
 class JitCall {
 public:
-  friend CPPINTEROP_API JitCall MakeFunctionCallable(TInterp_t I,
-                                                     TCppConstFunction_t func);
+  friend CPPINTEROP_API JitCall MakeFunctionCallable(CppInterp I,
+                                                     CppConstFunc func);
```
Vipul-Cariappa: Why should the function arguments be updated, when we are `using TInterp_t = CppInterp`?

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -42,6 +42,120 @@
...
+  void* data;
+#ifdef __cplusplus
+  CppDecl() : data(nullptr) {}
+  CppDecl(void* P) : data(P) {}
+  CppDecl(decltype(nullptr)) : data(nullptr) {}
+  explicit operator bool() const { return data != nullptr; }
+  friend bool operator==(CppDecl a, CppDecl b) { return a.data == b.data; }
+  friend bool operator!=(CppDecl a, CppDecl b) { return !(a == b); }
+#endif
+} CppDecl;
```
Vipul-Cariappa: I believe `data` should be a `private` member.
Why don't we add the following conversion operator:
```c++
template <typename T,
              typename = std::enable_if_t<std::is_base_of_v<clang::Decl, T>>>
    operator T*() {
        return static_cast<T*>(data); // we will first cast to clang::Decl* then do a llvm::dyn_cast here
    }
```
Then the casts will become implicit and reduce code size. Look at https://godbolt.org/z/bbnr9Gs8Y. Maybe restrict `T=clang::Decl` only for better type safety. But maybe it is not possible to do this without leaking the `clang::Decl` symbol? But I can think of ways to be it...

Vipul-Cariappa (changes requested):
I do not understand some of the reasoning behind some of these...

include/CppInterOp/CppInterOpTypes.h
vgvassilev: Private `data` will make the c layer harder for a small gain. The auto conversion is exactly what we want to avoid, that's why we are moving away from `void*`.

include/CppInterOp/CppInterOpTypes.h
vgvassilev: I thought that would make the cppyy transition easier but you are right, let's drop all of these `using` shadows.

include/CppInterOp/CppInterOpTypes.h
vgvassilev: That would make the class const, not the underlying pointer.

include/CppInterOp/CppInterOpTypes.h
vgvassilev: Done.

include/CppInterOp/CppInterOpTypes.h
Vipul-Cariappa: I meant the opposite.
We should keep the `using TInterp_t = CppInterp` and **not** change the signature of all the functions.

vgvassilev: I disagree -- they have the wrong `T` prefix, inconsistent `_t` suffix. This is a good opportunity to clean things up.

Vipul-Cariappa (approved):
LGTM!
But let's merge this on Monday after lunch. Around 1 or 2 PM CET. Because this will break CI of other repos.

vgvassilev:
> LGTM! But let's merge this on Monday after lunch. Around 1 or 2 PM CET. Because this will break CI of other repos.

In that case the merging is on you.

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -100,17 +245,21 @@ typedef struct TemplateArgInfo {
 } TemplateArgInfo;
```
Vipul-Cariappa, line 266: Hi @vgvassilev, these types are not in the `Cpp` namespace, nor within the `CppImpl` namespace. They are global types. Was this intended?

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -100,17 +245,21 @@ typedef struct TemplateArgInfo {
...
-using TInterp_t = void*;
-using TCppObject_t = void*;
+namespace Cpp {
+
+// Pull file-scope C-compatible structs into the namespace.
+// The handle types (DeclRef etc.) are already defined in this namespace
+// above; only the genuinely-C-visible structs need pulling in.
+using ::CppInterOpArray;
+using ::CppInterOpStringArray;
+using ::TemplateArgInfo;
```
Vipul-Cariappa, line 276: This reintroduces it back in the `Cpp` namespace.

> This is causing some linker related errors. As the mangled names of some functions are different. Discovered from cppyy.

include/CppInterOp/CppInterOpTypes.h
vgvassilev: Yes, they work for the C layer too...

include/CppInterOp/CppInterOpTypes.h
vgvassilev: What errors do you get?

Vipul-Cariappa: We redefine the at
https://github.com/compiler-research/CPyCppyy/blob/b91ef2232a4f888cb63699940e28a423a032229b/src/Cppyy.h#L46-L52

Vipul-Cariappa: Shouldn't be a problem after the move to monorepo.

vgvassilev: Ok, sure, that needs to be dropped.

PR #1015, opened 2026-05-27
Add MakeVTableOverlay for runtime virtual-slot override

vgvassilev:
Language bindings (cppyy, and prospectively ROOT) need to redirect selected virtual methods of a live C++ object to a foreign implementation -- e.g. a Python override -- without recompiling the class or its callers. MakeVTableOverlay copies the object's vtable, overwrites the chosen slots, and repoints the vptr at the copy; the returned handle remembers the original vptr and restores it on destruction. Non-overlaid slots and the ABI prefix (offset-to-top and type_info on Itanium) are preserved, so the destructor, RTTI and unrelated virtuals keep dispatching normally.

The public entry point works in reflected methods rather than ABI slot indices: callers pass the base scope and the virtual methods to replace, and the slot index and vtable size are derived from Clang's VTableContext, correct for both the Itanium and Microsoft ABIs. The reflection-to-slot translation and the raw pointer surgery both live in CppInterOp.cpp, with the slot/size queries and the vtable copy kept as file-local helpers rather than public API; only MakeVTableOverlay and DestroyVTableOverlay are exported. A header-only UniqueVTableOverlay wraps the handle in unique_ptr for RAII install/restore.

The unit tests construct a reflected, JIT-built object and exercise slot replacement, prefix and unrelated-slot preservation, restore-on-destroy, multi-slot overlay and input validation, dispatching through the installed slot via a function pointer so the compiler cannot devirtualize past the overlay. A cross-TU benchmark drives the dispatch loop from a separate shared library -- where the call site cannot be devirtualized -- to assert, in Release non-sanitizer builds, that overlaying a slot adds no per-call cost over an ordinary virtual call. Google Benchmark is fetched as an ExternalProject the same way as googletest, so the perf check builds and runs as part of the default test suite.

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1993,6 +1995,119 @@ bool IsVirtualMethod(TCppFunction_t method) {
...
+}
+
+VTableOverlay* MakeVTableOverlay(void* inst, TCppScope_t base,
+                                 const TCppConstFunction_t* methods,
+                                 void* const* overlay_fns,
+                                 std::size_t n_overlays) {
+  INTEROP_TRACE(inst, base, methods, overlay_fns, n_overlays);
+  int total_method_slots = vtableMethodSlotCount(base);
+  if (!inst || total_method_slots < 2)
+    return INTEROP_RETURN(nullptr);
```
Vipul-Cariappa, line 2135: Why can't `total_method_slots = 1`? Maybe something to do with the compiler doing stuff. Can we get a comment.
```c++
struct A {
  virtual ~A() {}
};
```
The above should have only 1 slot, right?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1993,6 +1995,119 @@ bool IsVirtualMethod(TCppFunction_t method) {
...
+
+  void** orig_vptr = *reinterpret_cast<void***>(inst);
+  void** block = new void*[total];
+  std::memcpy(block, orig_vptr - kPrefix, total * sizeof(void*));
+
+  for (std::size_t i = 0; i < n; ++i)
+    block[kPrefix + slots[i]] = fns[i];
+
+  *reinterpret_cast<void***>(inst) = block + kPrefix;
+  return new VTableOverlay{block, reinterpret_cast<void*>(orig_vptr), inst};
```
Vipul-Cariappa, line 2113: Does this globally change the vtable for all current and future objects (until `DestroyVTableOverlay` is called) of type, or only for this object, `inst`?
From my understanding, it is only changing the vtable for this object, not globally.
A comment on this would be nice.

unittests/CppInterOp/VTableOverlayCrossTUBench.cpp
```diff
@@ -0,0 +1,95 @@
...
+  for (auto _ : state)
+    benchmark::DoNotOptimize(OverlayDispatchOnce(&impl, 7));
+}
+BENCHMARK(BM_XTU_OverlayDirectFn);
+
+TEST(VTableOverlayCrossTU, OverlayAddsNoPerCallCost) {
+#if !defined(NDEBUG) || defined(__SANITIZE_ADDRESS__)
+  GTEST_SKIP() << "Perf assertions need a Release, non-sanitizer build.";
+#endif
+  EXPECT_NOT_SLOWER_THAN(BM_XTU_OverlayDirectFn, BM_XTU_BareVirtual);
```
Vipul-Cariappa, line 92: From my understanding of this PR:
This is not devirtualizing the call. Instead, it is swapping the implementation to something else at runtime. I don't understand why swapping should make it faster?
Can this be explained?

unittests/CppInterOp/VTableOverlayTest.cpp
```diff
@@ -0,0 +1,141 @@
...
+// has a single deleting-dtor slot, so the user-virtual indices differ.
+Cpp::TCppScope_t DeclareBase() {
+  // -include new: Construct's wrapper uses placement new, so every
+  // interpreter compilation must see <new>.
+  Cpp::CreateInterpreter({"-include", "new"});
+  Cpp::Declare("struct OverlayB {"
+               "  virtual ~OverlayB() {}"
+               "  virtual int alpha(int x) { return x + 10; }"
+               "  virtual int beta(int x) { return x + 20; }"
+               "};");
```
Vipul-Cariappa, line 69: Can we get the following combination of tests:
- Simple Inheritance with & without overriding the virtual functions
- Multi-level inheritance
  Ex:
```c++
struct A { ... };
struct B : A { ... };
struct C : B { ... };
```
- Multiple inheritance
  Ex:
```c++
struct A { ... };
struct B { ... };
struct C : A, B { ... };
```
- virtual inheritance
  Ex:
```c++
struct A { ... };
struct B : virtual A { ... };
```

The current struct does not have any attributes; we need attributes too.
The overlayed function does not use `this` pointer; we need tests that access `this` pointer.

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: Depending on the actual behaviour, the vtable is changed globally or not. Can we get tests for that? For example, we have two objects, one object we overlay, and check if the other object's vtable is also modified, or not.

Vipul-Cariappa (commented):
Can we get a single-line (not too verbose) comment on what each test is testing?

LGTM! A few comments on more tests, and some minor clarification.

Vipul-Cariappa:
Does this account for "virtual thunk"?
Ref: https://stackoverflow.com/a/39182009
Example: https://godbolt.org/z/c56Mv8xbe

We will need tests.

lib/CppInterOp/CppInterOp.cpp
vgvassilev: No, that changes the vtable per instance `inst`.

Vipul-Cariappa (commented):
Only comment about one more test for the following:
```c++
struct A {
  int m_x;
  A(int x) : m_x(x) {}
  virtual int method() { return m_x; } // replace this function with foo given below
};

struct B : A {
  double m_d;
  B(int x, double d) : A(x), m_d(d) {}
  int method() override { return (int)m_d; } // replace this function with bar given below
};
```
```c++
int foo(void* self) {
  return (static_cast<A*>(self))->m_x + 10;
}

int bar(void* self) {
  return (int)((static_cast<B*>(self))->m_x + (static_cast<B*>(self))->m_d);
}
```
Initialise `A` & `B` with non-zero values.

Otherwise, everything else is good.

Vipul-Cariappa (approved):
LGTM!

Issue #1016, opened 2026-05-27
[Bug]: GetClassDecls returns empty for template specializations


keremsahn:
### Description of bug

Because of clang's lazy instantiation klass->hasDefinition() returns false and implicit members are not created for template specializations.
From GetClassDecl function:
```cpp
  if (CXXRD->hasDefinition())
    CXXRD = CXXRD->getDefinition();
  getSema().ForceDeclarationOfImplicitMembers(CXXRD);
```
Test:
```cpp
TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetClassDecls2) {
  std::vector<Decl*> Decls, SubDecls;
  std::string code = R"(template<typename T>
  class Klass {
    T value;
  };
 using KlassInt = Klass<int>;)";

  GetAllTopLevelDecls(code, Decls);

  std::vector<Cpp::TCppFunction_t> methods;
  Cpp::GetClassMethods(Decls[1], methods);

  EXPECT_EQ(methods.size(), 10);
}
```
this test fails because of this bug with this output:

```cpp
[ RUN      ] CppInterOpTest/InProcessJIT.FunctionReflection_GetClassDecls2
/home/keremsahn/CppInterOp/unittests/CppInterOp/FunctionReflectionTest.cpp:352: 
Failure
Expected equality of these values:
  methods.size()
    Which is: 0
10
```

### What operating system was you using when the bug occured?

Ubuntu

### What is the architechture of the cpu on your system?

X86

keremsahn:
@aaronj0 I will create a pr tomorrow to solve that via using RequireCompleteType

keremsahn:
This is solved

PR #1018, opened 2026-05-28
Force instantiation for template specializations

keremsahn:
Because of lazy instantiation for template specializations 
```cpp
if (CXXRD->hasDefinition())
``` 
was returning false and GetClassDecls function was not working as expected. So RequireCompleteType function was used to force instantiation on template specalizations

links to #1016


keremsahn:
@aaronj0

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1408,6 +1406,18 @@ static void GetClassDecls(TCppScope_t klass,
...
+  if (auto* CTSD = dyn_cast<ClassTemplateSpecializationDecl>(CXXRD)) {
+    QualType QT;
+#if CLANG_VERSION_MAJOR < 22
+    QT = QualType(CTSD->getTypeForDecl(), 0);
+#else
+    QT = getSema().getASTContext().getCanonicalTagType(CTSD);
+#endif
+    if (getSema().RequireCompleteType(CTSD->getLocation(), QT,
+                                      diag::err_incomplete_type))
+      return; // Unsuccesfull instantiaton
```
Vipul-Cariappa: I believe you need `SynthesizingCodeRAII` before the call to `RequireCompleteType`.
Look at other uses of `RequireCompleteType` in the file.

Vipul-Cariappa: Please rebase the PR against master.

keremsahn:
```cpp

  auto* CXXRD = dyn_cast<CXXRecordDecl>(D);
  compat::SynthesizingCodeRAII RAII(&getInterp());
  if (auto* CTSD = dyn_cast<ClassTemplateSpecializationDecl>(CXXRD)) {
    QualType QT;
#if CLANG_VERSION_MAJOR < 22
    QT = QualType(CTSD->getTypeForDecl(), 0);
#else
    QT = getSema().getASTContext().getCanonicalTagType(CTSD);
#endif
```

Yeah it is actually called before

Vipul-Cariappa: Hmm. And the cppyy tests still fail. That is weird. I will try to check it locally.

Vipul-Cariappa (approved):
LGTM!
Thank you!

PR #1019, opened 2026-05-31
Add per-instance state lifecycle to VTableOverlay

vgvassilev:
Bindings layering on top of MakeVTableOverlay -- cppyy's Python handlers, a Lua binding's closures, RooFit's per-object trampolines -- need two things from the overlay primitive: a place to stash per-instance state adjacent to the vtable so a thunk reads it with a fixed-offset load, and a tear-down moment tied to the C++ object's lifetime so the state can be released. Without the first they pay a hash + mutex per dispatch through a process-global handler map (BM_XTU_OverlayGlobalMap baseline at ~17 ns/call); without the second they either leak the state (refcounts on the host wrapper pin it open and deadlock the GC) or thread cleanup through every potential destruction path -- a temporary expiring at semicolon, container element drop, exception unwind, shared_ptr cycle break.

MakeVTableOverlay grows three new optional parameters. n_extra_prefix_slots reserves nullptr-initialized void* cells before the ABI prefix in the owned block; thunks reach them through the inline VTableOverlayExtraSlot(self, i) helper -- one fixed-offset load, no registry. on_destroy and cleanup_data install a per-instance trampoline at the deleting-dtor slot (D0 on Itanium, the single deleting dtor on MSVC) that runs the original destructor first, then the cleanup callback. Original-first ordering keeps virtuals dispatched from inside ~T()'s body resolving through the still- installed overlay; the callback runs only when no further virtuals can fire. The callback must not throw -- it runs inside a C++ destructor, so an escaping exception is undefined. Per-instance dtor-hook metadata lives on the active CppImpl::Interpreter (mirroring https://github.com/compiler-research/CppInterOp/pull/900's DLM move) and the wrapper walks the interpreter list at fire time, so the orig_dtor JIT pointer never outlives its emitter and overlay blocks carry no routing slot. DestroyVTableOverlay is idempotent against both directions of the race: an internal dtor_fired flag suppresses the vptr restore on a dead instance, and caller-driven teardown drops the matching hook entry across all interpreters before restoring.

Five new unit tests pin the contract. SetsExtraPrefixSlots verifies the slots are nullptr-initialized + writable and that overlaid dispatch still flows through the new layout. DestructorHookFires (callback runs once after operator delete with the original inst), DestructorHookIsPerInterpreter (two interpreters, no cross-map bleed), DestructorHookOptInZero (with on_destroy=nullptr the deleting-dtor slot is the verbatim copy from the original vtable), and DestructorHookSkippedOnCallerTeardown (race where teardown beats the wrapper, callback never fires) pin the four corners of the dtor-hook contract. Two new bench cells track dispatch-path cost. BM_XTU_OverlayExtraSlot at ~3.9 ns/call asserts >= 2x faster than BM_XTU_OverlayGlobalMap, the registry-lookup alternative. BM_XTU_OverlayWithDtorHook prints its per-call cost for trend visibility -- the wrapper patches a different vtable slot from the dispatched method's, so cost is structurally unchanged, but at sub-10 ns precision macOS CI jitter dominates any relative threshold. Both features are opt-in: n_extra_prefix_slots = 0 and on_destroy = nullptr (the defaults) preserve the prior behavior bit for bit.

### DIFF
--- a/include/CppInterOp/CppInterOpThunks.h
+++ b/include/CppInterOp/CppInterOpThunks.h
@@ -1,0 +1,53 @@
+//===--- CppInterOpThunks.h - Variadic dispatcher template ------*- C++ -*-===//
+//
+// Part of the compiler-research project, under the Apache License v2.0 with
+// LLVM Exceptions.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Header-only variadic-template dispatcher that bindings instantiate per
+// signature × slot to populate a precompiled thunk catalog. CppInterOp ships
+// the template body; the binding contributes the per-instance handler lookup
+// and the language-specific marshaling.
+//
+// The binding provides a HandlerTraits struct of this shape:
+//
+//   struct HandlerTraits {
+//     using Handler = /* binding-specific per-instance state */;
+//
+//     // Per-instance lookup -- typically reads the handler pointer from an
+//     // extra prefix slot reserved via MakeVTableOverlay's
+//     // n_extra_prefix_slots, i.e. vptr[-(kVTableOverlayPrefixSize + 1)].
+//     static Handler& From(void* self);
+//
+//     // Marshal Args... into the target language, call the bound callable
+//     // at index Slot, and marshal the result back as R.
+//     template <class R, class... Args>
+//     static R Call(Handler&, std::size_t Slot, Args... args);
+//   };
+//
+// Each instantiation of dispatch<T, Slot, R, Args...> produces a concrete
+// function with the right calling convention to install in a patched vtable
+// slot. The catalog is a per-binding table of such instantiations indexed by
+// (signature shape, slot).
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CPPINTEROP_CPPINTEROPTHUNKS_H
+#define CPPINTEROP_CPPINTEROPTHUNKS_H
+
+#include <cstddef>
+
+namespace CppImpl {
+namespace Thunks {
+
+template <class T, std::size_t Slot, class R, class... Args>
+R dispatch(void* self, Args... args) {
+  return T::template Call<R, Args...>(T::From(self), Slot, args...);
+}
+
+} // namespace Thunks
+} // namespace CppImpl
+
+#endif // CPPINTEROP_CPPINTEROPTHUNKS_H
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -2082,17 +2082,106 @@ static bool hasComplexVTableLayout(const CXXRecordDecl* RD) {
   return false;
 }
 
+// ABI-only prefix size (excludes the hidden self-pointer slot CppInterOp
+// interposes for dtor-hook routing). Itanium has 2 slots (offset-to-top,
+// type_info); MSVC has 1 (complete-object-locator).
+#ifdef _WIN32
+constexpr int kABIPrefixSize = 1;
+#else
+constexpr int kABIPrefixSize = 2;
+#endif
+static_assert(detail::kVTableOverlayPrefixSize == kABIPrefixSize + 1,
+              "public prefix size must equal ABI prefix + 1 hidden slot");
+
+// Single locus for vptr reads/writes and slot arithmetic in this file;
+// the public header's VTableOverlayExtraSlot covers the symmetric pun
+// thunks need on the read side. The owned block layout is:
+//
+//   [ user extras (N) ] [ hidden self-ptr ] [ ABI prefix ] [ methods ]
+//                                          ^               ^
+//                            address_point - kVTableOverlayPrefixSize
+//                                                          address_point
+//
+// The wrapper at the deleting-dtor slot recovers its VTableOverlay from
+// the hidden self-ptr slot via a fixed-offset load from `self`'s vptr.
 struct VTableOverlay {
-  void** block;        // owned copy; the published vtable is block + kPrefix
-  void* original_vptr; // restored by DestroyVTableOverlay
-  void* inst;          // object whose vptr was replaced
+  void** block;         // owned, freed in ~VTableOverlay
+  void** original_vptr; // restored on caller-driven teardown
+  void* inst;           // object whose vptr was replaced
+  std::size_t n_extra_prefix_slots;
+  bool dtor_fired = false; // wrapper started -- skip vptr restore
+  // Dtor-hook fields. orig_dtor stays null when the caller passed
+  // on_destroy = nullptr; the wrapper is then not installed at all.
+  void (*orig_dtor)(void*) = nullptr;
+  VTableOverlayDtorHook cleanup = nullptr;
+  void* cleanup_data = nullptr;
+
+  VTableOverlay(void** block, void** orig_vptr, void* inst,
+                std::size_t n_extra)
+      : block(block), original_vptr(orig_vptr), inst(inst),
+        n_extra_prefix_slots(n_extra) {
+    // Stash self-pointer in the hidden slot before publishing the vptr;
+    // the wrapper reads it at fire time via a fixed offset from vptr.
+    *hidden_slot() = this;
+    WriteVPtr(inst, address_point());
+  }
+  ~VTableOverlay() {
+    if (!dtor_fired)
+      WriteVPtr(inst, original_vptr);
+    delete[] block;
+  }
+  VTableOverlay(const VTableOverlay&) = delete;
+  VTableOverlay& operator=(const VTableOverlay&) = delete;
+
+  void** address_point() const {
+    return block + n_extra_prefix_slots + detail::kVTableOverlayPrefixSize;
+  }
+  VTableOverlay** hidden_slot() const {
+    return reinterpret_cast<VTableOverlay**>(block + n_extra_prefix_slots);
+  }
+
+  // reinterpret_cast between function and void* is only conditionally
+  // supported per [expr.reinterpret.cast]/6; memcpy is the well-defined
+  // alternative on every platform CppInterOp targets.
+  template <class To, class From> static To BitCastFn(From f) noexcept {
+    static_assert(sizeof(To) == sizeof(From));
+    To to;
+    std::memcpy(&to, &f, sizeof(to));
+    return to;
+  }
+
+  static void** ReadVPtr(void* inst) {
+    return *reinterpret_cast<void***>(inst);
+  }
+  static void WriteVPtr(void* inst, void** new_vptr) {
+    *reinterpret_cast<void***>(inst) = new_vptr;
+  }
 };
+
+// Wrapper installed in the deleting-dtor slot when MakeVTableOverlay was
+// called with a non-null on_destroy. Recovers the owning VTableOverlay
+// from `self`'s vptr (hidden-slot fixed-offset load) and runs the
+// callback BEFORE the original destructor: `self` is alive at that
+// point so the callback can inspect it, and after the original
+// deleting-destructor returns memory has been freed.
+extern "C" void cppinteropVTableOverlayDtorWrapper(void* self) {
+  void** vptr = VTableOverlay::ReadVPtr(self);
+  VTableOverlay* ov = *reinterpret_cast<VTableOverlay**>(
+      vptr - detail::kVTableOverlayPrefixSize);
+  // Snapshot orig_dtor before user code runs: a misbehaving callback
+  // that destroys the overlay must not strand the C++ destructor.
+  auto orig_dtor = ov->orig_dtor;
+  ov->dtor_fired = true;
+  if (ov->cleanup)
+    ov->cleanup(self, ov->cleanup_data);
+  orig_dtor(self);
+}
 
 // Minimum slot count a polymorphic class can have from its address point:
 // Itanium emits the destructor pair (D1 + D0) so the count is at least 2;
 // Microsoft emits a single deleting-dtor slot so the count is at least 1.
 // vtableMethodSlotCount returns -1 for non-polymorphic scopes.
-#if defined(_WIN32)
+#ifdef _WIN32
 constexpr int kMinVTableMethodSlots = 1;
 #else
 constexpr int kMinVTableMethodSlots = 2;
@@ -2101,9 +2190,13 @@ constexpr int kMinVTableMethodSlots = 2;
 // Reflection-free pointer surgery: copy inst's vtable, overwrite slots with
 // fns, install the copy and return a handle owning it. The slot indices and
 // count are resolved from reflection by the caller (MakeVTableOverlay).
+// n_extra_prefix_slots prepends nullptr-initialized void* slots before the
+// ABI prefix; the caller stashes per-instance data there and thunks read it
+// via vptr[-(kPrefix + 1 + i)].
 static VTableOverlay* applyVTableOverlay(void* inst, int total_method_slots,
                                          const int* slots, void* const* fns,
-                                         std::size_t n) {
+                                         std::size_t n,
+                                         std::size_t n_extra_prefix_slots) {
   if (!inst || total_method_slots < kMinVTableMethodSlots)
     return nullptr;
   for (std::size_t i = 0; i < n; ++i) {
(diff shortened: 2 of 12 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2142,18 +2264,34 @@ VTableOverlay* MakeVTableOverlay(void* inst, TCppScope_t base,
...
+    void** vptr = ov->address_point();
+    auto orig_dtor =
+        VTableOverlay::BitCastFn<void (*)(void*)>(vptr[kDeletingDtorSlot]);
+    vptr[kDeletingDtorSlot] =
+        VTableOverlay::BitCastFn<void*>(&cppinteropVTableOverlayDtorWrapper);
+    auto& info = getInterpInfo();
+    std::lock_guard<std::mutex> lk(info.DtorHooksMu);
+    info.DtorHooks.insert(
+        {inst, VTableOverlayDtorHookEntry{orig_dtor, on_destroy,
+                                          cleanup_data, ov}});
```
Vipul-Cariappa: You make an assumption that the interpreter to use is the top of the stack. But in `cppinteropVTableOverlayDtorWrapper` you iterate over all interpreters and don't make this assumption. Why?

lib/CppInterOp/CppInterOp.td
```diff
@@ -1517,6 +1517,23 @@ overlay does not touch and is rejected outright -- see \returns.
 \param[in] methods Virtual methods of \c base whose slots are replaced.
 \param[in] overlay_fns Function pointers to install at those slots.
 \param[in] n_overlays Length of \c methods and \c overlay_fns.
+\param[in] n_extra_prefix_slots Optional \c void* slots prepended *before* the
+           ABI prefix in the owned block, initialized to nullptr. Bindings use
+           them to stash per-instance data adjacent to the vtable -- a thunk
+           reads its slot via \c VTableOverlayExtraSlot(self, i) with a single
+           fixed-offset load, avoiding a runtime registry lookup.
+\param[in] on_destroy Optional callback invoked once after the C++ destructor
+           of \c inst runs (operator-delete path: D0 on Itanium, the single
```
Vipul-Cariappa: Why after? Why not before the destruction of the actual object?
I believe it should be before.

unittests/CppInterOp/VTableOverlayCrossTUBench.cpp
```diff
@@ -92,8 +95,134 @@ TEST(VTableOverlayCrossTU, OverlayAddsNoPerCallCost) {
...
+}
+
+// Bench-only stand-in for the naive alternative: a process-global registry
+// keyed by `void* self`. The production library does NOT use a global mutex
+// for per-instance state -- that's the whole point of n_extra_prefix_slots.
+std::unordered_map<void*, Handler*>& NaiveHandlerMap() {
+  static std::unordered_map<void*, Handler*> m;
+  return m;
+}
+std::mutex& NaiveHandlerMapMu() {
```
Vipul-Cariappa:
```suggestion
std::mutex& NaiveHandlerMapMutex() {
```

unittests/CppInterOp/VTableOverlayCrossTUBench.cpp
```diff
@@ -92,8 +95,134 @@ TEST(VTableOverlayCrossTU, OverlayAddsNoPerCallCost) {
...
+  return h->add(x);
+}
+
+// Bench-only stand-in for the naive alternative: a process-global registry
+// keyed by `void* self`. The production library does NOT use a global mutex
+// for per-instance state -- that's the whole point of n_extra_prefix_slots.
+std::unordered_map<void*, Handler*>& NaiveHandlerMap() {
+  static std::unordered_map<void*, Handler*> m;
+  return m;
+}
```
Vipul-Cariappa: Why does this need to be guarded by a function? Why can't `m` just be a global static variable? Similarly, for the mutex below.

unittests/CppInterOp/VTableOverlayCrossTUBench.cpp
```diff
@@ -92,8 +95,134 @@ TEST(VTableOverlayCrossTU, OverlayAddsNoPerCallCost) {
...
+
+TEST(VTableOverlayCrossTU, ExtraSlotBeatsGlobalMap) {
+#if !defined(NDEBUG) || defined(__SANITIZE_ADDRESS__)
+  GTEST_SKIP() << "Perf assertions need a Release, non-sanitizer build.";
+#endif
+  // The map path is hash + mutex per dispatch; the slot path is one mov.
+  // Even with a fast hash the gap is real; require >= 2x to ride out jitter.
+  EXPECT_AT_LEAST_N_TIMES_FASTER(BM_XTU_OverlayExtraSlot,
+                                 BM_XTU_OverlayGlobalMap, 2.0);
+}
```
Vipul-Cariappa, line 186: How is this benchmark relevant? And what is this benchmarking?
The `xtu_thunk_extra_slot` is a simple function, whereas `xtu_thunk_global_map` has a mutex and hash maps. It is definitely expected behaviour.
Both take the same code path until this function is hit.

unittests/CppInterOp/VTableOverlayCrossTUBench.cpp
```diff
@@ -92,8 +95,134 @@ TEST(VTableOverlayCrossTU, OverlayAddsNoPerCallCost) {
...
+  auto* ov = Cpp::MakeVTableOverlay(
+      &impl, base, &method, &fn, /*n=*/1,
+      /*n_extra_prefix_slots=*/0,
+      /*on_destroy=*/&xtu_noop_cleanup,
+      /*cleanup_data=*/nullptr);
+  for (auto _ : state)
+    benchmark::DoNotOptimize(OverlayDispatchOnce(&impl, 7));
+  Cpp::DestroyVTableOverlay(ov);
+}
+BENCHMARK(BM_XTU_OverlayWithDtorHook);
```
Vipul-Cariappa, line 212: You are benchmarking this, but don't `EXPECT_AT_LEAST_N_TIMES_FASTER`, why?

unittests/CppInterOp/VTableOverlayTest.cpp
```diff
@@ -430,8 +431,160 @@ TEST(VTableOverlay, RejectsVirtualInheritance) {
...
+  auto* B = DeclareBase();
+  void* inst = Cpp::Construct(B);
+  ASSERT_NE(inst, nullptr);
+
+  struct State { int fires = 0; void* last_inst = nullptr; };
+  State state;
+  auto* ov = Cpp::MakeVTableOverlay(
+      inst, B, /*methods=*/nullptr, /*overlay_fns=*/nullptr,
+      /*n_overlays=*/0, /*n_extra_prefix_slots=*/0,
+      /*on_destroy=*/+[](void* i, void* data) {
```
Vipul-Cariappa:
```suggestion
      /*on_destroy=*/[](void* i, void* data) {
```
Is that `+` required? If yes, why?

unittests/CppInterOp/VTableOverlayTest.cpp
```diff
@@ -430,8 +431,160 @@ TEST(VTableOverlay, RejectsVirtualInheritance) {
...
+// finds nothing and the callback never fires.
+TEST(VTableOverlay, DestructorHookSkippedOnCallerTeardown) {
+  auto* B = DeclareBase();
+  void* inst = Cpp::Construct(B);
+  ASSERT_NE(inst, nullptr);
+
+  int fires = 0;
+  auto* ov = Cpp::MakeVTableOverlay(
+      inst, B, nullptr, nullptr, 0, 0,
+      +[](void* /*i*/, void* data) { *static_cast<int*>(data) += 1; },
```
Vipul-Cariappa:
```suggestion
      [](void* /*i*/, void* data) { *static_cast<int*>(data) += 1; },
```
?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -170,6 +181,10 @@ struct InterpreterInfo {
   // interpreter, so the caches must be destroyed together with it.
   std::map<const FunctionDecl*, void*> WrapperStore;
   std::map<const Decl*, void*> DtorWrapperStore;
+  // VTableOverlay destructor-hook registry; see the dtor wrapper for the
+  // routing protocol. Mutex covers concurrent install/fire/teardown.
+  std::mutex DtorHooksMu;
+  std::unordered_map<void*, VTableOverlayDtorHookEntry> DtorHooks;
```
Vipul-Cariappa: Why do we need the mutex?
To me, it does not look like CppInterOp needs to provide thread-safety for execution of the destructors. The user should make sure that the code is thread-safe. The only reason you might need a mutex here is if the `cleanup_data` is shared or the destructor is changing a global variable. Both are the responsibility of the user.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2059,17 +2074,110 @@ static bool hasComplexVTableLayout(const CXXRecordDecl* RD) {
...
+  for (auto& info : GetInterpreters()) {
+    std::lock_guard<std::mutex> lk(info.DtorHooksMu);
+    auto it = info.DtorHooks.find(self);
+    if (it == info.DtorHooks.end())
+      continue;
+    entry = it->second;
+    info.DtorHooks.erase(it);
+    found = true;
+    break;
+  }
```
Vipul-Cariappa: Is it possible to remove this loop somehow?

Vipul-Cariappa (commented):
My understanding of this PR is as follows:
We introduce a mechanism to store some additional data on the vtable of the instance. These data are `n` `void*`. We also add an optional destructor overload. A custom destructor that the user wants to call after the call to the actual C++ destructor (I believe the custom dtor should be called before the actual dtor). We also add an argument to the overloaded destructor: `void*` for any additional data.
Please correct me if my understanding of things is wrong.

The commit message is horrible. Does not tell what is being done or why. Please make sure the commit message is firstly human readable. I also believe the comments in the source file are too verbose. But maybe it is required if someone needs to understand this code. But there are repetitive comments; these can be removed.

I cannot guarantee that the pointer arithmetic magic being does is correct; I expect there to be sufficient tests to catch any errors in it. FYI, I could not think of any edge cases that the current tests don't already cover.

unittests/CppInterOp/VTableOverlayCrossTUBench.cpp
vgvassilev: Done.

unittests/CppInterOp/VTableOverlayCrossTUBench.cpp
vgvassilev: Added some comment to explain this.

unittests/CppInterOp/VTableOverlayTest.cpp
vgvassilev: Normally that forces conversion but here we don't need it.

lib/CppInterOp/CppInterOp.cpp
vgvassilev: Added some clarification comment. It could but we shouldn't do it but unstead make GetInterpreters() inlineable and very fast for N==1. Right now it is not really painful when it becomes a pain we have a way forward described in a comment.

lib/CppInterOp/CppInterOp.cpp
vgvassilev: We are providing a safe context in which the user can provide their user code. Updated the comment.

include/CppInterOp/CppInterOpThunks.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include <cstddef>
+
+namespace CppImpl {
+namespace Thunks {
+
+template <class T, std::size_t Slot, class R, class... Args>
+R dispatch(void* self, Args... args) {
+  return T::template Call<R, Args...>(T::From(self), Slot, args...);
+}
```
Vipul-Cariappa, line 48: I see this file, but I don't see it being used anywhere. Can we get tests for this? We should have at least one test as an example of how to use it. When I don't know how to use an API, I look at the tests to figure it out.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -170,6 +181,13 @@ struct InterpreterInfo {
   // interpreter, so the caches must be destroyed together with it.
   std::map<const FunctionDecl*, void*> WrapperStore;
   std::map<const Decl*, void*> DtorWrapperStore;
+  // VTableOverlay destructor-hook registry. Mutex guards the map's own
+  // structure -- concurrent MakeVTableOverlay / DestroyVTableOverlay /
+  // dtor-wrapper-fire on the same interpreter race on insert/erase/find,
+  // so the lookups themselves must serialize. Thread safety of the user's
+  // on_destroy body and cleanup_data is the caller's responsibility.
+  std::mutex DtorHooksMu;
+  std::unordered_map<void*, VTableOverlayDtorHookEntry> DtorHooks;
```
Vipul-Cariappa: I believe `VTableOverlayDtorHookEntry` and `VTableOverlay` can be merged into the same struct. Looking at the usage of `dtor_fired`: the lifetimes of both the objects are the same. Unless, I am wrong, please let me know.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2070,17 +2077,110 @@ static bool hasComplexVTableLayout(const CXXRecordDecl* RD) {
...
+  for (auto& info : GetInterpreters()) {
+    std::lock_guard<std::mutex> lk(info.DtorHooksMu);
+    auto it = info.DtorHooks.find(self);
+    if (it == info.DtorHooks.end())
+      continue;
+    entry = it->second;
+    info.DtorHooks.erase(it);
+    found = true;
+    break;
+  }
```
Vipul-Cariappa: This loop and the mutex can be removed, if we, by default, have 1 slot that stores a pointer to `VTableOverlayDtorHookEntry`. Because this function gets the instance, we can figure out the original destructor and the custom destructor from that slot.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2070,17 +2077,110 @@ static bool hasComplexVTableLayout(const CXXRecordDecl* RD) {
...
+  if (!found)
+    return; // shouldn't happen; defensive
+  entry.orig_dtor(self);
+  if (entry.overlay)
+    entry.overlay->dtor_fired = true;
+  if (entry.cleanup)
+    entry.cleanup(self, entry.cleanup_data);
```
Vipul-Cariappa: The original destructor should be called after the custom. What if the custom destructor wants to read the data for some reason? It will be reading invalid memory.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2070,17 +2077,110 @@ static bool hasComplexVTableLayout(const CXXRecordDecl* RD) {
...
+    if (it == info.DtorHooks.end())
+      continue;
+    entry = it->second;
+    info.DtorHooks.erase(it);
+    found = true;
+    break;
+  }
+  if (!found)
+    return; // shouldn't happen; defensive
+  entry.orig_dtor(self);
```
```suggestion

```

lib/CppInterOp/CppInterOp.cpp
```suggestion
    entry.cleanup(self, entry.cleanup_data);
  entry.orig_dtor(self);
```

lib/CppInterOp/CppInterOp.cpp
vgvassilev: Ok, that was difficult to argue -- we have a new approach where we store the hidden self next to the slots.

unittests/CppInterOp/VTableOverlayCrossTUBench.cpp
vgvassilev: The new approach does not require mutex since we store it next to the overlay.

include/CppInterOp/CppInterOpThunks.h
vgvassilev: Yeah, that's the catalog with the fast pre-compiled thunks -- now it has a test.

lib/CppInterOp/CppInterOp.cpp
vgvassilev: Good catch. Reworked.

Vipul-Cariappa (approved):
I don't completely understand why `CppImpl::Thunks::dispatch` is required. But maybe I will understand it better once I see the associated cppyy PR.
Other than that, this looks good to me!

PR #1020, opened 2026-06-02
Add IsAllocator and IsDeallocator API functions

keremsahn:
Adds enum classes for allocation and deallocation types, creates API functions for queries.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1581,27 +1581,26 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
 
 bool IsAllocator(ConstFuncRef Fn) {
   INTEROP_TRACE(Fn);
-  if (Fn) {
-    const auto* D = unwrap<clang::Decl>(Fn);
-    if (const auto* FD = dyn_cast<FunctionDecl>(D)) {
-      if (FD->getBuiltinID() == Builtin::ID::BImalloc)
-        return INTEROP_RETURN(true);
-      if (const auto* FDA = FD->getAttr<RestrictAttr>()) {
-        if (FDA->getSemanticSpelling() != RestrictAttr::Declspec_restrict)
-          return INTEROP_RETURN(true);
-      }
-
-      if (const auto* FDA = FD->getAttr<OwnershipAttr>()) {
-        if (FDA->getOwnKind() == OwnershipAttr::Returns)
-          return INTEROP_RETURN(true);
-      }
-      // FIXME: These attributes are not currently compatible with our memory
-      // representation, they are added for APINotes' test
-      if (FD->hasAttr<CFReturnsRetainedAttr>() ||
-          FD->hasAttr<NSReturnsRetainedAttr>() ||
-          FD->hasAttr<OSReturnsRetainedAttr>())
+  if (!Fn)
+    return INTEROP_RETURN(false);
+  const auto* D = unwrap<clang::Decl>(Fn);
+  if (const auto* FD = dyn_cast<FunctionDecl>(D)) {
+    if (FD->getBuiltinID() == Builtin::ID::BImalloc)
+      return INTEROP_RETURN(true);
+    if (const auto* FDA = FD->getAttr<RestrictAttr>()) {
+      if (FDA->getSemanticSpelling() != RestrictAttr::Declspec_restrict)
         return INTEROP_RETURN(true);
     }
+
+    if (const auto* FDA = FD->getAttr<OwnershipAttr>()) {
+      if (FDA->getOwnKind() == OwnershipAttr::Returns)
+        return INTEROP_RETURN(true);
+    }
+
+    if (FD->hasAttr<CFReturnsRetainedAttr>() ||
+        FD->hasAttr<NSReturnsRetainedAttr>() ||
+        FD->hasAttr<OSReturnsRetainedAttr>())
+      return INTEROP_RETURN(true);
   }
 
   return INTEROP_RETURN(false);
@@ -1609,15 +1608,15 @@ bool IsAllocator(ConstFuncRef Fn) {
 
 bool IsDeallocator(ConstFuncRef Fn) {
   INTEROP_TRACE(Fn);
-  if (Fn) {
-    const auto* D = unwrap<clang::Decl>(Fn);
-    if (const auto* FD = dyn_cast<FunctionDecl>(D)) {
-      if (FD->getBuiltinID() == Builtin::ID::BIfree)
+  if (!Fn)
+    INTEROP_RETURN(false);
+  const auto* D = unwrap<clang::Decl>(Fn);
+  if (const auto* FD = dyn_cast<FunctionDecl>(D)) {
+    if (FD->getBuiltinID() == Builtin::ID::BIfree)
+      return INTEROP_RETURN(true);
+    if (const auto* FDA = FD->getAttr<OwnershipAttr>()) {
+      if (FDA->getOwnKind() == OwnershipAttr::Takes)
         return INTEROP_RETURN(true);
-      if (const auto* FDA = FD->getAttr<OwnershipAttr>()) {
-        if (FDA->getOwnKind() == OwnershipAttr::Takes)
-          return INTEROP_RETURN(true);
-      }
     }
   }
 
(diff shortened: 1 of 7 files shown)


keremsahn:
> Adds enum classes for allocation and deallocation types, creates API functions for queries.

@Vipul-Cariappa this PR is a prototype, I just searched for most obvious attributes there are more attributes should be added, please provide me feedback. Also it seems not to be possible to detect allocation-deallocation type via attributes. So it may not to be a good idea to build query functions based on attribute searching

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1558,6 +1558,44 @@ TCppType_t GetFunctionReturnType(TCppFunction_t func) {
   return INTEROP_RETURN(nullptr);
 }
 
+bool IsAllocator(TCppConstFunction_t Fn) {
+  INTEROP_TRACE(Fn);
+  assert(Fn && "Not func");
+  auto* D = static_cast<const Decl*>(Fn);
+  assert(D && "Not func");
+  auto* FD = dyn_cast<FunctionDecl>(D);
+  assert(FD && "Not func");
```
Vipul-Cariappa: We don't want to crash the application if `Fn = nullptr`, instead if return a sensible default. In this case, `false`. You should remove the the `assert`s.
Asserts are used to detect bugs within the CppInterOp codebase itself. In this case, it is the user providing a null value.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1558,6 +1558,44 @@ TCppType_t GetFunctionReturnType(TCppFunction_t func) {
...
+  return INTEROP_RETURN(false);
+}
+
+bool IsDeallocator(TCppConstFunction_t Fn) {
+  INTEROP_TRACE(Fn);
+  assert(Fn && "Not func");
+  auto* D = static_cast<const Decl*>(Fn);
+  assert(D && "Not func");
+  auto* FD = dyn_cast<FunctionDecl>(D);
+  assert(FD && "Not func");
```
Vipul-Cariappa: Please remove the `assert`s. As explained above.

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -235,6 +235,23 @@ enum class InterpreterLanguageStandard : unsigned char {
...
+  Unknown // Could not detect type, for now
+};
+
+enum class DeallocType : unsigned char {
+  Delete,
+  DeleteArr,
+  Free,
+  None,
+  Unknown // Could not detect type, for now
+};
```
Vipul-Cariappa: These are not yet used. So let's remove them from this PR. But this looks good. We should model based on this. Are there attributes that differentiate between `malloc` and `new`?
Also looking at https://clang.llvm.org/docs/AttributeReference.html#ownership-holds-ownership-returns-ownership-takes-clang-static-analyzer. There is `ownership_holds`. I believe the `DeallocType` should also express that.

Vipul-Cariappa (changes requested):
This looks good. Nice job. Some minor comments.

Vipul-Cariappa:
Please consider rebasing the PR before continuing the work.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1558,6 +1558,44 @@ TCppType_t GetFunctionReturnType(TCppFunction_t func) {
...
+  if (auto* FDA = FD->getAttr<OwnershipAttr>()) {
+    auto AttrType = FDA->getOwnKind();
+    if (AttrType == OwnershipAttr::Returns)
+      return INTEROP_RETURN(true);
+  }
+
+  return INTEROP_RETURN(false);
+}
+
+bool IsDeallocator(TCppConstFunction_t Fn) {
```
vgvassilev: Don't forget the builtins: https://github.com/vgvassilev/clad/blob/b10626e37d7d831ac94ca8a559ce9d2c93fa6ed4/lib/Differentiator/CladUtils.cpp#L865

include/CppInterOp/CppInterOpTypes.h
keremsahn: Yeah this is a problem we can not differentiate allocation-deallocation types from attributes. I will change the PR by your suggestions next week.

lib/CppInterOp/CppInterOp.cpp
keremsahn: I added for malloc and free, we may extend it for realloc etc. later

include/CppInterOp/CppInterOpTypes.h
keremsahn: > These are not yet used. So let's remove them from this PR. But this looks good. We should model based on this. Are there attributes that differentiate between `malloc` and `new`? Also looking at https://clang.llvm.org/docs/AttributeReference.html#ownership-holds-ownership-returns-ownership-takes-clang-static-analyzer. There is `ownership_holds`. I believe the `DeallocType` should also express that.

I did not handle OwnershipHolds in current commit but this can be another ownership category, probably cppyy will handle this same as a OwnershipTakes but it may be better to distinguish them. What do you think

keremsahn:
@vgvassilev @aaronj0 @Vipul-Cariappa  Hello all, can you review the PR, it is getting larger and harder to debug for me

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1586,6 +1587,51 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
   return INTEROP_RETURN(nullptr);
 }
 
+bool IsAllocator(ConstFuncRef Fn) {
+  INTEROP_TRACE(Fn);
+  if (Fn) {
+    const auto* D = unwrap<clang::Decl>(Fn);
```
Vipul-Cariappa: You can do an early return:
```c++
if (!Fn) return false;
// true logic without nesting whitespace
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1586,6 +1587,51 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
...
+    if (const auto* FD = dyn_cast<FunctionDecl>(D)) {
+      if (FD->getBuiltinID() == Builtin::ID::BImalloc)
+        return INTEROP_RETURN(true);
+      if (const auto* FDA = FD->getAttr<RestrictAttr>()) {
+        if (FDA->getSemanticSpelling() != RestrictAttr::Declspec_restrict)
+          return INTEROP_RETURN(true);
```
```suggestion
          return INTEROP_RETURN(true);
```
Shouldn't this be the following?
I am looking at the definition of `RestrictAttr::getSemanticSpelling`.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1586,6 +1587,51 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
...
+    }
+  }
+
+  return INTEROP_RETURN(false);
+}
+
+bool IsDeallocator(ConstFuncRef Fn) {
+  INTEROP_TRACE(Fn);
+  if (Fn) {
+    const auto* D = unwrap<clang::Decl>(Fn);
```
Vipul-Cariappa: Invert the `if` conditional here, too. This reduces the amount of nesting in the code.

Vipul-Cariappa (commented):
Minor comments:

lib/CppInterOp/CppInterOp.cpp
keremsahn: https://learn.microsoft.com/en-us/cpp/cpp/restrict?view=msvc-170
The definition of declspec(restrict) is kinda different from gnu's restrict. It implies pointer returned by the function has no alias, so it is not a neccesity for function to allocate memory.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1550,6 +1551,51 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
...
+      if (FDA->getSemanticSpelling() != RestrictAttr::Declspec_restrict)
+        return INTEROP_RETURN(true);
+    }
+
+    if (const auto* FDA = FD->getAttr<OwnershipAttr>()) {
+      if (FDA->getOwnKind() == OwnershipAttr::Returns)
+        return INTEROP_RETURN(true);
+    }
+    // FIXME: These attributes are not currently compatible with our memory
+    // representation, they are added for APINotes' test
```
```suggestion
```

Note required here. This FIXME should be part of the other PR.

Vipul-Cariappa (approved):
LGTM! Just a small comment. You also need to rebase, so that I can merge.

PR #1023, opened 2026-06-03
Add Cpp::Box, the typed result carrier for the AOT/JIT boundary

vgvassilev:
Add a small tagged-union type, Cpp::Box, that bindings and the interpreter use to exchange typed values in both directions. A binding constructs one with Box::Create<T>(x) and hands it to the runtime; the interpreter returns one from Evaluate and the binding extracts it with unbox<T>(), convertTo<T>(), or visit(v). The same shape works for the fundamental scalar types (stored inline) and for object payloads (stored behind a refcounted pointer the producer attaches via a {retain, release} ops table). Without Box every binding has to write its own Kind switch and duplicate the per-type storage logic.

Evaluate now returns a Box (replacing the old intptr_t + bool* out-param). Object payloads round-trip through a small ValueRefCount wrapper in Compatibility.h next to compat::Value; the wrapper uses clang::Value's copy ctor instead of move ctor to work around an upstream bug fixed by https://github.com/llvm/llvm-project/pull/200888, and a static_assert(LLVM_VERSION_MAJOR < 23) makes sure we revisit it.

Box.h itself is kept cheap to parse since the JIT pulls it whenever a catalog thunk or evaluated snippet references Cpp::Box. We avoid CppInterOpTypes.h (it transitively brings in <vector>, <string>, <set>), use unsigned char for the Kind underlying type instead of <cstdint>, gate <cassert> behind NDEBUG, and inline the cross-platform always_inline macro locally. Under NDEBUG the header pulls in zero project or standard headers; with asserts on, only <cassert>. Tests cover the fundamentals and K_PtrOrObj round-trip (POD, non-trivial destructor, class-with-static-member), refcounted copy + move semantics, and visit / convertTo on both AOT and runtime paths.

### DIFF
--- a/include/CppInterOp/Box.h
+++ b/include/CppInterOp/Box.h
@@ -1,0 +1,295 @@
+//===--- Box.h - Typed result carrier across the AOT/JIT boundary -*- C++ -*-=//
+//
+// Part of the compiler-research project, under the Apache License v2.0 with
+// LLVM Exceptions.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Cpp::Box is the typed result vocabulary that crosses the binding-AOT /
+// clang-repl-JIT boundary in both directions: catalog thunks Create<T> a
+// Box to hand to the runtime, Evaluate returns one to hand back. Storage
+// is hybrid -- fundamentals (K_Bool ... K_LongDouble) live inline in the
+// union; K_PtrOrObj keeps an owned pointer plus a type-erased
+// {retain, release} pair so the public header sees only `void*` and
+// function pointers, never the concrete payload type. Copy is shallow +
+// refcounted on K_PtrOrObj (matching clang::Value's model); the bridge
+// composes only when both layers share that semantic.
+//
+// The AOT path `Box::Create<T>(x).visit(v)` is designed to fold to a
+// direct `v(x)` -- the always_inline annotations below preserve this
+// against compiler inliner heuristics (validated separately on clang 17
+// and gcc 12 -O3; this commit does not pin the property in-tree).
+//
+// CPP_BOX_BUILTIN_TYPES is the single source of truth: enum positions,
+// storage union slots, KindOf<T> specialisations, and visit's switch are
+// all generated from this one X-macro -- one edit per new fundamental.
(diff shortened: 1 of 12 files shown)


include/CppInterOp/Box.h
```diff
@@ -0,0 +1,295 @@
...
+  };
+
+private:
+  union Storage {
+#define X(type, name) type m_##name;
+    CPP_BOX_BUILTIN_TYPES
+#undef X
+    struct {
+      void* m_Ptr;
+      const ObjectOps* m_Ops;
```
Vipul-Cariappa, line 133:
```suggestion
      const ObjectOps m_Ops;
```
Is storing the struct itself possible?

unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -84,27 +84,190 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, Interpreter_Evaluate) {
...
+  EXPECT_EQ(Cpp::Evaluate("long double c = 14.; c").unbox<long double>(), 14.L);
+  EXPECT_EQ(Cpp::Evaluate("long double d = 15.; d").unbox<long double>(), 15.L);
+  EXPECT_EQ(
+      Cpp::Evaluate("unsigned long long e = 16; e").unbox<unsigned long long>(),
+      16ULL);
+  // Object payload: K_PtrOrObj carries an owned pointer to the JIT-managed
+  // instance; lifetime ends with the Value going out of scope.
+  Cpp::Box sV = Cpp::Evaluate("struct S{} s; s");
+  EXPECT_EQ(sV.getKind(), Cpp::Box::K_PtrOrObj);
+  EXPECT_NE(sV.getObjectPtr(), nullptr);
```
Vipul-Cariappa, line 106: Can we also test the `QualType` stored? Using `getType`. I could not see `getType` being tested.

Vipul-Cariappa (approved):
Looks good to me! Modulo minor comments!

Vipul-Cariappa:
I guess the cppyy failure is expected. We will need to update all usages of `Evaluate` to the new overload.

vgvassilev:
[osx26-x86-clang-cling-llvm20-cppyy](https://github.com/compiler-research/CppInterOp/actions/runs/26954036493/job/79526369911?pr=1023#logs) seems stuck -- moving forward, the fixup pr is here https://github.com/compiler-research/cppyy/pull/222

PR #1030, opened 2026-06-08
[error-model] Add Cpp::Result<T>.

vgvassilev:
Fallible CppInterOp APIs today return ad-hoc nulls, out-params, or crash on bad input. Result<T> is the single [[nodiscard]] return shape every future fallible API will use: it carries either a value of T or an ErrorRef, callers check .ok() then read .value() or .error(), and the type is sized at parity with llvm::Expected<T> (pinned by static_assert).

Debug builds enforce must-handle: an error-bearing Result destroyed without being inspected aborts. Explicit .ignore() opts out. Release builds drop the bit and the check.

This PR is only the type plus its tests and a size + Ok-path benchmark. The boundary translator that produces Results, per-call diagnostics, structured payloads, and existing-API migrations are follow-ups -- each landed on top, layered against this contract.

# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

## Testing

## Checklist

### DIFF
--- a/include/CppInterOp/Error.h
+++ b/include/CppInterOp/Error.h
@@ -1,0 +1,244 @@
+//===--- Error.h - Cpp::Result<T> -------------------------------*- C++ -*-===//
+//
+// Part of the compiler-research project, under the Apache License v2.0 with
+// LLVM Exceptions.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Cpp::Result<T> -- the [[nodiscard]] return type fallible APIs use.
+// A Result holds either a value of T or an ErrorRef. The discipline
+// matches llvm::Expected: call .ok() (or coerce to bool), then .value()
+// on success or .error() on failure. value() on an error aborts;
+// value_or(fallback) is the lenient form.
+//
+// In debug builds, dropping an error-bearing Result without inspecting
+// it aborts. .ignore() opts out of that check.
+//
+// This PR is just the type. Later PRs add the boundary translator,
+// per-call diagnostics, and structured failure payloads.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CPPINTEROP_ERROR_H
+#define CPPINTEROP_ERROR_H
+
+#include "CppInterOp/CppInterOpTypes.h"
+
+#ifdef __cplusplus
+
+#include <cstddef>
+#include <cstdint>
+#include <new>
+#include <type_traits>
+#include <utility>
+
+namespace Cpp {
+
+/// Coarse outcome code. Status::Failed is a placeholder -- later PRs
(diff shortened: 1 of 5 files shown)


include/CppInterOp/Error.h
```diff
@@ -0,0 +1,245 @@
...
+  ~Result() {
+#ifndef NDEBUG
+    // Must-handle enforcement: an error-bearing Result that wasn't
+    // inspected aborts. Use .ignore() to drop on purpose.
+    if (m_Unchecked && m_HasError) {
+      std::fprintf(stderr,
+                   "Cpp::Result<T> destroyed without check (likely a "
+                   "dropped error). Call .ok() / .error() / .value() to "
+                   "inspect, or .ignore() to acknowledge.\n");
+      std::abort();
```
Vipul-Cariappa: Repeated code with `ResultAbort_ValueOnError`.

unittests/CppInterOp/ResultBench.cpp
```diff
@@ -0,0 +1,128 @@
...
+// Size invariants. Each row caps Cpp::Result<T> at llvm::Expected<T>'s
+// size; a future refactor that inflates the layout fails the build.
+
+static_assert(sizeof(Cpp::ErrorRef) == sizeof(void*),
+              "ErrorRef must be a single pointer");
+
+// Result<void> in release is just the ErrorRef pointer. Debug adds the
+// Unchecked bit; allow up to llvm::Error's debug-fudge ceiling.
+static_assert(sizeof(Cpp::Result<void>) <= sizeof(llvm::Error) * 2,
+              "Result<void> should not exceed llvm::Error layout");
```
Vipul-Cariappa: Should we have flaky checks like this?

unittests/CppInterOp/ResultBench.cpp
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+BENCHMARK(BM_ExpectedReturn_Ok);
+
+// Result<T> Ok-path within 2x of a bare return (Result is 16 B vs
+// bare 8 B: two register moves vs one).
+TEST(ResultPerfTest, OkPathNotMoreThan2xSlowerThanBareReturn) {
+  EXPECT_AT_LEAST_N_TIMES_FASTER(BM_BareReturn, BM_ResultReturn_Ok, 0.5);
+}
+
```
Vipul-Cariappa: Potentially flaky?

Vipul-Cariappa (approved):
I don't think we should be adding benchmarks for these changes.
Other than that, LGTM!

include/CppInterOp/Error.h
vgvassilev: Good catch.

unittests/CppInterOp/ResultBench.cpp
vgvassilev: Removed.

PR #1032, opened 2026-06-09
Support C++23 deducing this (explicit object parameters)

conrade-ctc:
Add support for C++23 explicit object parameters (P0847R7, "deducing this") across function introspection, the JIT call wrapper, and template overload resolution. The object parameter binds to the receiver, not the argument list, so each layer skips it:

- **Introspection** (`GetFunctionNumArgs`/`RequiredArgs`/`ArgType`/`ArgName`/`ArgDefault`): exclude the object parameter via `getNumNonObjectParams()` / `getMinRequiredExplicitArguments()` / `getNonObjectParameter()`.
- **JIT wrapper**: drop the object parameter from the emitted args/arity; bind rvalue-ref-qualified methods as rvalues (covers `this T&&` and the traditional `f() &&`).
- **Template deduction** (`BestOverloadFunctionMatch`): route explicit-object method templates through `AddMethodTemplateCandidate` with a synthesized receiver so `Self` deduces from it (handles `template<class Self>` and abbreviated `this auto&&`).

Also centralizes the cross-version `TypeDecl`→`QualType` conversion into `compat::GetTypeFromDecl` (handles the Clang 22 `getCanonicalTagType` change). Builds and passes on LLVM 21 and 22.

## Type of change

- [x] New feature

## Testing

New `FunctionReflection_DeducingThis*` tests in `FunctionReflectionTest.cpp`: introspection (arity/type/name/default), end-to-end JIT calls (plain, extra-arg, rvalue-ref, by-value, inheritance), template overload matching, and the use-cases from the MSVC "Deducing this" blog post. JIT-execution tests skip on Emscripten+LLVM>21 (known in-process-JIT limitation).

## Checklist

---

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -935,7 +935,7 @@ static std::string GetCompleteNameImpl(ConstDeclRef DRef, bool qualified) {
 
     if (const auto* TD = llvm::dyn_cast<TagDecl>(ND)) {
       std::string type_name;
-      QualType QT = C.Get_Tag_Type(TD);
+      QualType QT = compat::GetTypeFromDecl(TD);
       QT.getAsStringInternal(type_name, Policy);
       return type_name;
     }
@@ -1425,12 +1425,7 @@ static void GetClassDecls(ConstDeclRef DRef, std::vector<HandleType>& methods) {
   auto* CXXRD = dyn_cast<CXXRecordDecl>(D);
   compat::SynthesizingCodeRAII RAII(&getInterp());
   if (auto* CTSD = dyn_cast<ClassTemplateSpecializationDecl>(CXXRD)) {
-    QualType QT;
-#if CLANG_VERSION_MAJOR < 22
-    QT = QualType(CTSD->getTypeForDecl(), 0);
-#else
-    QT = getSema().getASTContext().getCanonicalTagType(CTSD);
-#endif
+    QualType QT = compat::GetTypeFromDecl(CTSD);
     if (!getSema().isCompleteType(CTSD->getLocation(), QT))
       return; // Unsuccesfull instantiaton
   }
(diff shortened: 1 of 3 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1878,14 +1893,36 @@ BestOverloadFunctionMatch(const std::vector<FuncRef>& candidates,
...
+        // the method-template entry point so Sema deduces the object parameter
+        // from a synthesized receiver of the enclosing record's type. (The
+        // plain AddTemplateOverloadCandidate path has no object and cannot
+        // deduce Self.)
+        CXXRecordDecl* RD = MD->getParent();
+#if CLANG_VERSION_MAJOR < 22
+        QualType ObjectType(RD->getTypeForDecl(), 0);
+#else
+        QualType ObjectType = C.getCanonicalTagType(RD);
+#endif
```
vgvassilev: That seems a repetitive pattern in the codebase. Can we move it into Compatibility.h?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1626,10 +1637,14 @@ size_t GetFunctionRequiredArgs(ConstFuncRef func) {
...
   if (const auto* FD = llvm::dyn_cast_or_null<FunctionDecl>(D))
-    return INTEROP_RETURN(FD->getMinRequiredArguments());
+    return INTEROP_RETURN(FD->getMinRequiredArguments() -
+                          NumExplicitObjectParams(FD));
 
-  if (const auto* FD = llvm::dyn_cast_or_null<clang::FunctionTemplateDecl>(D))
-    return INTEROP_RETURN((FD->getTemplatedDecl())->getMinRequiredArguments());
+  if (const auto* FD = llvm::dyn_cast_or_null<clang::FunctionTemplateDecl>(D)) {
+    const auto* TD = FD->getTemplatedDecl();
+    return INTEROP_RETURN(TD->getMinRequiredArguments() -
```
vgvassilev: Out of curiosity: why clang's `getMinRequiredArguments` does not account for `NumExplicitObjectParams`? Or the whole point of the patch is to drop the `this` argument so that the system figures it out by itself?

vgvassilev (commented):
Looks reasonable to me. I'd like @Vipul-Cariappa to take a look, too.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -3124,3 +3124,472 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_IsExplicitDeductionGuide) {
   EXPECT_EQ(Cpp::GetTypeAsString(Cpp::GetFunctionReturnType(guide)),
             "Wrapper<T *>");
 }
+
+// C++23 "deducing this" (explicit object parameters, P0847R7). The explicit
+// object parameter (`this Self self`) is a real ParmVarDecl but is bound to the
+// receiver, not supplied as a callee argument; these tests pin the behavior of
+// the argument-introspection helpers, the JIT call wrapper, and template
+// argument deduction around it. Deducing this requires Clang >= 18.
+#if CLANG_VERSION_MAJOR >= 18
```
vgvassilev: We only support 3 revisions of llvm back. That seems redundant.

Vipul-Cariappa (approved):
LGTM!

vgvassilev:
Can we use the llvm ai policy which suggests always a human behind the contribution: https://llvm.org/docs/AIToolPolicy.html

conrade-ctc:
Thanks, I went ahead and skipped the failing WASM llvm22 jit tests since it appears there are a bunch of these skipped already. We have some known issues with that combo, yes? @Vipul-Cariappa @vgvassilev

lib/CppInterOp/CppInterOp.cpp
conrade-ctc: good catch... turns out that there is already an upstream implementation to wrap this:`getMinRequiredExplicitArguments()`. i'll clean now and update the PR (i'll also rebase, since there is an update to main). thanks.

conrade-ctc: FYI, this is updated, I think we're clean now.

PR #1034, opened 2026-06-09
[error-model] Carry the Clang diagnostic on a failed call.

vgvassilev:
When a CppInterOp call fails the caller sees a return code -- "this didn't work" -- but not the Clang diagnostic underneath ("expected ';' at foo.cpp:12:5"). Bindings end up scraping stderr to recover that detail.

DiagnosticRef carries one captured diagnostic (severity, message, file, line, column). A clang::DiagnosticConsumer wired into every interpreter pipes parser and Sema output into a per-interpreter buffer that drains into a slice on the failure path. The slice sits behind the returned ErrorRef.

Errors also carry an optional structured payload for the two cases where Status alone leaves the caller flying blind: overload resolution (candidate list) and template deduction (failing param and reason). CapturedError extends an error's lifetime past its Result; ErrorRecord deep-copies into a plain value type. ErrorRef stays one pointer-word: an inline Status when there is no body, a slice pointer otherwise.

The consumer forwards to whatever client was installed previously so existing stderr output is preserved. takeClient transfers ownership when the engine owns its client (clang-REPL). cling installs with ShouldOwnClient=false, so takeClient returns null there; getClient still returns the original raw pointer between takeClient and setClient, and the consumer forwards through that. Without the raw-pointer leg, cling diagnostics would vanish from stderr -- which cppyy's failing-cppcode tests detect by grepping for "fatal error:".

### DIFF
--- a/include/CppInterOp/Error.h
+++ b/include/CppInterOp/Error.h
@@ -1,4 +1,4 @@
-//===--- Error.h - Cpp::Result<T> -------------------------------*- C++ -*-===//
+//===--- Error.h - Result<T>, ErrorRef, DiagnosticRef -----------*- C++ -*-===//
 //
 // Part of the compiler-research project, under the Apache License v2.0 with
 // LLVM Exceptions.
@@ -6,17 +6,10 @@
 //
 //===----------------------------------------------------------------------===//
 //
-// Cpp::Result<T> -- the [[nodiscard]] return type fallible APIs use.
-// A Result holds either a value of T or an ErrorRef. The discipline
-// matches llvm::Expected: call .ok() (or coerce to bool), then .value()
-// on success or .error() on failure. value() on an error aborts;
-// value_or(fallback) is the lenient form.
-//
-// In debug builds, dropping an error-bearing Result without inspecting
-// it aborts. .ignore() opts out of that check.
-//
-// This PR is just the type. Later PRs add the boundary translator,
-// per-call diagnostics, and structured failure payloads.
+// Cpp::Result<T> carries either a T or an ErrorRef. ErrorRef encodes
+// either an inline Status code (no allocation) or a pointer to an
+// ErrorSlice owning drained Clang diagnostics, an optional structured
+// payload, and producer attribution.
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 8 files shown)


lib/CppInterOp/ErrorInternal.cpp
```diff
@@ -1,29 +1,296 @@
-//===--- ErrorInternal.cpp - Cpp::Result<T> abort helper --------*- C++ -*-===//
+//===--- ErrorInternal.cpp - Cpp::Result<T> impl helpers --------*- C++ -*-===//
```
Vipul-Cariappa: Can we rename the file to `ErrorInterOp.cpp`?
I was initially a bit confused to see the implementation of the client APIs in a file called "internal".

unittests/CppInterOp/DiagnosticTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+  GTEST_SKIP() << "chained diagnostic trips MSBuild error scanning on Windows";
+#endif
+  TestFixture::CreateInterpreter();
+  Cpp::ClearPendingDiagnostics();
+  ASSERT_EQ(Cpp::GetPendingDiagnosticCount(), 0U);
+
+  // Trigger a parse error. silent=false lets the diagnostic reach the
+  // consumer; silent=true sets SuppressAllDiagnostics, which short-
+  // circuits before the consumer chain.
+  EXPECT_NE(0, Cpp::Declare("int err = ;", /*silent=*/false));
```
Vipul-Cariappa, line 122: What happens if we have `silent=true` here?

include/CppInterOp/Error.h
```diff
@@ -30,57 +23,277 @@
...
+      AmbiguousConversion, // FailingArgIndex valid
+      DeductionFailed,
+      ConstraintViolated,
+      ViableLessSpecialized, // viable but worse than another candidate
+    } Kind;
+    uint8_t FailingArgIndex = 0;
+    TypeRef ParamType;
+    TypeRef ProvidedType;
+  };
+  ArrayView<Candidate> Candidates;
```
Vipul-Cariappa: This PR does not integrate this diagnostic with the actual implementation.
Therefore, this PR is not complete. We should either drop `OverloadFailure` and `TemplateDeductionFailure` from this PR and open separate PRs for them individually or implement the necessary changes in this PR itself. I prefer the former.

Vipul-Cariappa (commented):
Few comments. Looks good in general.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -331,6 +429,38 @@ static compat::Interpreter& getInterp(InterpRef I = nullptr) {
   return *getInterpInfo().Interpreter;
 }
 
+CPPINTEROP_API unsigned GetPendingDiagnosticCount(InterpRef I) {
+  return static_cast<unsigned>(
+      getInterpInfo(I ? unwrap<compat::Interpreter>(I) : nullptr)
```
Vipul-Cariappa: Should we move this function (and functions below) to `ErrorInternal.cpp`?

vgvassilev:
@guitargeek, ping.

include/CppInterOp/Error.h
```diff
@@ -127,20 +286,18 @@ template <typename T> class [[nodiscard]] Result {
...
-    // inspected aborts. Use .ignore() to drop on purpose.
     if (m_Unchecked && m_HasError)
-      ResultAbort_UncheckedOnDtor(ErrorRef{m_ErrState});
+      ResultAbort_UncheckedOnDtor(m_ErrState);
 #endif
-    if (!m_HasError)
+    if (m_HasError)
+      ReleaseErrorRef(m_ErrState);
+    else
       valuePtr()->~T();
```
aaronj0, line 292: Do we know that all possible T on the callee end are trivially destructable?

include/CppInterOp/Error.h
```diff
@@ -101,24 +252,32 @@ template <typename T> class [[nodiscard]] Result {
...
 #endif
   }
 
   Result(Result&& Other) noexcept : m_HasError(Other.m_HasError) {
-    if (m_HasError)
+    if (m_HasError) {
       m_ErrState = Other.m_ErrState;
-    else
+      Other.m_ErrState = ErrorRef{};
+      Other.m_HasError = false;
```
aaronj0, line 274: If we keep this true, won't we take the ReleaseErrorRef(Ok) branch instead of ~T()?

vgvassilev: We moved the value of `m_HasError` in the ctor initialization section and now we just empty it so that the clients don't have the incorrect assumption that the object was not moved.

aaronj0 (approved):
LGTM, can you run clang-format before merging

PR #1037, opened 2026-06-11
[test] Unify path defines and add CPPINTEROP_BIN_DIR env override

conrade-ctc:
## Summary
- Collapse the test suite's two artifacts-prefix concepts (`CPPINTEROP_LIB_PATH` for the dispatch tests, `CPPINTEROP_BINARY_DIR` for TracingTests) into a single `CPPINTEROP_BIN_DIR`, overridable at runtime by an environment variable of the same name so relocated test binaries can locate `libclangCppInterOp`.
- Add `unittests/CppInterOp/TestPaths.h` (dependency-free) with `GetCppInterOpDirPath()` / `GetCppInterOpLibPath()`: env var wins, else the compile-time default, else empty — no hard dependency on the macro being defined.
- Drop `CPPINTEROP_LIB_PATH`; the dispatch tests now derive the library path from the prefix and fail with a clear message instead of `std::abort()` when the location is unknown.
- Rename the TracingTests source-tree macro to `CPPINTEROP_SRC_DIR` (it reads `.cpp`/`.h` sources, which never live in an artifacts prefix) and route its artifacts lookups through the shared helper.

(this is related to https://github.com/compiler-research/cppyy-backend/pull/211 as well)

## Test plan
- `check-cppinterop`: 7/7 passing, including `DispatchTests` and the `DispatchTests-isolation` ldd contract.
- Verified all three resolution paths: env override → installed lib, env unset → baked default, bogus env → clean failure (exit 1, no abort).

### DIFF
--- a/unittests/CppInterOp/CMakeLists.txt
+++ b/unittests/CppInterOp/CMakeLists.txt
@@ -106,9 +106,12 @@ add_cppinterop_unittest(TracingTests
   TracingTests.cpp
   Utils.cpp
 )
+# CPPINTEROP_SRC_DIR: source checkout (the coverage test reads .cpp sources).
+# CPPINTEROP_BIN_DIR: default artifacts prefix; the CPPINTEROP_BIN_DIR
+# environment variable overrides it at runtime (see TestPaths.h).
 target_compile_definitions(TracingTests PRIVATE
-  "CPPINTEROP_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}/../../\""
-  "CPPINTEROP_BINARY_DIR=\"${CMAKE_CURRENT_BINARY_DIR}/../../\""
+  "CPPINTEROP_SRC_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}/../../\""
+  "CPPINTEROP_BIN_DIR=\"${CMAKE_BINARY_DIR}\""
 )
 # GCC 12 and (less often) 13/14 emit a -Wmaybe-uninitialized false positive
 # inside libstdc++'s <regex> headers when the std::function subobject of
@@ -182,8 +185,10 @@ if(NOT EMSCRIPTEN AND BUILD_SHARED_LIBS)
     ${CMAKE_CURRENT_BINARY_DIR} ${GTEST_INCLUDE_DIR})
   target_link_libraries(DispatchTests PRIVATE ${gtest_libs} ${link_pthreads_lib})
   # Deliberately no target_link_libraries(... clangCppInterOp).
+  # Default artifacts prefix; the CPPINTEROP_BIN_DIR environment variable
+  # overrides it at runtime (see TestPaths.h).
   target_compile_definitions(DispatchTests PRIVATE
-    CPPINTEROP_LIB_PATH="${CMAKE_BINARY_DIR}/lib/libclangCppInterOp${CMAKE_SHARED_LIBRARY_SUFFIX}"
+    CPPINTEROP_BIN_DIR="${CMAKE_BINARY_DIR}"
   )
   set_output_directory(DispatchTests
     BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/bin/$<CONFIG>/
--- a/unittests/CppInterOp/TestPaths.h
+++ b/unittests/CppInterOp/TestPaths.h
@@ -1,0 +1,40 @@
+#ifndef CPPINTEROP_UNITTESTS_LIBCPPINTEROP_TESTPATHS_H
+#define CPPINTEROP_UNITTESTS_LIBCPPINTEROP_TESTPATHS_H
+
+#include <cstdlib>
+#include <string>
+
+namespace TestUtils {
+
+/// CppInterOp artifacts prefix (a directory containing lib/ and include/).
+/// The CPPINTEROP_BIN_DIR environment variable overrides the compile-time
+/// default so relocated test binaries can find the library. Empty if
+/// neither is available.
+inline std::string GetCppInterOpDirPath() {
+  if (const char* env = std::getenv("CPPINTEROP_BIN_DIR"))
+    return env;
+#ifdef CPPINTEROP_BIN_DIR
+  return CPPINTEROP_BIN_DIR;
+#else
+  return {};
+#endif
+}
+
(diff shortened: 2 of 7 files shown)


conrade-ctc:
I had a more localized change originally, but there were some inconsistencies across the repos that I thought might be nice to tackle now since we're migrating to a consolidated repo setup, and consistency across that set seemed like a good idea. Let me know if you think that's not the right approach. @vgvassilev @Vipul-Cariappa @aaronj0

unittests/CppInterOp/CMakeLists.txt
```diff
@@ -105,9 +105,12 @@ add_cppinterop_unittest(TracingTests
...
   Utils.cpp
 )
+# CPPINTEROP_SRC_DIR: source checkout (the coverage test reads .cpp sources).
+# CPPINTEROP_BIN_DIR: default artifacts prefix; the CPPINTEROP_BIN_DIR
+# environment variable overrides it at runtime (see TestPaths.h).
 target_compile_definitions(TracingTests PRIVATE
-  "CPPINTEROP_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}/../../\""
-  "CPPINTEROP_BINARY_DIR=\"${CMAKE_CURRENT_BINARY_DIR}/../../\""
+  "CPPINTEROP_SRC_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}/../../\""
+  "CPPINTEROP_BIN_DIR=\"${CMAKE_BINARY_DIR}\""
```
vgvassilev: I believe that change breaks ROOT. @aaronj0, can you review?

unittests/CppInterOp/TestPaths.h
```diff
@@ -0,0 +1,40 @@
+#ifndef CPPINTEROP_UNITTESTS_LIBCPPINTEROP_TESTPATHS_H
+#define CPPINTEROP_UNITTESTS_LIBCPPINTEROP_TESTPATHS_H
```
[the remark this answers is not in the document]
conrade-ctc: I think these need to be ignored, right? Maybe there is a clang-tidy config update so this doesn't get flagged @vgvassilev?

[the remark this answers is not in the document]
vgvassilev: Yes!

unittests/CppInterOp/CMakeLists.txt
conrade-ctc: These defines are `PRIVATE` to the `TracingTests` target (test-internal only) — they're not exported to the install tree and can't reach ROOT. I also checked: neither ROOT's own CMake files nor its bundled CppInterOp use `CPPINTEROP_DIR` or `CPPINTEROP_BINARY_DIR` as CMake variables. So this rename shouldn't affect ROOT at all.

conrade-ctc:
Gentle ping @aaronj0 — would appreciate a review here when you get a chance. I've addressed the ROOT concern above (the defines are test-internal `PRIVATE`, not exported). This is the last piece of the CPPINTEROP_BIN_DIR env-override work.

unittests/CppInterOp/CMakeLists.txt
aaronj0: `CMAKE_BINARY_DIR` will evaluate to the build directory of the current project (ROOT), in which CppInterOp is built as a subproject (which is why that was changed to `${CMAKE_CURRENT_SOURCE_DIR}/../../`) . That will be problematic, since that is not where the CppInterOp binary lives in that build tree.

I can test the patch with my ongoing CppInterOp upgrade (https://github.com/root-project/root/pull/22728) and figure out what works for both usecases

aaronj0 (approved):
The changes look good to me!

aaronj0:
> The changes look good to me!

On a second glance at the [ROOT CI log](https://github.com/compiler-research/CppInterOp/actions/runs/32526801291/job/96910433384?pr=1037):

```
3: [ RUN      ] TracingTest.ReproducerCompilesViaInterpreter
3: In file included from input_line_4:1:
3: In file included from /tmp/cppinterop-reproducer-7f81f5.cpp:44:
3: /home/runner/work/CppInterOp/CppInterOp/root/interpreter/CppInterOp/unittests/CppInterOp/../..//include/CppInterOp/CppInterOp.h:35:10: fatal error: 'CppInterOp/CppInterOpDecl.inc' file not found
3: #include "CppInterOp/CppInterOpDecl.inc"
```

I think the definition change has affected the TracingTest as suspected. Can you try to make that definition relative to `CMAKE_CURRENT_BINARY_DIR` similar to the CppInterOp source? There should be a way to satisfy that build tree as well

conrade-ctc:
Good catch, fixed. `CPPINTEROP_BIN_DIR` is now `${CMAKE_CURRENT_BINARY_DIR}/../..`, so it resolves to CppInterOp's own build directory in both a standalone build and an `add_subdirectory` build, where `CMAKE_BINARY_DIR` is the host project's build directory instead. That is the directory `CPPINTEROP_GEN_DIR` writes the generated `include/CppInterOp/*.inc` into, so the include path the reproducer needs is correct in both layouts. I reproduced your layout with a stub host project that adds CppInterOp as a subproject under `<build>/interpreter/CppInterOp`: TracingTests fails there with the old value and passes 80/80 with the new one, and the standalone build stays green. DispatchTests keeps `CMAKE_BINARY_DIR`, because the shared library is written to the host's top-level `lib/`. I also renamed the last `CPPINTEROP_DIR` (the APINotes check in `FunctionReflectionTest.cpp`), so no old-style name is left.

I also fixed `BUILD.bazel`, which still passed the old define names, did not declare the new header in the test `srcs`, and never gave `CppInterOpTests` the path define at all. The Bazel dispatch test will keep failing until the matching change to the cppyy Bazel rules lands as well.

conrade-ctc:
@aaronj0 Tightened the comments as requested — comment-only amend, code unchanged (same 120 content lines vs main). Happy to prune further if any are still too verbose or unclear.

vgvassilev:
@aaronj0, is that good to go?

aaronj0:
> @aaronj0, is that good to go?

Yep, after the last commit, the patch works with ROOT and looks good to me! Thanks @conrade-ctc

PR #1039, opened 2026-06-19
Add an experimental Bazel build

conrade-ctc:
# Description

Adds an experimental, best-effort Bazel build alongside the supported CMake build. It builds `libclangCppInterOp.so` and the full unit-test suite against a local LLVM/Clang tree (selected via the `LLVM_DIR` env var), consuming the shared `cppyy_bazel` module (the `@llvm` repo + an opt-in clang toolchain + shared defs/rules) from the sibling `cppyy` repo.

The Bazel build is **not gated in CI** and may lag the CMake build; CMake remains the supported build. The one non-Bazel change is a small robustness fix to `InterpreterTest.cpp`'s DLM-isolation test so it asserts about path *leakage* rather than the cwd-dependent default library scan (needed under the Bazel runfiles layout; harmless under CMake).

`cppyy_bazel` lives in the `cppyy` repo (`bazel/` subdir); this repo references it via `local_path_override(... path = "../cppyy/bazel")`, so check out `cppyy` alongside CppInterOp. Part of a set of 4 coordinated PRs across the cppyy stack (review cppyy first).

Fixes # (n/a)

## Type of change

- [x] New feature

## Testing

All 7 Bazel test targets pass against a local LLVM tree (`LLVM_DIR=/path/to/llvm bazelisk test //:tests`): CppInterOpTests (incl. the full 420-case unit suite), the dispatch/isolation tests, DynamicLibraryManager, TracingTests, the downstream loader, and the VTableOverlay benchmark — matching the CMake baseline.

## Checklist


vgvassilev (approved):
LGTM!

PR #1041, opened 2026-06-22
Adds memory allocation analysis functions

keremsahn:
Does not support detection for multiple branches, that is another milestone will be done after discussed.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -114,6 +114,7 @@
 #include <map>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <set>
 #include <sstream>
 #include <stack>
@@ -1598,16 +1599,20 @@ static AllocType handleNew(const clang::CXXNewExpr* CNE) {
   return AllocType::New;
 }
 
+static AllocType AnalyzeAllocType(const clang::FunctionDecl* Fn);
+
 static AllocType handleCall(const clang::CallExpr* CE) {
   if (const auto* FD = CE->getDirectCallee()) {
     if (FD->getBuiltinID() == Builtin::ID::BImalloc)
       return AllocType::Malloc;
+    return AnalyzeAllocType(FD);
   }
-  return AllocType::None;
+  // Function pointer calle
+  return AllocType::Unknown;
 }
 
 static AllocType handleExpr(const clang::Expr* expr,
-                            std::map<const VarDecl*, AllocType>& map) {
+                            std::map<const VarDecl*, AllocType>& varMap) {
   const clang::Expr* finExpr = expr->IgnoreParenCasts();
   // Case: return new __type__
   if (const auto* CNE = dyn_cast<CXXNewExpr>(finExpr))
@@ -1616,10 +1621,9 @@ static AllocType handleExpr(const clang::Expr* expr,
   // Case: returns a variable
   else if (const auto* DRE = dyn_cast<DeclRefExpr>(finExpr)) {
     if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
-      auto it = map.find(VD);
-      if (it != map.end())
+      auto it = varMap.find(VD);
+      if (it != varMap.end())
         return it->second;
-      return AllocType::None;
     }
     // FIXME: BindingDecl, NonTypeTemplateParmDecl are not handled
     return AllocType::None;
@@ -1632,8 +1636,8 @@ static AllocType handleExpr(const clang::Expr* expr,
   return AllocType::None;
 }
 
-static void getAllRetStmt(const clang::CompoundStmt* CS,
-                          std::vector<const ReturnStmt*>& allRetStmt) {
+static std::vector<const ReturnStmt*>
+getAllRetStmt(const clang::CompoundStmt* CS) {
   struct RetStmtVisitor : RecursiveASTVisitor<RetStmtVisitor> {
     std::vector<const ReturnStmt*> vec;
     bool VisitReturnStmt(ReturnStmt* RS) {
@@ -1643,19 +1647,19 @@ static void getAllRetStmt(const clang::CompoundStmt* CS,
   };
   RetStmtVisitor Visitor;
   Visitor.TraverseStmt(const_cast<CompoundStmt*>(CS));
-  allRetStmt = std::move(Visitor.vec);
+  return Visitor.vec;
 }
 
-static void getAllVarDecl(const clang::CompoundStmt* CS,
-                          std::map<const VarDecl*, AllocType>& varMap) {
+static std::map<const VarDecl*, AllocType>
+getAllVarDecl(const clang::CompoundStmt* CS) {
   struct VarVisitor : RecursiveASTVisitor<VarVisitor> {
-    std::map<const VarDecl*, AllocType> map;
+    std::map<const VarDecl*, AllocType> varMap;
     bool VisitVarDecl(VarDecl* VD) {
       Expr* expr = VD->getInit();
       if (expr)
-        map[VD] = handleExpr(expr, map);
+        varMap[VD] = handleExpr(expr, varMap);
       else
-        map[VD] = AllocType::None;
+        varMap[VD] = AllocType::None;
       return true;
     }
 
@@ -1666,7 +1670,7 @@ static void getAllVarDecl(const clang::CompoundStmt* CS,
         if (auto* DRE = dyn_cast<DeclRefExpr>(LHS)) {
           if (auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
             Expr* RHS = BO->getRHS();
-            map[VD] = handleExpr(RHS, map);
+            varMap[VD] = handleExpr(RHS, varMap);
           }
         }
       }
@@ -1675,10 +1679,10 @@ static void getAllVarDecl(const clang::CompoundStmt* CS,
   };
   VarVisitor Visitor;
   Visitor.TraverseStmt(const_cast<CompoundStmt*>(CS));
-  varMap = std::move(Visitor.map);
+  return Visitor.varMap;
 }
 
-static AllocType AllocCheck(const clang::FunctionDecl* Fn) {
+static AllocType AnalyzeAllocType(const clang::FunctionDecl* Fn) {
   const clang::QualType QT = Fn->getReturnType();
   if (!QT->isPointerType())
     return AllocType::None;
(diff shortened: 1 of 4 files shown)


keremsahn:
@vgvassilev @Vipul-Cariappa @aaronj0

lib/CppInterOp/CppInterOp.td
```diff
@@ -763,6 +763,15 @@ def IsFunctionProtoType : CppInterOpAPI {
   ];
 }
 
+def AllocCheckAPI: CppInterOpAPI {
```
Vipul-Cariappa:
```suggestion
def GetAllocType: CppInterOpAPI {
```

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -692,6 +692,105 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_FunctionTypes) {
...
+    FnPtr18 func18(int n){ return func18_t; }
+
+    int* func19_helper(int n){ return nullptr; }
+    int* func19(int n){ return func19_helper(n); }
+
+    int* func20(int* (*fp)(int), int n){ return fp(n); }
+
+    int* func21(int n){ static char buf[16]; return new (&buf) int(n); }
+
+    int* func22(int n){ []{ return; }(); return new int(n); }
```
Vipul-Cariappa, line 810:
```c++
int *func23(int n) {
  return func22(n); // return New; recursive case
}
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1593,6 +1596,133 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+  std::vector<const ReturnStmt*> allRetStmt;
+  getAllRetStmt(CmpStmt, allRetStmt);
+  getAllVarDecl(CmpStmt, varMap);
+  for (const ReturnStmt* retStmt : allRetStmt) {
+    const clang::Expr* retExpr = retStmt->getRetValue();
+    if (retExpr == nullptr)
+      continue;
+    AllocType tmp = handleExpr(retExpr, varMap);
+    if (tmp != AllocType::None && tmp != AllocType::Unknown)
+      return tmp;
```
Vipul-Cariappa: This should be outside the loop, after matching all return statements.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1593,6 +1596,133 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+  if (const auto* CNE = dyn_cast<CXXNewExpr>(finExpr))
+    return handleNew(CNE);
+
+  // Case: returns a variable
+  else if (const auto* DRE = dyn_cast<DeclRefExpr>(finExpr)) {
+    if (const auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+      auto it = map.find(VD);
+      if (it != map.end())
+        return it->second;
+      return AllocType::None;
```
```suggestion
```
Unreachable?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1557,6 +1561,137 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+      vec.push_back(RS);
+      return true;
+    }
+  };
+  RetStmtVisitor Visitor;
+  Visitor.TraverseStmt(const_cast<CompoundStmt*>(CS));
+  return Visitor.vec;
+}
+
+static std::map<const VarDecl*, AllocType>
```
```suggestion
static std::unordered_map<const VarDecl*, AllocType>
```

Let's use `unordered_map` where possible.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1557,6 +1561,137 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+      if (expr)
+        varMap[VD] = handleExpr(expr, varMap);
+      else
+        varMap[VD] = AllocType::None;
+      return true;
+    }
+
+    bool VisitBinaryOperator(clang::BinaryOperator* BO) {
+      if (BO->getOpcode() == BO_Assign) {
+        Expr* LHS = BO->getLHS();
```
Vipul-Cariappa: Invert the conditional, forcing an early return and reducing nesting of code.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1557,6 +1561,137 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+
+static AllocType handleExpr(const clang::Expr* expr,
+                            std::map<const VarDecl*, AllocType>& varMap) {
+  const clang::Expr* finExpr = expr->IgnoreParenCasts();
+  // Case: return new __type__
+  if (const auto* CNE = dyn_cast<CXXNewExpr>(finExpr))
+    return handleNew(CNE);
+
+  // Case: returns a variable
+  else if (const auto* DRE = dyn_cast<DeclRefExpr>(finExpr)) {
```
```suggestion
  if (const auto* DRE = dyn_cast<DeclRefExpr>(finExpr)) {
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1557,6 +1561,137 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+        return it->second;
+    }
+    // FIXME: BindingDecl, NonTypeTemplateParmDecl are not handled
+    return AllocType::None;
+  }
+
+  // Case: malloc
+  else if (const auto* CE = dyn_cast<CallExpr>(finExpr)) {
+    return handleCall(CE);
+  }
```
```suggestion
  if (const auto* CE = dyn_cast<CallExpr>(finExpr))
    return handleCall(CE);
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1557,6 +1561,137 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+}
+
+static std::map<const VarDecl*, AllocType>
+getAllVarDecl(const clang::CompoundStmt* CS) {
+  struct VarVisitor : RecursiveASTVisitor<VarVisitor> {
+    std::map<const VarDecl*, AllocType> varMap;
+    bool VisitVarDecl(VarDecl* VD) {
+      Expr* expr = VD->getInit();
+      if (expr)
+        varMap[VD] = handleExpr(expr, varMap);
```
Vipul-Cariappa, line 1661: Does the `handleExpr` here cause it to be run 2 times on the same input, because we also have it in `VisitBinaryOperator`?

Vipul-Cariappa (commented):
Minor code quality related comments:

lib/CppInterOp/CppInterOp.cpp
keremsahn: These two functions handle different situations.
```cpp
int* x = new int(n);
```
AST of this statement: 
```cpp
-DeclStmt <line:16:5, col:24>
|   | `-VarDecl <col:5, col:23> col:10 used x 'int *' cinit
|   |   `-CXXNewExpr <col:14, col:23> 'int *' Function 0x4a8e6ff0 'operator new' 'void *(std::size_t)'
|   |     `-ImplicitCastExpr <col:22> 'int' <LValueToRValue>
|   |       `-DeclRefExpr <col:22> 'int' lvalue ParmVar 0x4a999868 'n' 'int'
```

```cpp
y=x;
```
AST of this: 
```cpp
-BinaryOperator <line:18:5, col:7> 'int *' lvalue '='
|   | |-DeclRefExpr <col:5> 'int *' lvalue Var 0x4a999b10 'y' 'int *'
|   | `-ImplicitCastExpr <col:7> 'int *' <LValueToRValue>
|   |   `-DeclRefExpr <col:7> 'int *' lvalue Var 0x4a9999c8 'x' 'int *'
```

So basically in variable declarations there is no BinaryOperator node, hence the intersection

Vipul-Cariappa: Thanks for the explanation. 👍🏽

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1557,6 +1562,138 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+        return it->second;
+    }
+    // FIXME: BindingDecl, NonTypeTemplateParmDecl are not handled
+    return AllocType::None;
+  }
+
+  // Case: malloc or another func call
+  if (const auto* CE = dyn_cast<CallExpr>(finExpr)) {
+    return handleCall(CE);
+  }
```
```suggestion
  if (const auto* CE = dyn_cast<CallExpr>(finExpr))
    return handleCall(CE);
```

Vipul-Cariappa (approved):
LGTM! Small formatting thing. Please rebase so that I can merge.

I have a question:
What happens if we analyse the following:
```c++
int* heap_alloc_tail_recr_sum(int n, int acc = 0) {
    if (n)
        return heap_alloc_tail_recr_sum(n - 1, acc + n);
    return new int(acc);
}
```
If this does not work, please open an issue for it once this PR is merged.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1587,6 +1592,137 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+      if (expr)
+        varMap[VD] = handleExpr(expr, varMap);
+      else
+        varMap[VD] = AllocType::None;
+      return true;
+    }
+
+    bool VisitBinaryOperator(clang::BinaryOperator* BO) {
+      if (BO->getOpcode() != BO_Assign)
+        return true;
```
vgvassilev, line 1669: Can we add coverage for this line?

keremsahn: Hi Vassil, I added one check the fourth commit. Our first assumption was analyzed function does not leak memory, but i wrote the test this way to point the issue so maybe I can fix it later.

vgvassilev: Can we add a comment capturing that intent?

keremsahn: Actually I added a FIXME up to the 
```cpp
TESTAC(27, NewArr);
 ```
Furthermore, I will open some issues I want to work on when the PR is merged, I can open for this one too. No worries

File: unittests/CppInterOp/FunctionReflectionTest.cpp [hunk 12]
```diff
@@ -749,6 +749,126 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_FunctionTypes) {
...
+  TESTAC(19, None);
+  TESTAC(20, Unknown);
+  TESTAC(21, None);
+  TESTAC(22, New);
+  TESTAC(23, None);
+  TESTAC(24, New);
+  TESTAC(25, NewArr);
+  TESTAC(26, Unknown);
+  // FIXME: Pointer overwriten by a non-assignment operator
+  TESTAC(27, NewArr);
```
Vipul-Cariappa, line 866: What if the pointer is overwritten by an assignment operator?
Example:
```c++
x = new int(10);
x = (int*)0x12345678;
return x;
```
I believe we should ignore pointer arithmetic. We assume the user will return a valid value after all arithmetic operations.

keremsahn: The last assignment overwrites and labels x as AllocType::None, I also think we can ignore pointer arithmetic but no harm to handle non-assign operators, at least the variable should be labeled as overwriten. This is a low priority FIXME I think

PR #1044, opened 2026-07-01
[test] Make handle-taking fn ptr coming from dlsym volatile

aaronj0:
In this test, the handle-taking pointer comes from dlsym at runtime, so the compiler cannot see through it. Here the pointer is a compile-time constant and GCC 11 exploits the type mismatch (which is UB) to constant- fold the call giving a wrong reslt. Route the call through a volatile local so the compiler treats the pointer as runtime-opaque, matching dispatch

### DIFF
--- a/unittests/CppInterOp/HandleTypesTest.cpp
+++ b/unittests/CppInterOp/HandleTypesTest.cpp
@@ -122,7 +122,14 @@ TEST(HandleTypes, AbiCompatibleWithVoidPtr) {
     bool (*void_ptr_fn)(void*);
     bool (*handle_fn)(DeclRef);
   } pun{&void_ptr_takes_nonnull};
+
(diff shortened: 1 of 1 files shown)


unittests/CppInterOp/HandleTypesTest.cpp
```diff
@@ -122,7 +122,9 @@ TEST(HandleTypes, AbiCompatibleWithVoidPtr) {
     bool (*void_ptr_fn)(void*);
     bool (*handle_fn)(DeclRef);
   } pun{&void_ptr_takes_nonnull};
+
+  volatile auto handle_fn = pun.handle_fn;
```
vgvassilev, line 131:
```suggestion
  // Make sure the compiler does not fold away pun.
  volatile auto handle_fn = pun.handle_fn;
```

vgvassilev (approved):
LGTM!

PR #1045, opened 2026-07-02
Work around ORC emulated-TLS weak-discard crash on the in-process JIT

conrade-ctc:
## Problem

On the in-process JIT, two incremental modules that each define the *same*
weak/`linkonce_odr` `thread_local` with a non-zero initializer crash the
interpreter. This happens whenever an `inline` function (or template static)
that odr-uses such a `thread_local` is pulled into two `PartialTranslationUnit`s
— exactly what `MakeFunctionCallable` and repeated `process()` calls do.

Root cause is in LLVM ORC: `IRMaterializationUnit`'s constructor registers
`__emutls_t.<var>` (emitted for non-zero-init emulated-TLS globals) in
`SymbolFlags` but **not** in `SymbolToDefinition`. When the second module
re-defines the weak symbol, `JITDylib::defineImpl` discards the duplicate via
`IRMaterializationUnit::discard()`, which looks the symbol up in
`SymbolToDefinition`, doesn't find it, and dereferences `end()` — an assertion
in `+Asserts` builds, heap corruption otherwise.

Fixed upstream in **llvm/llvm-project#207161** (relanded as #208413). The fix
ships in clang 24, not clang 23, so this protects CppInterOp users on 21.x
through 23.x in the meantime.

## Fix

`compat::dedupeWeakEmulatedTLS()` (in `Compatibility.h`) runs on each module
before it reaches the JIT, on ELF only. For every weak/`linkonce_odr`
`thread_local` definition seen more than once, it demotes the duplicate to
`available_externally`. The `IRMaterializationUnit` constructor skips
`available_externally` globals, so the duplicate contributes no symbol, the
buggy `discard()` is never reached, and the first module remains the sole
definition that later modules resolve against within the JITDylib — i.e. the
dedup `discard()` was trying to do, done in IR before the crash.

It's wired into `process()` and the no-value `ParseAndExecute()`, covering both
user `process()` and wrapper compilation. Blast radius is limited to weak
`thread_local`s; every other weak/COMDAT symbol keeps the normal path,
where `discard()` already works.

Gated `#if CLANG_VERSION_MAJOR < 24`, so it compiles out once the fix is in the
baseline.

## Test

`FunctionReflection_DiscardDuplicateWeakEmulatedTLS` reproduces the two-module
duplicate-weak-TLS scenario. `...ZeroInit` covers a `thread_local` with a
non-trivial constructor: zeroinitializer in IR, yet it still emits the
emulated-TLS companion. Both skip on COFF and Mach-O, where the pass does not run. Verified against a deliberately-unpatched
`libLLVMOrcJIT.a`: passes with the workaround, aborts without it (negative
control). Full suite stays green.

🤖 Done with the help of [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8, human in the loop)

### DIFF
--- a/lib/CppInterOp/Compatibility.h
+++ b/lib/CppInterOp/Compatibility.h
@@ -31,7 +31,12 @@
 
 #include "llvm/ADT/IntrusiveRefCntPtr.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringSet.h"
+#include "llvm/IR/GlobalValue.h"
+#include "llvm/IR/GlobalVariable.h"
+#include "llvm/IR/Module.h"
 #include "llvm/Support/FileSystem.h"
+#include "llvm/TargetParser/Triple.h"
 
 #include "CppInterOp/Box.h"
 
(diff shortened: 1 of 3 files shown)


conrade-ctc:
@vgvassilev, here's a compatability patch/work-around for the bug that I just posted a PR to clang for.

lib/CppInterOp/Compatibility.h
```diff
@@ -593,6 +596,57 @@ inline void maybeMangleDeclName(const clang::GlobalDecl& GD,
...
+    // discard() would have done (available_externally + no comdat) but without
+    // hitting the crashing code path.
+    if (!Defined.insert(GV.getName()).second) {
+      GV.setLinkage(llvm::GlobalValue::AvailableExternallyLinkage);
+      GV.setComdat(nullptr);
+    }
+  }
+}
+#else
+inline void dedupeWeakEmulatedTLS(llvm::Module&, llvm::StringSet<>&) {}
```
vgvassilev: I suspect that PR to land in 23 and all the fixes should be around that -- that will remind us one day when we drop support for llvm 22 to delete all of the relevant pieces of this workaround.

conrade-ctc: Done, rather than burying it in the stub in Compatibility.h, moved the guard out into the implementation so that it's a clearer cleanup. 👍

conrade-ctc:
@vgvassilev rebased onto current `main` (picks up #1064; the workaround commit is patch-identical, no content change). CI is green across the board — the only red is `ubu24-x86-gcc14-cling-llvm20-root`, which fails at the *Configure ROOT* step (`ninja: error: … etc/cppinterop/CppInterOp/Box.h is defined as an output multiple times`), i.e. a ROOT build-graph issue unrelated to this diff (this PR only edits the contents of existing headers + a test). `review`/clang-tidy is clean. Ready for another look whenever you have a moment — thanks!

vgvassilev (approved):
Lgtm!

PR #1046, opened 2026-07-02
[ci] Add code coverage for OS X

aaronj0:
OS X only codepaths currently are not run through codecov. Add a hosted osx26-arm cell with coverage: true and enable shared coverage. Install lcov via brew, and run gcov emulation through the llvm-cov and sink the different args into a substitution on the lcov invocations to keep the common logic similar

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,7 +53,7 @@ jobs:
             clang-runtime: '22'
             Valgrind: On
           # MacOS Arm
-          - { name: osx26-arm-clang-llvm22, os: macos-26, compiler: clang, clang-runtime: '22' }
+          - { name: osx26-arm-clang-llvm22-cov, os: macos-26, compiler: clang, clang-runtime: '22', coverage: true }
           - { name: osx26-arm-clang-llvm21-cppyy, os: macos-26, compiler: clang, clang-runtime: '21', cppyy: On, flavor: system }
           - { name: osx26-arm-clang-cling-llvm20-cppyy, os: macos-26, compiler: clang, clang-runtime: '20', cling: 'On', cppyy: On, flavor: cling, flavor-version: 'ROOT-llvm20' }
           # MacOS X86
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -130,11 +134,16 @@ jobs:
...
+        if [[ "$RUNNER_OS" == "macOS" ]]; then
+          LCOV_ARGS=(--gcov-tool "${CC%/*}/llvm-cov" --gcov-tool gcov
+                     --ignore-errors inconsistent,format,unsupported)
+        else
+          LCOV_ARGS=(--gcov-tool "/usr/bin/gcov-${CC#*-}")
+        fi
+        lcov --directory build/ --capture --output-file coverage.info "${LCOV_ARGS[@]}"
+        lcov --remove coverage.info '/usr/*' ${{ github.workspace }}'/install/*' --ignore-errors unused "${LCOV_ARGS[@]}" --output-file coverage.info
+        lcov --remove coverage.info '${{ github.workspace }}/unittests/*' --ignore-errors unused "${LCOV_ARGS[@]}" --output-file coverage.info
+        lcov --remove coverage.info  '${{ github.workspace }}/build/*' --ignore-errors unused "${LCOV_ARGS[@]}" --output-file coverage.info
```
vgvassilev, line 146: I realize that it might have better use if this is a reusable action on the ci-workflows end...

aaronj0: Yes. That would require a refactor of things across all the projects that run codecov. Does clad also have the same invocation flags? Maybe it can be an action that is parameterised on platform and the custom lcov flags

aaronj0: I would prefer landing this so that we can proceed with https://github.com/compiler-research/CppInterOp/pull/1043 and following up with making this reusable.

vgvassilev: Yes, all is copied and adapted from clad. Sure, that works as long as we keep track of that somewhere as an issue.

vgvassilev (approved):
Lgtm provided the ci works.

aaronj0:
> Lgtm provided the ci works.

Failure on linux unrelated, I have a fixup on the other repos following the valgrind drop on cppyy

PR #1050, opened 2026-07-05
[ci] Bump the OS X compiler toolchain to llvm21

aaronj0:
llvm-cov 20's gcov emulation fabricates out-of-range line records. See: https://github.com/compiler-research/CppInterOp/actions/runs/28658480421/job/85005593960?pr=1041. I was also able to reproduce it with a similar TU locally with llvm-cov 20. This mechanism was fixed in llvm-cov 21, and testing locally, it was able to parse the header cleanly. Bump the toolchain in OS X and drop the stale llvm15 install for x86 mac which is no longer tested.


vgvassilev (approved):
Lgtm!

PR #1048, opened 2026-07-03
Fix VTableOverlay dtor wrapper & test bitcasts on 32-bit MSVC

aaronj0:
Add __thiscall to the bitcast type on _M_IX86 so the caller reproduces the compiler's virtual-dispatch ABI.
Fixes all failing tests of the VTableOverlay on 32-bit MSVC:

| Test | Assert | Got | Expected |
|---|---|---|---|
| ReplacesSlotPreservingOthers | `call_slot(inst, kBeta, 5)` | -1211418038 | -5 |
| ReplacesMultipleSlots | `call_slot(inst, kAlpha, 5)` | -1211418038 | -5 |
| OverlayIsPerInstance | `call_slot(a, kBeta, 5)` | -1211418038 | -5 |
| ThunkReadsThisAndDataMember | — | SEH 0xc0000005 | 105 |
| DerivedClassWithOverride | `call_slot(inst, kAlpha, 5)` | -1211418038 | -5 |
| MultiLevelInheritance | `call_slot(inst, kAlpha, 5)` | -1872131220 | 10 |
| OverlayThroughHierarchyAccessesDataMembers | `call_slot_no_arg(&a/&b, kAlpha)` | 826692698 / 826692688 | 15 / 10 |
| SetsExtraPrefixSlots | `call_slot(inst, kBeta, 5)` | -1211418038 | -5 |

🤖 AI-assisted

PR #1051, opened 2026-07-06
[ci] Add workflow for 32-bit MSVC (windows-2025 x86)

aaronj0:
Should surface failures fixed by #1048

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1584,7 +1584,7 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
   INTEROP_TRACE(TyRef);
   QualType QT = QualType::getFromOpaquePtr(TyRef.data);
   const auto* T = QT.getTypePtr();
-  return llvm::isa_and_nonnull<clang::FunctionProtoType>(T);
+  return INTEROP_RETURN(llvm::isa_and_nonnull<clang::FunctionProtoType>(T));
 }
 
 void GetFnTypeSignature(ConstTypeRef fn_type, std::vector<TypeRef>& sig) {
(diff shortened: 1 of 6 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2200,21 +2232,29 @@ struct VTableOverlay {
 
 // Wrapper installed in the deleting-dtor slot when MakeVTableOverlay was
 // called with a non-null on_destroy. Recovers the owning VTableOverlay
-// from `self`'s vptr (hidden-slot fixed-offset load) and runs the
-// callback BEFORE the original destructor: `self` is alive at that
+// from `this`'s vptr (hidden-slot fixed-offset load) and runs the
```
vgvassilev, line 2235: Why we change the terminology here?

vgvassilev (commented):
Consider fixing the clang-tidy suggestions.

lib/CppInterOp/CppInterOp.cpp
aaronj0: the old wrapper was a free function `extern "C" void cppinteropVTableOverlayDtorWrapper(void* self)` and here `self` names the parameter. The fix makes the wrapper a member function (`VTableOverlayDtorHost::Wrapper`) so 32-bit MSVC annotates it `__thiscall` and with that, the self parameter no longer exists and the object arrives as the implicit this, so the comment changes that because it would be inaccurate otherwise.

vgvassilev: Ok. Fair enough.

aaronj0:
> Consider fixing the clang-tidy suggestions.

Addressed those that seemed relevant

vgvassilev (approved):
Lgtm!

PR #1052, opened 2026-07-06
Deallocation: Adds memory deallocation analysis functions

keremsahn:
FIXME 's and some other cases will be handled after I open issues.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1820,6 +1820,185 @@ AllocType GetAllocType(ConstFuncRef Fn) {
   return INTEROP_RETURN(AllocType::None);
 }
 
+namespace {
+struct DeallocationTraverser : RecursiveASTVisitor<DeallocationTraverser> {
+  std::unordered_map<const clang::VarDecl*, const clang::ParmVarDecl*> aliasMap;
+  std::set<const clang::ParmVarDecl*> overwritenParms;
+  std::vector<DeallocType>& VPP;
+  DeallocationTraverser(std::vector<DeallocType>& v) : VPP(v) {}
+
+  bool TraverseDecl(clang::Decl* D) {
+    if (D && llvm::isa<clang::TagDecl>(D))
+      return true;
+    return RecursiveASTVisitor::TraverseDecl(D);
+  }
+
+  bool VisitVarDecl(clang::VarDecl* VD) {
+    if (llvm::isa<clang::ParmVarDecl>(VD) || !VD->getType()->isPointerType())
+      return true;
+    const clang::Expr* E = VD->getInit();
+    updateMap(VD, E);
+    return true;
+  }
+
+  bool VisitBinaryOperator(clang::BinaryOperator* BO) {
+    if (BO->getOpcode() != BO_Assign)
+      return true;
+    const auto* VD = resolveExpr(BO->getLHS());
+    updateMap(VD, BO->getRHS());
+    return true;
+  }
+
+  bool VisitCXXDeleteExpr(clang::CXXDeleteExpr* CDE) {
+    const clang::Expr* E = CDE->getArgument();
+    const auto* PVR = resolveParam(E);
+    if (!PVR)
+      return true;
+    if (CDE->isArrayForm()) {
+      updateVPP(PVR->getFunctionScopeIndex(), DeallocType::DelArr);
+      return true;
+    }
+    updateVPP(PVR->getFunctionScopeIndex(), DeallocType::Del);
+    return true;
+  }
+
+  bool VisitCallExpr(clang::CallExpr* CE) {
+    const clang::FunctionDecl* FD = CE->getDirectCallee();
+    if (!FD)
+      return true;
+    if (FD->getBuiltinID() == Builtin::ID::BIfree) {
+      handleFree(CE);
+      return true;
+    }
+    if (!FD->getReturnType()->isVoidType())
+      return true;
+    // FIXME: Recursive Case
+    return true;
+  }
+
+  const clang::ParmVarDecl* resolveParam(const clang::Expr* E) {
+    const auto* VD = resolveExpr(E);
+    if (!VD)
+      return nullptr;
+    if (const auto* PVD = dyn_cast<clang::ParmVarDecl>(VD)) {
+      if (overwritenParms.count(PVD))
+        return nullptr;
+      return PVD;
+    }
+    auto it = aliasMap.find(VD);
+    if (it != aliasMap.end() && it->second != nullptr) {
+      return it->second;
+    }
+    return nullptr;
+  }
+
+  void updateMap(const clang::VarDecl* VD, const clang::Expr* E) {
+    if (!VD)
+      return;
+
+    if (!VD->getType()->isPointerType())
+      return;
+    if (const auto* PVD = dyn_cast<clang::ParmVarDecl>(VD)) {
+      overwritenParms.insert(PVD);
+      return;
+    }
+    const auto* rhsVD = resolveExpr(E);
+
+    if (!rhsVD) {
+      aliasMap[VD] = nullptr;
+      return;
+    }
+
+    if (const auto* rhsPVD = dyn_cast<clang::ParmVarDecl>(rhsVD)) {
+      if (overwritenParms.count(rhsPVD)) {
+        aliasMap[VD] = nullptr;
+        return;
+      }
+      aliasMap[VD] = rhsPVD;
+      // FIXME: does not handle situation where one param is assigned to
+      // another param
+      return;
+    }
+
+    auto it = aliasMap.find(rhsVD);
+    if (it != aliasMap.end()) {
+      aliasMap[VD] = it->second;
+      return;
+    }
+    aliasMap[VD] = nullptr;
+  }
+
+  void updateVPP(unsigned index, DeallocType DT) {
+    if (index >= VPP.size())
+      return;
+    if (VPP[index] == DeallocType::Unknown)
+      return;
+    if (VPP[index] != DeallocType::None && VPP[index] != DT) {
+      VPP[index] = DeallocType::Unknown;
+      return;
+    }
+    VPP[index] = DT;
+  }
+
+  void handleFree(clang::CallExpr* CE) {
+    if (CE->getNumArgs() < 1)
+      return;
+    const clang::Expr* paramExpr = CE->getArg(0);
+    const auto* PVR = resolveParam(paramExpr);
+    if (!PVR)
+      return;
+    updateVPP(PVR->getFunctionScopeIndex(), DeallocType::Free);
+  }
+
+  const clang::VarDecl* resolveExpr(const clang::Expr* E) {
+    if (!E)
+      return nullptr;
+    E = E->IgnoreParenCasts();
+    const auto* DRE = dyn_cast<clang::DeclRefExpr>(E);
+    if (!DRE)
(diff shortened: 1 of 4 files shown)


keremsahn:
@Vipul-Cariappa @aaronj0 @vgvassilev

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1587,6 +1591,192 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+  bool VisitBinaryOperator(clang::BinaryOperator* BO) {
+    if (BO->getOpcode() != BO_Assign)
+      return true;
+    const clang::Expr* LHS = BO->getLHS();
+    const clang::Expr* finLHS = LHS->IgnoreParenCasts();
+    const auto* DRE = dyn_cast<clang::DeclRefExpr>(finLHS);
+    if (!DRE)
+      return true;
+
+    const auto* VD = dyn_cast<clang::VarDecl>(DRE->getDecl());
```
vgvassilev: Here we should call Visit(DRE) to get to the VisitVarDecl and avoid duplication of code.

keremsahn: I created a resolveExpr helper; it does Expr-> non-castedExpr->DRE->VD with null checks, this pattern were used multiple times

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1790,6 +1790,232 @@ AllocType GetAllocType(ConstFuncRef Fn) {
...
+    // FIXME: try catch block are not handled
+    valPerParam.assign(numParam, DeallocType::Opaque);
+    visitedFuncs[FD] = valPerParam;
+    return false;
+  }
+  valPerParam.assign(numParam, DeallocType::None);
+  DeallocationTraverser Traverser(valPerParam, visitedFuncs);
+  Traverser.TraverseStmt(const_cast<CompoundStmt*>(CmpStmt));
+  visitedFuncs[FD] = valPerParam;
+  return true;
```
Vipul-Cariappa, line 2029: Shouldn't you first insert the function into `visitedFuncs` before `TraverseStmt`?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1790,6 +1790,232 @@ AllocType GetAllocType(ConstFuncRef Fn) {
...
+      aliasMap[VD] = it->second;
+      return;
+    }
+    aliasMap[VD] = nullptr;
+  }
+
+  void updateVPP(unsigned index, DeallocType DT) {
+    // Probably dead code, but better to keep
+    if (index >= valPerParam.size())
+      return;
```
```suggestion
```
You can remove this. No problem.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1790,6 +1790,232 @@ AllocType GetAllocType(ConstFuncRef Fn) {
...
+      return true;
+    }
+
+    // FIXME: Recursive Case
+    auto it = visitedFuncs.find(FD);
+    if (it != visitedFuncs.end()) {
+      // Function calls itself, or a cycle recursion
+      if (it->second == std::nullopt)
+        return true;
+      joinVectors(*(it->second), CE);
```
Vipul-Cariappa, line 1912: Is this line required? Shouldn't the ' joinVectors ' in line 1889 have already updated the values?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1790,6 +1790,232 @@ AllocType GetAllocType(ConstFuncRef Fn) {
...
+
+  void updateMap(const clang::VarDecl* VD, const clang::Expr* E) {
+    if (!VD)
+      return;
+    if (!VD->getType()->isPointerType() && !VD->getType()->isRecordType())
+      return;
+    if (const auto* PVD = dyn_cast<clang::ParmVarDecl>(VD)) {
+      overwritenParms.insert(PVD);
+      return;
+    }
```
Vipul-Cariappa, line 1946: What about:
```c++
void func(void *ptr) {
  ptr = ptr; // no over writting
}
```
You can add this as a FIXME because people do not write code like this.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -829,6 +829,270 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_FunctionTypes) {
...
+  TESTGDT(9, true, DT::Delete, DT::None);
+  TESTGDT(10, true, DT::Unknown, DT::None);
+  TESTGDT(11, false, DT::Opaque);
+  TESTGDT(12, false, DT::Opaque);
+  TESTGDT(13, true, DT::None);
+  TESTGDT(15, true, DT::Delete);
+  TESTGDT(16, true, DT::None);
+  TESTGDT(17, true, DT::None);
+  TESTGDT(18, true, DT::None, DT::None);
+  TESTGDT(19, true, DT::None);
```
Vipul-Cariappa: This should be `Opaque`?

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -829,6 +829,270 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_FunctionTypes) {
...
+  TESTGDT(20, true, DT::None);
```
Vipul-Cariappa: 19 and 20 are the same. Remove one of them.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -829,6 +829,270 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_FunctionTypes) {
...
+  TESTGDT(21, true, DT::Delete);
+  TESTGDT(22, true, DT::Delete);
+  TESTGDT(23, true, DT::Delete);
+  TESTGDT(24, true, DT::None, DT::None);
```
Vipul-Cariappa: `p` is deleted, and `q` is none here. But there is a FIXME for this.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -829,6 +829,270 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_FunctionTypes) {
...
+      delete ptr;
+    }
+
+    // FIXME: Can not resolve parameter location in recursive call
+    // Probably impossible to solve statically
+    void func33(int* ptr1, int* ptr2, int n){
+      if(n > 0)
+        func33(ptr2, ptr1, n-1);
+      delete ptr1;
+    }
```
Vipul-Cariappa, line 945: Nice catch. The current behaviour is good enough. You cannot resolve this statically.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -829,6 +829,270 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_FunctionTypes) {
...
+  TESTGDT(34, true, DT::Delete, DT::None);
+  TESTGDT(35, true, DT::None, DT::Delete);
+  TESTGDT(36, true, DT::Delete, DT::Delete);
+  TESTGDT(37, true, DT::Delete);
+  TESTGDT(38, true, DT::Delete);
+  TESTGDT(39, true, DT::None);
+  TESTGDT(40, true, DT::Unknown, DT::None);
+  // FIXME: check top of VisitCallExpr function
+  TESTGDT(41, true, DT::None, DT::None);
+  TESTGDT(42, true, DT::None);
```
Vipul-Cariappa, line 1072: If lambda `l` is executed, this should be deleted. Is this the case?

Vipul-Cariappa (approved):
LGTM! Modulo small changes. Add FIXME comments for now. Will merge it. You can fix it in a fresh PR.

lib/CppInterOp/CppInterOp.cpp
keremsahn: Well, there are two cases:
1)If function is analyzed via API:
```cpp
bool GetDeallocType(ConstFuncRef Fn, std::vector<DeallocType>& valPerParam) {
  INTEROP_TRACE(Fn, INTEROP_OUT(valPerParam));
  if (!Fn)
    return INTEROP_RETURN(false);

  const auto* D = unwrap<clang::Decl>(Fn);
  const auto* FD = dyn_cast<clang::FunctionDecl>(D);
  if (!FD)
    return INTEROP_RETURN(false);
  std::unordered_map<const FunctionDecl*,
                     std::optional<std::vector<DeallocType>>>
      visitedFuncs;
  visitedFuncs[FD] = std::nullopt;  //Check here we label it as "in progress"
  return INTEROP_RETURN(AnalyzeDeallocType(FD, valPerParam, visitedFuncs));
}
```

2)Analyzed via recursion:
```cpp
 auto it = visitedFuncs.find(FD);
    if (it != visitedFuncs.end()) {
      // Function calls itself, or a cycle recursion
      if (it->second == std::nullopt)
        return true;
      joinVectors(*(it->second), CE);
      return true;
    }
    std::vector<DeallocType> calleeVec;
    visitedFuncs[FD] = std::nullopt;   //Check here we label it as "in progress"
    AnalyzeDeallocType(FD, calleeVec, visitedFuncs);
    joinVectors(calleeVec, CE);
    return true;
```
So the line you saw is not injection, it is rewriting, injection is done before

lib/CppInterOp/CppInterOp.cpp
keremsahn: This join, works when a function is already in visitedFuncs and it is analyzed so VPP is not nullopt. 
The join in 1889, first makes recursive calls, then joins current VPP with freshly analyzed function's VPP.
Different purposes

lib/CppInterOp/CppInterOp.cpp
keremsahn: Parameter overwrite is handled at next PR, this situation is also not a problem anymore

unittests/CppInterOp/FunctionReflectionTest.cpp
keremsahn: That was a design choice, I did not let functions could not be analyzed make the VPP unknown
```cpp
    void func19(int* p){ helper19(p); }
```
Also this should be either None or Unknown, not Opaque. Opaque is for functions traverser never worked on. If you ask why this is not unknown as i said this was by choice. When I did tests on libstdcxx not via source code but including libraries, there were lots of Unknown because of precompiled functions. We can discuss this

unittests/CppInterOp/FunctionReflectionTest.cpp
keremsahn: This situation is also handled in next PR

unittests/CppInterOp/FunctionReflectionTest.cpp
keremsahn: Lambda calls are OperatorCallExpr and they are not analyzed:
```cpp
 bool VisitCallExpr(clang::CallExpr* CE) {
    // FIXME: for some operators, arg and parameter match may not match because
    // of hidden __this__ parameter
    if (llvm::isa<clang::CXXOperatorCallExpr>(CE))
      return true;
```
This guard was for parameter offset, it was also handled in next PR, so in the next PR, inside of lambda is actually analyzed but it does not deallocate any of its arguments, and ptr is not recognized when the recursive call is made, so it still returns none

PR #1053, opened 2026-07-07
[cmake] Use ExternalProject to generate GTest sub-build config

aaronj0:
Use declarative `CMAKE_ARGS`: ExternalProject generates the configure command itself and forwards the parent generator, platform and toolset. Fixes the case where a parent configured with `-A Win32 -Thost=x64` spawned sub-builds on the VS generator's default x64 platform, which tries to link x64 objects against the forwarded /machine:X86 linker flags and fail.

Required by #1051


vgvassilev (approved):
Lgtm!

PR #1054, opened 2026-07-07
[interp] Annotate IsFunctionProtoType's exit for tracing

aaronj0:
Found by running the win32 suite under `CPPINTEROP_LOG=1`. the traced function returns through a bare return instead of `INTEROP_RETURN`, so the tracing layer's unannotated-exit assert aborts the process when the API is called with tracing.


vgvassilev (approved):
Nice catch!

vgvassilev:
Why did we not discover that in the existing setup?

aaronj0:
> Why did we not discover that in the existing setup?

I was debugging something else (the DLM issue on winx86) and ran with `CPPINTEROP_LOG=1`. My reproducer had calls to `IsFunctionProtoType` and aborted because the return did not trace. I've added a simple check in the test to cover that:

```
EXPECT_FALSE(Cpp::IsFunctionProtoType(Cpp::ConstTypeRef()));
```

PR #1056, opened 2026-07-08
[interp] Resolve i386-COFF decorated symbols in clang-repl JIT

aaronj0:
On 32-bit MSVC the REPL JIT can't resolve process/DLL symbols: JIT'd code referencing the
CRT dies with JIT session error: Symbols not found: [ _memset, __CxxThrowException@8, ... ]
and GetFunctionAddress misses runtime-loaded exports. COFF-i386 names are decorated 
(`_cdecl`, `@N stdca`, `__imp_ dllimport`). ORC only strips prefixes on MachO and export
tables carry undecorated names so the symbol lookups fail.

Fix: add a DefinitionGenerator on the main JITDylib that undecorates and retries against loaded
modules, ported from cling's `platform::DLSym`: `strip __imp_`. manglings are undecorated and
stay with the default resolver. this is i386-only, cling mode unaffected

### DIFF
--- a/lib/CppInterOp/CppInterOpInterpreter.h
+++ b/lib/CppInterOp/CppInterOpInterpreter.h
@@ -41,6 +41,11 @@
 #include <sched.h>
 #include <unistd.h>
 #endif
+#if defined(_WIN32) && (defined(_M_IX86) || defined(__i386__))
+#include "llvm/ExecutionEngine/Orc/AbsoluteSymbols.h"
+#include "llvm/Support/DynamicLibrary.h"
+#include <deque>
+#endif
 #include <algorithm>
 #include <cstdio>
 #include <memory>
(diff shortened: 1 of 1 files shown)


lib/CppInterOp/CppInterOpInterpreter.h
```diff
@@ -144,6 +149,62 @@ inline clang::NamedDecl* Named(clang::Sema* S, const char* Name,
 
 namespace CppInternal {
 
+#if defined(_WIN32) && (defined(_M_IX86) || defined(__i386__))
+/// Fallback resolver for decorated COFF-i386 symbols, based on cling's
+/// platform::DLSym. i386 linker names are decorated ('_' cdecl prefix,
+/// '@N' stdcall suffix, '__imp_' dllimport indirection) and ORC strips
+/// such prefixes only on MachO, so JIT'd references to the CRT and to
+/// runtime-loaded DLLs fail to materialize. Undecorate and retry.
+class COFFi386SymbolGenerator : public llvm::orc::DefinitionGenerator {
```
vgvassilev, line 158: This is something we should bring up with Lang. Why upstream llvm is not sufficient...

aaronj0: Yes. The DLM tests works when run with Cling on Winx86, and this fix is based on Cling.  cc @lhames

edit: I will open an issue upstream with a repro

vgvassilev (approved):
Lgtm!

lib/CppInterOp/CppInterOpInterpreter.h
lhames: We should probably be addressing `__imp_` symbols at the JIT linker level, rather than in a generator. There are some PRs out for this that I need to look at.

For `_` and `@N` I _think_ we should be able to assert that these are resolved inside the JITDylib: external references should be `dllimported` and generate an `__imp_` symbol as above.

Issue #1057, opened 2026-07-08
[Bug]: Endless recursion in AnalyzeAllocType


keremsahn:
### Description of bug

If a func1 calls itself or calls another function named func2, which calls func1, then the analysis function recursively analyzes same functions over and over again. 

Solution: Create a map for visited FunctionDecls which will also cut repeated computation.

### What operating system was you using when the bug occured?

Ubuntu

### What is the architechture of the cpu on your system?

X86

PR #1064, opened 2026-07-08
Rebase computation under RASTV and support for recursion

keremsahn:
I moved helper functions and main analysis mechanism under RASTV struct. Also made almost every AllocType as std::optional, because to represent whether a variable or functions' AllocType value is still under computation or analysis hit a cycle, we needed another enum value, instead of this I made things std::optional . 

Solves #1057 

@Vipul-Cariappa @vgvassilev @aaronj0

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1637,98 +1637,127 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
   return INTEROP_RETURN(llvm::isa_and_nonnull<clang::FunctionProtoType>(T));
 }
 
-static AllocType handleNew(const clang::CXXNewExpr* CNE) {
-  if (CNE->getNumPlacementArgs() > 0)
-    return AllocType::None;
-  if (CNE->isArray())
-    return AllocType::NewArr;
-  return AllocType::New;
-}
+static std::optional<AllocType>
+AnalyzeAllocType(const clang::FunctionDecl* Fn,
+                 std::unordered_map<const clang::FunctionDecl*,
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1637,98 +1637,123 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
       return true;
+    if (!result.has_value())
+      result = tmp;
+    // If function's allocation behaviour differs between different cases,
+    // analyzer returns unknown.
+    else if (*result != tmp) {
+      result = AllocType::Unknown;
+      return false;
+    }
+    return true;
```
Vipul-Cariappa, line 1706: Can you please add a comment on how and why you are using the `result` variable.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1637,98 +1637,123 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
...
+  AllocationTraverser(std::unordered_map<const clang::FunctionDecl*,
+                                         std::optional<AllocType>>& cache)
+      : visitedFuncs(cache) {}
+
+  bool TraverseLambdaExpr(clang::LambdaExpr*) { return true; }
+
+  bool TraverseDecl(clang::Decl* D) {
+    if (D && llvm::isa<clang::TagDecl>(D))
+      return true;
+    return RecursiveASTVisitor::TraverseDecl(D);
```
Vipul-Cariappa, line 1665: Add a single line comment why this is done. This is an optimization.

Vipul-Cariappa (changes requested):

Vipul-Cariappa (approved):

PR #1065, opened 2026-07-09
FIX: Evaluating assignments on if-else branches

keremsahn:
Algorithm: 

Instead of copying whole varMap, each branch owns their own undoMap. What undoMap does is, 1)it keeps the original values in varMap to reverse the changes when if branch ends, 2) It keeps the original value of changed variables, and in varMap there is changed version, and when you compare values in varMap and undoMap you find out the deterministic result. undoMap is set as a struct member and keeps the track of undoMap of current branch, it is also kind of like a letter, which inner branches inform their upper branches about what they done, what would be the values if inner did not work.

### DIFF
--- a/include/CppInterOp/CppInterOpTypes.h
+++ b/include/CppInterOp/CppInterOpTypes.h
@@ -412,7 +412,10 @@ enum class AllocType : unsigned char {
   NewArr,
   Malloc,
   Unknown,
-  CustomAlloc
+  CustomAlloc,
+  Null,
+  OperatorNew,
+  OperatorNewArr
 };
 
 enum class DeallocType : unsigned char {
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1709,6 +1709,9 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
   // Result var keeps the combination all possible values of previous return
   // statements
   std::optional<AllocType> result;
+  // A map every branch writes the previous value of overwriten variables
+  std::unordered_map<const clang::VarDecl*, std::optional<AllocType>>* undoLog =
+      nullptr;
 
   AllocationTraverser(std::unordered_map<const clang::FunctionDecl*,
                                          std::optional<AllocType>>& cache)
@@ -1724,6 +1727,109 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
     return RecursiveASTVisitor::TraverseDecl(D);
   }
 
+  std::optional<AllocType> join(std::optional<AllocType> a,
+                                std::optional<AllocType> b) {
+    if (a == AllocType::Null)
+      return b;
+
+    if (b == AllocType::Null)
+      return a;
+
+    if (a == b)
+      return a;
+    return AllocType::Unknown;
+  }
+  void undoOnVarMap() {
+    for (auto& [VD, oldVal] : *undoLog) {
+      auto it = varMap.find(VD);
+      if (it == varMap.end())
+        continue;
+      varMap[VD] = oldVal;
+    }
+  }
+
+  bool TraverseIfStmt(clang::IfStmt* IS) {
+    TraverseStmt(IS->getConditionVariableDeclStmt());
+    TraverseStmt(IS->getCond());
+    auto* undoLogCopy = undoLog;
+    std::unordered_map<const clang::VarDecl*, std::optional<AllocType>>
+        undoLogThen;
+    undoLog = &undoLogThen;
+    TraverseStmt(IS->getThen());
+    auto* elseBranch = IS->getElse();
+    // There is no else
+    if (!elseBranch) {
+      for (auto& [VD, val] : undoLogThen) {
+        auto it = varMap.find(VD);
+        if (it == varMap.end())
+          continue;
+        // Overwrite varMap with join of overwriten and previous value
+        it->second = join(val, it->second);
+        // If branch is nested, inform upper branch about your changes
+        if (undoLogCopy)
+          undoLogCopy->try_emplace(VD, val);
+      }
+      undoLog = undoLogCopy;
+      return true;
+    }
+
+    // Save overwriten values in if branch to join
+    std::unordered_map<const clang::VarDecl*, std::optional<AllocType>>
+        valuesInIfBranch;
+    for (auto& [VD, val] : undoLogThen) {
+      auto it = varMap.find(VD);
+      if (it == varMap.end())
+        continue;
+      valuesInIfBranch[VD] = it->second;
+    }
+
+    // Take changes on varMap back before going to else branch
+    undoOnVarMap();
+
+    std::unordered_map<const clang::VarDecl*, std::optional<AllocType>>
+        undoLogElse;
+    undoLog = &undoLogElse;
+    TraverseStmt(elseBranch);
+
+    for (auto& [VD, val] : undoLogElse) {
+      auto it = varMap.find(VD);
+      if (it == varMap.end())
+        continue;
+      auto it2 = valuesInIfBranch.find(VD);
+      // If var is just changed in else branch
+      if (it2 == valuesInIfBranch.end()) {
+        it->second = join(val, it->second);
+        continue;
+      }
+      // If var is changed in both
+      it->second = join(it->second, it2->second);
+    }
+
+    for (auto& [VD, val] : valuesInIfBranch) {
+      auto it = varMap.find(VD);
+      if (it == varMap.end())
+        continue;
+      // If var is just changed in if branch
+      if (undoLogElse.find(VD) == undoLogElse.end()) {
+        it->second = join(val, it->second);
+        continue;
+      }
+      // If var is changed in both, here for extra safety
+      it->second = join(val, it->second);
+    }
+
+    // Inform upper branch about changes done for both inner if and else branch
+    if (undoLogCopy) {
+      for (auto& [VD, val] : undoLogThen)
+        undoLogCopy->try_emplace(VD, val);
+      for (auto& [VD, val] : undoLogElse)
+        undoLogCopy->try_emplace(VD, val);
+    }
+
+    undoLog = undoLogCopy;
+    return true;
+  }
+
   bool VisitVarDecl(VarDecl* VD) {
     Expr* expr = VD->getInit();
     if (expr)
@@ -1734,15 +1840,28 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
   }
 
   bool VisitBinaryOperator(clang::BinaryOperator* BO) {
-    if (BO->getOpcode() != BO_Assign)
-      return true;
     Expr* LHS = BO->getLHS();
     LHS = LHS->IgnoreParenCasts();
-    if (auto* DRE = dyn_cast<DeclRefExpr>(LHS)) {
-      if (auto* VD = dyn_cast<VarDecl>(DRE->getDecl())) {
-        Expr* RHS = BO->getRHS();
-        varMap[VD] = handleExpr(RHS);
-      }
+    auto* DRE = dyn_cast<DeclRefExpr>(LHS);
+    if (!DRE)
+      return true;
+
+    auto* VD = dyn_cast<VarDecl>(DRE->getDecl());
+    if (!VD)
+      return true;
+
+    if (BO->getOpcode() == BO_Assign) {
+      if (undoLog)
+        undoLog->try_emplace(VD, varMap[VD]);
+      Expr* RHS = BO->getRHS();
+      varMap[VD] = handleExpr(RHS);
+      return true;
+    }
+    if (BO->isCompoundAssignmentOp()) {
+      if (undoLog)
+        undoLog->try_emplace(VD, varMap[VD]);
+      varMap[VD] = AllocType::Unknown;
+      return true;
     }
     return true;
   }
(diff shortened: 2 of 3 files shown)


unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -981,14 +1207,30 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetAllocType) {
...
-  // FIXME: Pointer overwriten by a non-assignment operator
-  TESTAC(27, NewArr);
+  TESTAC(27, Unknown);
   TESTAC(28, New);
   TESTAC(29, New);
   TESTAC(30, New);
   TESTAC(31, New);
   TESTAC(32, New);
   TESTAC(33, New);
+  TESTAC(34, Unknown);
```
Vipul-Cariappa: This should return `New`, not `Unknown`. `nullptr` should be ignored as "someother type".

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -981,14 +1207,30 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetAllocType) {
...
+  TESTAC(35, New);
+  TESTAC(36, Unknown);
+  TESTAC(37, Unknown);
+  TESTAC(38, New);
+  TESTAC(39, Unknown);
+  TESTAC(40, Unknown);
+  TESTAC(41, Unknown);
+  TESTAC(42, Unknown);
```
Vipul-Cariappa: Should be `New`.

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -412,7 +412,10 @@ enum class AllocType : unsigned char {
   NewArr,
   Malloc,
   Unknown,
-  CustomAlloc
+  CustomAlloc,
+  Null,
+  OperatorNew,
+  OperatorNewArr
```
Vipul-Cariappa, line 418: For `CustomAlloc` and `OperatorNew`, we may also want an API to retrieve the allocating function. For example:
```c++
auto res = GetAllocType(F);
if (res == CustomAlloc || res == OperatorNew || res == OperatorNewArr)
  auto *AllocatingFunction = GetAllocatingFunction(F);
```
Also, `Null` should be second in the list, and `Unknown` the last.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -945,8 +946,340 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetAllocType) {
...
+    struct Box {
+    int* p;
+    Box(int* p) : p(p) {}
+    Box operator*(const Box& o) const { return Box(nullptr); }
+    };
+    Box func66() {
+      Box a(new int);
+      Box b(nullptr);
+      return a * b;
+    }
```
Vipul-Cariappa: What happens if
```c++
Box operator*(const Box& o) const {
  if (p && o.p)
    return new int((*p) * (*p.o));
  return nullptr;
}
```

---

**Let's not do the below now. We can open an issue for this:**
if above return `New`, then even the following should return `New`:
```c++
Box operator*(const Box& o) const {
  if (p && o.p)
    return Box(new int((*p) * (*p.o)));
  return nullptr;
}
```

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -945,8 +946,340 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetAllocType) {
...
+
+    int* func68(){
+      int tmp = 10;
+      int* m = (int*)tmp;
+      for(int i = 0; i < 10; m+=1){
+        return m;
+      }
+      return m;
+    }
```
Vipul-Cariappa: Interesting test. 😄

unittests/CppInterOp/FunctionReflectionTest.cpp
Vipul-Cariappa: What if:
```c++
Box func66() {
      Box a(new int);
      Box b(nullptr);
      return some_func(a, b);
    }
```

Vipul-Cariappa: What if:
```c++
Box func66() {
      Box a(new int);
      Box b(nullptr);
      return a.some_func(b);
    }
```

Vipul-Cariappa (commented):
I have only partially reviewed the PR:

unittests/CppInterOp/FunctionReflectionTest.cpp
keremsahn:
```cpp
LLVM_LIBC_FUNCTION(void *, lsearch,
                   (const void *key, void *base, size_t *nmemb, size_t size,
                    int (*compar)(const void *, const void *))) {
  if (key == nullptr || base == nullptr || nmemb == nullptr ||
      compar == nullptr)
    return nullptr;

  size_t byte_len = 0;
  if (mul_overflow(*nmemb, size, byte_len))
    return nullptr;

  const cpp::byte *next = reinterpret_cast<const cpp::byte *>(base);
  const cpp::byte *end = next + byte_len;
  //same problem as lfind, problem 4
  for (; next < end; next += size)
    if (compar(key, next) == 0)
      return const_cast<cpp::byte *>(next);

  *nmemb += 1;
  inline_memcpy(const_cast<cpp::byte *>(end), key, size);
  return const_cast<cpp::byte *>(end);
}
```
Not really, there was a real fail in libc and this is why it is there

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1722,13 +1836,35 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
...
+    // Expected: std/cpp::optional constructor
+    const clang::CXXConstructorDecl* CCD = CCE->getConstructor();
+    // optional(nullopt_t) noexcept; -> one arg
+    if (CCD->getNumParams() != 1)
+      return false;
+    clang::QualType ParamTy =
+        CCD->getParamDecl(0)->getType().getNonReferenceType();
+    // FIXME: Copy constructor of optional is not handled
+    if (const auto* CRD = ParamTy->getAsCXXRecordDecl())
+      return CRD->getName() == "nullopt_t";
```
Vipul-Cariappa: This looks like it handles only one edge case.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1747,8 +1883,44 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
...
+      // ExprWithCleanups -> ImplicitCastExpr -> CXXConstructExpr pattern
+      if (const auto* EWC = dyn_cast<clang::ExprWithCleanups>(stripped))
+        stripped = EWC->getSubExpr();
+      // We do not want to analyze explicit conversions
+      const auto* ICE = dyn_cast<clang::ImplicitCastExpr>(stripped);
+      bool isImplicitConv =
+          ICE && ICE->getCastKind() == CK_ConstructorConversion;
+      // ImplicitCastExpr -> CXXConstructExpr pattern
+      if (isImplicitConv)
+        return handleExpr(CCE->getArg(0));
```
Vipul-Cariappa: Can we add the constructor logic in a separate PR?
This PR does more than just if-else branch analysis.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1675,15 +1760,29 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
...
+    auto* DRE = dyn_cast<DeclRefExpr>(LHS);
+    if (!DRE)
+      return true;
+
+    auto* VD = dyn_cast<VarDecl>(DRE->getDecl());
+    // FIXME: BindingDecls are not handled
+    if (!VD)
+      return true;
+
+    if (BO->getOpcode() == BO_Assign) {
```
Vipul-Cariappa, line 1833:
```suggestion
    if (VD->getType()->isPointerType() && BO->getOpcode() == BO_Assign) {
```
Is this required?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1665,6 +1668,88 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
...
+      // yields to same
+      varMap[VD] = join(val, varMap[VD]);
+
+    // Inform upper branch about changes done for both inner if and else branch
+    if (undoLogCopy) {
+      for (auto& [VD, val] : undoLogThen)
+        undoLogCopy->try_emplace(VD, val);
+      for (auto& [VD, val] : undoLogElse)
+        undoLogCopy->try_emplace(VD, val);
+    }
```
Vipul-Cariappa, line 1806: I do not completely understand this part.
Why emplace both `if` and `else` branches? Shouldn't we emplace the computed results from `varMap` itself?

Vipul-Cariappa (changes requested):
We should split this PR into two. One for the if-else branch analysis, and another for the constructor analysis.
The if-else branch part looks good to me; I have one question about it, though. I have not looked at the tests; I will do it when the constructor related changes are deleted.

lib/CppInterOp/CppInterOp.cpp
keremsahn: Yeah, the record support is really messy and it is need to be discussed, i will tell my thoughts on it today

lib/CppInterOp/CppInterOp.cpp
keremsahn: It is not required, 
```cpp
  bool VisitVarDecl(VarDecl* VD) {
    Expr* expr = VD->getInit();
    if (expr)
      varMap[VD] = handleExpr(expr);
    else
      varMap[VD] = AllocType::None;
    return true;
  }
```
varMap also keeps non-pointer variables
int m = 5; for example, handleExpr returns AllocType::None for RHS, the extra guard you suggested will be a shortcut because every value m can have evaluates to same AllocType value, just a shortcut

lib/CppInterOp/CppInterOp.cpp
keremsahn: Computed results varMap carries this value: What would be the value of this variable if `this` branch is worked.
```cpp
    int* func(int val){
        int* ptr = nullptr;
        if(val > 0){
            if(val > 5)
                ptr = foo(val);
            else
                ptr = bar(val);
        }
        return ptr;
    }
```
Look at this example, the computed result from varMap will carry join of what foo and bar returns, but we should also know what would be ptr's value if outer if did not work. And when inner if overwrites ptr it will save nullptr value to evaluate what would result be if I did not work. In this case both inner if-else have same undoLog because they both write ptr and ptr's before value is same for both. But if inner if-else wrote to the different variables they both had to tell upper what they changed, so both undoLog needs to tell upper what they did, thats why there is two loop

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: > varMap also keeps non-pointer variables

Shouldn't we filter non-pointers?
I guess we will need this for record types.

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -973,23 +1253,48 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetAllocType) {
...
+  TESTAC(51, New);
+  TESTAC(52, New);
+  TESTAC(53, New);
+  TESTAC(54, Unknown);
+  TESTAC(55, OperatorNew);
+  TESTAC(56, OperatorNewArr);
+  TESTAC(57, None);
+  TESTAC(58, OperatorNew);
+  TESTAC(59, Malloc);
+  TESTAC(70, None);
```
Vipul-Cariappa: Shouldn't `70` be unknown?

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: I still do not understand this completely. But other than this, I believe this PR is fine. We can discuss this today.

unittests/CppInterOp/FunctionReflectionTest.cpp
Vipul-Cariappa: @keremsahn This isn't fixed. Otherwise, this PR is good to go.

Vipul-Cariappa (approved):
LGTM!

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -412,7 +412,10 @@ enum class AllocType : unsigned char {
   NewArr,
   Malloc,
   Unknown,
-  CustomAlloc
+  CustomAlloc,
```
Vipul-Cariappa, line 415: Realising this now: `CustomAlloc` is defined, but never used?

include/CppInterOp/CppInterOpTypes.h
Vipul-Cariappa: Please make a note of this.

include/CppInterOp/CppInterOpTypes.h
keremsahn: It is deleted in the next PR

include/CppInterOp/CppInterOpTypes.h
keremsahn: GetAllocatingFunction would have no use-case I guess, but maybe we can create another function which would return deallocator pair for the AllocType, to use in cppyy

PR #1066, opened 2026-07-14
Use `::new` in wrappers that placement-new into a caller buffer

guitargeek:
The wrappers generated by make_narg_ctor_with_return and make_narg_call_with_return spell placement new as `new (buf) C(...)`. For a class that declares a class-scope operator new without a matching placement form (e.g. a custom allocator), that class-scope declaration hides the global `operator new(size_t, void*)`, so the wrapper fails to compile ("no matching function for call to 'operator new'") and Cpp::Construct / by-value invocation fail for such classes.

Qualify the placement forms as `::new (buf) C(...)` so lookup bypasses class scope. The non-arena branch keeps the unqualified `new` on purpose: heap construction should still go through the class-scope allocator when one exists.

Covers all three sites: scalar placement construction, the `nary > 1` array branch, and the placement new that stores a by-value return into the caller-provided buffer.

Adds FunctionReflection_ConstructClassScopeNew, which declares a class with class-scope operator new/new[]/delete/delete[] and checks heap, placement, array-placement construction, and a by-value return; it fails to compile the wrappers without the fix.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -3677,7 +3677,10 @@ void make_narg_ctor_with_return(const FunctionDecl* FD, const unsigned N,
       callbuf << "if (nary > 1) {\n";
       indent(callbuf, indent_level);
       callbuf << "(*(" << class_name << "**)ret) = ";
-      callbuf << "(is_arena) ? new (*(" << class_name << "**)ret) ";
+      // Use ::new to construct in the arena: a class-scope operator new
+      // (e.g. a custom allocator with no matching placement form) would
+      // otherwise hide the global placement operator new.
+      callbuf << "(is_arena) ? ::new (*(" << class_name << "**)ret) ";
       make_narg_ctor(FD, N, typedefbuf, callbuf, class_name, indent_level,
                      true);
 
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3701,7 +3704,8 @@ void make_narg_ctor_with_return(const FunctionDecl* FD, const unsigned N,
     // : new ClassName(args...);
     indent(callbuf, indent_level);
     callbuf << "(*(" << class_name << "**)ret) = ";
-    callbuf << "(is_arena) ? new (*(" << class_name << "**)ret) ";
+    // ::new for the same reason as in the array branch above.
+    callbuf << "(is_arena) ? ::new (*(" << class_name << "**)ret) ";
```
vgvassilev, line 3975: Here the challenge is preserving the eventual user-intended semantics. Can we construct a case where we have a globally accessible operator new (or JitCall inside a class with operator new)?

guitargeek: Good point. I added `FunctionReflection_ConstructClassScopePlacementNew` to pin this down (see added commit).

vgvassilev (approved):
Lgtm!

PR #1067, opened 2026-07-15
Do not instantiate non-template static data members in GetVariableOffset

conrade-ctc:
`Cpp::GetVariableOffset` calls `Sema::InstantiateVariableDefinition` on any static data member whose address is not found and whose declaration has no initializer. That Sema entry point requires a *templated* variable: for a non-template class's static member, `getTemplateInstantiationPattern()` returns null, the assert is compiled out in release builds, and clang dereferences null (`this == nullptr` inside `VarDecl::getDefinition`).

The libstdc++ `<compare>` ordering types hit the exact triple condition — declared in-class without an initializer, defined out-of-line as `inline constexpr` (so no exported symbol to find), not yet JIT-emitted. Minimal repro on a release build:

```cpp
Cpp::Declare("#include <compare>");
Cpp::GetVariableOffset(Cpp::GetNamed("less", Cpp::GetScope("std::partial_ordering")));  // SIGSEGV
```

Per review (keeping the constexpr fast path cheap): an already-parsed out-of-line definition is now resolved *first* — it may carry the initializer the constexpr evaluation consumes, with no Sema work at all — and the pattern-guarded `InstantiateVariableDefinition` is only the fallback for genuinely templated statics. `getDefinition()` is null-checked in both arms; the existing fallthrough then does the right thing for these variables (discardable-ODR linkage → codegen emission → real address). The regression test crashes at exactly this stack without the fix and passes with it.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -2779,9 +2779,21 @@ intptr_t GetVariableOffset(compat::Interpreter& I, Decl* D,
       address = I.getAddressOfGlobal(GD);
     if (!address) {
       if (!VD->hasInit()) {
-        compat::SynthesizingCodeRAII RAII(&getInterp());
-        getSema().InstantiateVariableDefinition(SourceLocation(), VD);
-        VD = VD->getDefinition();
+        // The initializer feeding the constexpr fast path below may live on
+        // an already-parsed out-of-line definition (a non-template class's
+        // static data member, e.g. std::partial_ordering::less): prefer it,
+        // with no Sema work at all. Only a variable instantiated from a
+        // template can need Sema::InstantiateVariableDefinition — and
+        // without an instantiation pattern that call dereferences a null
+        // VarDecl in release builds.
+        if (VarDecl* Def = VD->getDefinition()) {
+          VD = Def;
+        } else if (VD->getTemplateInstantiationPattern()) {
+          compat::SynthesizingCodeRAII RAII(&getInterp());
+          getSema().InstantiateVariableDefinition(SourceLocation(), VD);
+          if (VarDecl* Inst = VD->getDefinition())
+            VD = Inst;
+        }
       }
       if (VD->hasInit() &&
           (VD->isConstexpr() || VD->getType().isConstQualified())) {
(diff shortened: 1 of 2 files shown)


conrade-ctc:
cc @vgvassilev @rathorey-ctc

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2779,9 +2779,18 @@ intptr_t GetVariableOffset(compat::Interpreter& I, Decl* D,
...
+        // instantiation and just look up the out-of-line definition.
+        if (VD->getTemplateInstantiationPattern()) {
+          compat::SynthesizingCodeRAII RAII(&getInterp());
+          getSema().InstantiateVariableDefinition(SourceLocation(), VD);
+        }
+        if (VarDecl* Def = VD->getDefinition())
+          VD = Def;
       }
       if (VD->hasInit() &&
           (VD->isConstexpr() || VD->getType().isConstQualified())) {
```
vgvassilev, line 2799: The constexpr branch is critical for performance. Can the new code come after it?

conrade-ctc:
CI triage for the two red jobs — neither traces to this diff:

- **ubu24-x86-gcc14-cling-llvm20-root**: ROOT configure fails with `ninja: build.ninja: etc/cppinterop/CppInterOp/Box.h is defined as an output multiple times` — same failure on other open PRs (e.g. #1064's last run).
- **ubu24-x86-gcc14-cling-llvm20-asan**: LSAN exit-leak report, byte-identical to a failure on unchanged `main` (nightly run https://github.com/compiler-research/CppInterOp/actions/runs/29124927678): 74184/57 + 29144/1 + 25758/58 … = 269,234 bytes in 231 allocations, all indirect, allocated under an unrelated test's `CreateInterpreter`. The new tests may perturb how often the pre-existing leak manifests, but the leaked object predates this change.

All cling/clang-repl functional suites pass.

lib/CppInterOp/CppInterOp.cpp
conrade-ctc: updated: the constexpr branch hasn't moved, and its hot path runs no new code: the guarded block sits inside `if (!VD->hasInit())`, and when there's no initializer the constexpr branch has nothing to evaluate yet — the instantiation exists to materialize the initializer that branch consumes, so it can't come after it. But the performance concern did point at a real win: upstream ran `InstantiateVariableDefinition` unconditionally (transaction push + Sema instantiation machinery). Reworked so an already-parsed out-of-line definition is preferred first — `std::partial_ordering::less` now reaches the constexpr evaluation with no Sema work at all — and the pattern-guarded instantiation is only the fallback for genuinely templated statics.

vgvassilev (approved):
LGTM!

PR #1068, opened 2026-07-15
Emit discardable-ODR variables via odr-use instead of UsedAttr in GetVariableOffset

conrade-ctc:
`ForceCodeGen` forces deferred emission by planting a permanent `__attribute__((used))` on the decl. Every global emitted that way gets a `WeakTrackingVH` in codegen's `llvm.used` list; when such a global is deleted before `emitUsed` runs (e.g. ORC freeing a materialized PTU's IR), the handle nulls and release-built clang dereferences it unchecked — SIGSEGV in `ConstantExpr::getPointerBitCastOrAddrSpaceCast` under `CodeGenModule::Release()` of a later incremental parse. The failure is **cumulative interpreter state, not a culprit type**: it only manifests after enough force-emitted statics plus JIT materialization cycles (we hit it on large reflection sweeps; per-type bisection never converged), so the regression test asserts the mechanism instead — no `UsedAttr` planted, no `llvm.used` residue in later PTU modules — plus a sweep-shaped smoke loop.

`GetVariableOffset` now emits `GVA_DiscardableODR` variables (the inline/constexpr class the crash traced to) by declaring an external-linkage odr-use of the qualified name: the definition flows through the regular deferred-decl path and leaves no used-list residue and no permanent attribute. Everything else deliberately stays on the `UsedAttr` path:

- Internal-linkage variables cannot be odr-used from a later PTU (the module-local symbol duplicates or goes missing) — getting this wrong broke `VariableReflection_GetVariableOffset` (`static int S`) and cppyy's `Lifeline::count` lookup.
- Available-externally definitions would not be emitted by a mere reference.
- Template-specialization and anonymous/lambda spellings do not reliably round-trip as source (a printed `LLONG_MIN` non-type argument re-parses as an overflowing literal), and a parse-failing declaration poisons the incremental interpreter.

Also evaluated and rejected: `Interpreter::Undo(1)` on parse failure to drop failed PTUs — cppyy depends on failed parses leaving side-effect declarations (`test_templates` test32's explicit-instantiation probe).

**Concurrency:** the dummy-name counter is deliberately a plain `static unsigned`, matching the file's existing `gWrapperSerial` idiom. The interpreter API is caller-serialized (the incremental parse either route drives is not thread-safe, and the file's only sync primitive is the process-init `once_flag`); an atomic here would advertise a guarantee the surrounding function doesn't have. Worst case if raced anyway: a duplicate dummy name fails the `Declare` and the caller falls back to `ForceCodeGen`.

Known residual, unchanged here: `ForceCodeGen`'s *function* path (`GetFunctionAddress`) still uses `UsedAttr` — same theoretical hazard, never observed.

Independent of #1067 (both touch `GetVariableOffset`; whichever lands second has a trivial test-file conflict).

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -59,6 +59,7 @@
 #include "clang/Basic/CharInfo.h"
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/DiagnosticSema.h"
+#include "clang/Basic/LLVM.h"
 #include "clang/Basic/LangStandard.h"
 #include "clang/Basic/Linkage.h"
 #include "clang/Basic/OperatorKinds.h"
@@ -400,6 +401,91 @@ static void ForceCodeGen(Decl* D, compat::Interpreter& I) {
     llvm::logAllUnhandledErrors(std::move(Err), llvm::errs(),
                                 "[ForceCodeGen] Failed to execute PTU:");
 #endif
+}
+
+// Force emission of a variable's definition by synthesizing an odr-use of
+// it, instead of ForceCodeGen's UsedAttr route. The UsedAttr route records
+// the emitted global in codegen's llvm.used list as a weak handle; the
+// handle can go null (global deleted) before a later incremental PTU's
+// emitUsed runs, and release-built clang dereferences it without checking —
+// a process crash that accumulates with interpreter state rather than
+// tracing to any one declaration. The odr-use anchor is a dummy global
+// initialized with the variable's address, built directly through Sema (the
+// AST that parsing "&var" would produce — no source-text round-trip), so
+// the definition flows through the regular deferred-decl path and leaves no
+// used-list residue. Returns false when the odr-use cannot be built — the
+// caller falls back to ForceCodeGen.
+// FIXME: Remove the synthesized-anchor mechanism once clang grows a direct
+// interpreter emission API for a GlobalDecl (the emitUsed hardening it
+// depends on landed via llvm/llvm-project#210959):
+#if CLANG_VERSION_MAJOR >= 24
+#warning "Revisit EmitVariableViaOdrUse: a direct-emission API may exist now"
+#endif
+static bool EmitVariableViaOdrUse(compat::Interpreter& I, VarDecl* VD) {
+  // A reference cannot conjure a definition that does not exist: the dummy
+  // would carry an unresolvable symbol into the JIT (Emscripten's dynamic
+  // loader rejects the whole module over it, and the stale entry then
+  // breaks every later load in the process). The UsedAttr fallback defers
+  // harmlessly for definition-less declarations.
+  VarDecl* Def = VD->getDefinition();
+  if (!Def)
+    return false;
+
+  Sema& S = I.getCI()->getSema();
+  ASTContext& C = S.getASTContext();
+
+  // Open the synthesizing region before ANY Sema work: odr-use marking in
+  // BuildDeclRefExpr can trigger an immediate static-member instantiation,
+  // and cling's DeclCollector aborts on instantiation callbacks that arrive
+  // outside an active transaction. (Inert under clang-repl — TODO destructor
+  // in Compatibility.h.)
+  compat::SynthesizingCodeRAII RAII(&I);
+
+  // Build '&VD' the way the parser would; the address-of marks Def
+  // odr-used, which is what schedules the deferred definition.
+  ExprResult Ref = S.BuildDeclRefExpr(Def, Def->getType().getNonReferenceType(),
+                                      VK_LValue, SourceLocation());
+  if (Ref.isInvalid())
+    return false;
+  ExprResult AddrOf =
+      S.CreateBuiltinUnaryOp(SourceLocation(), UO_AddrOf, Ref.get());
+  if (AddrOf.isInvalid())
+    return false;
+
+  // Anchor the odr-use in an external-linkage dummy the JIT must emit; the
+  // initializer's implicit conversion to 'const void*' is Sema-checked, so
+  // exotic types fail over to the fallback instead of asserting.
+  // FIXME: The synthesized nodes are parked in the TU for the rest of the
+  // session; move them under a scratch area whose AST nodes get deallocated
+  // after emission.
+  std::string DummyName = "__cppinterop_odr_use_v" +
+                          std::to_string(getInterpInfo(&I).OdrUseCounter++);
+  QualType Ty = C.getPointerType(C.VoidTy.withConst());
+  TranslationUnitDecl* TU = C.getTranslationUnitDecl();
+  VarDecl* Dummy = VarDecl::Create(C, TU, SourceLocation(), SourceLocation(),
+                                   &C.Idents.get(DummyName), Ty,
+                                   C.getTrivialTypeSourceInfo(Ty), SC_None);
+  S.AddInitializerToDecl(Dummy, AddrOf.get(), /*DirectInit=*/false);
+  if (Dummy->isInvalidDecl())
+    return false;
+  TU->addDecl(Dummy);
+
+  // Under cling the RAII's transaction commit (at scope exit) triggers
+  // emission of everything collected above plus this dummy.
+  I.getCI()->getASTConsumer().HandleTopLevelDecl(DeclGroupRef(Dummy));
+#ifndef CPPINTEROP_USE_CLING
+  // FIXME: Parsing an empty string is the only way to flush incremental
+  // CodeGen for a decl handed straight to the consumer — the state-reset
+  // bug SynthesizingCodeRAII's clang-repl destructor should eventually
+  // own. Drop this when it does.
+  auto GeneratedPTU = I.Parse("");
+  if (llvm::Error Err =
+          !GeneratedPTU ? GeneratedPTU.takeError() : I.Execute(*GeneratedPTU)) {
+    llvm::consumeError(std::move(Err));
+    return false;
+  }
+#endif
+  return true;
 }
 
 #define DEBUG_TYPE "jitcall"
@@ -2806,7 +2892,14 @@ intptr_t GetVariableOffset(compat::Interpreter& I, Decl* D,
     }
     if (!address) {
       auto Linkage = C.GetGVALinkageForVariable(VD);
-      if (isDiscardableGVALinkage(Linkage))
+      // Odr-use emission only for discardable-ODR entities (inline/constexpr
+      // statics) — the class the used-list crash traced to. Internal-linkage
+      // variables cannot be odr-used from a later PTU (module-local symbol:
+      // the reference duplicates or misses the entity), and an
+      // available-externally definition would not be emitted by a mere
+      // reference; both stay on the stock UsedAttr path.
+      if (isDiscardableGVALinkage(Linkage) &&
+          (Linkage != GVA_DiscardableODR || !EmitVariableViaOdrUse(I, VD)))
         ForceCodeGen(VD, I);
     }
     auto VDAorErr = compat::getSymbolAddress(I, StringRef(mangledName));
@@ -4989,7 +5082,7 @@ protected:
 };
 } // namespace
 
-int Declare(compat::Interpreter& I, const char* code, bool silent) {
+static int Declare(compat::Interpreter& I, const char* code, bool silent) {
   // Trap diagnostics on both paths: I.declare's rc is 0 even when
   // Parse recovered from emitted errors, so callers need the trap to
   // distinguish "parsed cleanly" from "parsed with errors".
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -59,6 +59,7 @@
 #include "clang/Basic/CharInfo.h"
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/DiagnosticSema.h"
+#include "clang/Basic/LLVM.h"
 #include "clang/Basic/LangStandard.h"
 #include "clang/Basic/Linkage.h"
 #include "clang/Basic/OperatorKinds.h"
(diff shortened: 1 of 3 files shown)


conrade-ctc:
cc @vgvassilev @rathorey-ctc

vgvassilev:
This entire approach of adding these attributes to force emission is sketchy to say the least. Can you ask Claude what alternatives we have including perhaps minimal reasonable modifications to clang.

conrade-ctc:
> This entire approach of adding these attributes to force emission is sketchy to say the least. Can you ask Claude what alternatives we have including perhaps minimal reasonable modifications to clang.

ya, i didn't love this at all either... felt pretty gross to me. let me iterate a bit and see what we might do upstream to address it.

vgvassilev:
> > This entire approach of adding these attributes to force emission is sketchy to say the least. Can you ask Claude what alternatives we have including perhaps minimal reasonable modifications to clang.
> 
> ya, i didn't love this at all either... felt pretty gross to me. let me iterate a bit and see what we might do upstream to address it.

I did not mean your patch but the overall logic in this area.

conrade-ctc:
CI triage for the three red jobs — none traces to this diff:

- **osx26-x86-clang-llvm21-cppyy**: runner network flake — CppInterOp's own tests passed 7/7, then the cppyy-stack step died on `git clone`: `Could not resolve host: github.com`. A rerun should clear it (I don't have rerun rights here).
- **ubu24-x86-gcc14-cling-llvm20-root**: pre-existing ROOT-configure ninja failure (`Box.h defined as an output multiple times`), same on other open PRs.
- **ubu24-x86-gcc14-cling-llvm20-asan**: pre-existing intermittent LSAN exit-leak, byte-identical (269,234 bytes / 231 allocations) to a failure on unchanged `main` (https://github.com/compiler-research/CppInterOp/actions/runs/29124927678); details in the sibling comment on #1067.

The six cling compile failures from the first push are fixed (PTU white-box assertions now gated to clang-repl builds).

conrade-ctc:
Asked Claude to survey the design space; here's the grounded version (file/line refs against release/21.x).

**Why the current approach is sketchy (agreed):** `ForceCodeGen` mutates the AST as a side effect of a reflection query — `UsedAttr::CreateImplicit` is visible to every later semantic check — and the attr drags the emitted global into `llvm.used`, a `std::vector<llvm::WeakTrackingVH>`. Incremental PTUs can delete such a global, the handle goes null, and `emitUsed` does `cast<llvm::Constant>(&*List[i])` with no null check (`CodeGenModule.cpp:~3203`) — that's the crash this PR routes around. On top, the generated module's internal-linkage globals get rewritten to weak.

**What CppInterOp can do without touching clang:** essentially this PR — a source-level odr-use where the qualified name round-trips (falling back to `UsedAttr` for template-specialization spellings and anonymous scopes). The semantically honest AST-level version — `Sema::MarkVariableReferenced` + `HandleTopLevelDecl` — dead-ends because `ASTContext::DeclMustBeEmitted` doesn't consult `isUsed()` for variables (the FIXME in `ForceCodeGen` says exactly this), so codegen still defers the definition.

**Minimal clang modification #1 (tiny, worth doing regardless):** null-skip deleted globals in `emitUsed`'s loop. In incremental mode a used global can legitimately be destroyed between PTUs; today that's a release-build null deref. ~3 lines, protects every existing embedder of the `UsedAttr` pattern — including this PR's own fallback branch — and needs no design discussion. I'll send that patch independently.

**Minimal clang modification #2 (the principled end-state):** an emission API on `clang::Interpreter` — e.g. `emitGlobalDecl(GlobalDecl)`. The pieces already exist: `CodeGenerator::GetAddrOfGlobal(GD, /*isForDefinition=*/true)` is public (`ModuleBuilder.h:98`) and schedules the deferred definition (`GetOrCreateLLVMGlobal` promotes the `DeferredDecls` entry, which `moveLazyEmissionStates` explicitly carries across incremental modules — `CodeGenModule.cpp:8111`); all that's missing is that `Interpreter::getCodeGen()` is private, so embedders can't reach it. Such an API would emit with natural `linkonce_odr` linkage — no attribute, no `llvm.used` residue, no name round-trip, works for the template/anonymous shapes this PR has to punt on — and would let `ForceCodeGen`'s whole `UsedAttr` + `Parse("")` + linkage-rewrite dance be deleted.

**Suggested sequencing:** land this PR now as the crash fix — it helps every clang version that exists today, and CppInterOp's supported-version window means the eventual cleanup is a version gate rather than a revert (prefer `emitGlobalDecl` under `#if CLANG_VERSION_MAJOR >= N` once the API lands, keep this mechanism as the legacy branch for older majors). The `emitUsed` null-skip goes upstream immediately on its own; `emitGlobalDecl` goes through design review on its own cycle. Happy to draft both clang patches.

vgvassilev:
Sounds like a good plan actually.

conrade-ctc:
Rebased onto main to clear the conflict with #1067 — its definition-first structure is preserved, with the odr-use dispatch layered on top of the discardable-linkage branch; both PRs' tests kept. Gates green locally (check-cppinterop, both `GetVariableOffset` tests pass in InProcess and OutOfProcess JIT modes; clang-format clean).

Step 2 of the plan is also up: llvm/llvm-project#210959 null-skips deleted globals in `emitUsed`. While writing it I found a pure clang-repl reproducer — no CppInterOp needed: `GetOrCreateLLVMGlobal` erases a same-mangled-name/different-type global without RAUW when it has no IR uses, and value handles aren't uses, so the used-list entry nulls instead of retargeting:

```
clang-repl> __attribute__((used)) int a asm("sym") = 1; float b asm("sym") = 2.0f;
clang-repl> int ok = 0;   // Assertion `V && "Dereferencing deleted ValueHandle"' failed
```

@vgvassilev — ready for another look.


CI note: the three emscripten job failures are infra — micromamba env setup got an HTTP 500 from prefix.dev fetching repodata, before anything was compiled. They should clear on a rerun (I lack rerun rights here). The root job is the known pre-existing configure breakage.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -402,6 +402,45 @@ static void ForceCodeGen(Decl* D, compat::Interpreter& I) {
...
+// used-list residue. Returns false when the variable cannot be named from
+// a fresh chunk of source (anonymous scopes, or printed names that fail to
+// parse) — the caller falls back to ForceCodeGen.
+static bool EmitVariableViaOdrUse(compat::Interpreter& I, VarDecl* VD) {
+  std::string name;
+  {
+    llvm::raw_string_ostream OS(name);
+    VD->printQualifiedName(OS);
+  }
+  if (name.find("(anonymous ") != std::string::npos ||
```
vgvassilev: We should not do string comparisons but we should use the relevant api to get the same information from the ast/sema.

conrade-ctc: Done in 5f96515 — the string sniffing is gone. The guard now asks the AST directly: bail out for declarations without an `IdentifierInfo` or that are `VarTemplateSpecializationDecl`s, then walk the `DeclContext` chain rejecting anonymous namespaces, unnamed/lambda records, `ClassTemplateSpecializationDecl` scopes, and anything else that isn't a named namespace/record (linkage-spec/export blocks are skipped as transparent). `printQualifiedName` is now only used to spell the odr-use for names the walk has proven spellable. Added test coverage for the anonymous-namespace bailout and the linkage-spec passthrough.

conrade-ctc:
CI triage on the last run, all addressed or external:

- **emscripten lanes**: real failure, but an interaction — #1067's `NonTemplateStaticNoInit` (in via the rebase) force-emits a never-defined static, which under wasm poisons the shared interpreter: every later incremental module load fails with `bad export type for '_ZN12NeverDefined7missingE'`, and this PR's test runs right after it. The stale-handle scenario is native-JIT mechanics anyway, so 0b6e999 skips the test on `__EMSCRIPTEN__` (repo idiom). The underlying wasm poisoning predates this PR and reproduces without it — filed as #1071.
- **asan lane**: all 274 tests pass; the job fails on an exit-time LeakSanitizer report whose stacks trace to interpreter creation in `Interpreter_DLMIsPerInterpreter` (untouched here; same lane was red on #1067 pre-merge).
- **root lane**: pre-existing configure failure (ROOT rejects all available LLVMConfig versions).
- **codecov**: only one coverage session uploaded on the last run; will re-check on this head.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -402,6 +403,62 @@ static void ForceCodeGen(Decl* D, compat::Interpreter& I) {
...
+    llvm::raw_string_ostream OS(name);
+    VD->printQualifiedName(OS);
+  }
+
+  // External linkage on the dummy keeps it — and therefore the referenced
+  // definition — from being discarded as unused internal state.
+  static unsigned Counter = 0;
+  std::string code = "namespace __cppinterop_odr_use { const void* __v" +
+                     std::to_string(Counter++) +
+                     " = (const void*)__builtin_addressof(::" + name + "); }";
```
vgvassilev: Can we avoid that string manipulation, too? This leads to memory hoarding on long sessions... Perhaps we should just do what `__builtin_addressof` would do by calling it directly in Sema if possible...

conrade-ctc: Done in e11ca91 — the source-text mechanism is gone entirely. The odr-use anchor is now built directly in Sema: `BuildDeclRefExpr` + `CreateBuiltinUnaryOp(UO_AddrOf)` (the AST parsing `&var` would produce), stored into a synthesized external-linkage `const void*` dummy at TU scope whose initializer conversion Sema checks, then handed to codegen the way `ForceCodeGen` already does (`HandleTopLevelDecl` + empty-parse flush). No per-query source buffers or string assembly beyond the dummy's identifier.

Two knock-on wins:
- Template-specialization members now take the odr-use route as well — no spelling round-trip exists to fail, and the caller has already instantiated the definition (#1067's shape). Test expectations updated accordingly.
- A new definition-exists guard bails to the UsedAttr fallback for declared-but-never-defined variables, so the dummy can never carry an unresolvable reference into the JIT. That reference was what the emscripten lanes were dying on — a failed wasm module load poisons process-global dynamic-linking state for everything after it (filed as #1071; the wasm skip on the regression test stays since the scenario it pins is native-ORC mechanics).

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -402,6 +403,76 @@ static void ForceCodeGen(Decl* D, compat::Interpreter& I) {
...
+  VarDecl* Def = VD->getDefinition();
+  if (!Def)
+    return false;
+
+  Sema& S = I.getCI()->getSema();
+  ASTContext& C = S.getASTContext();
+
+  // Build '&VD' the way the parser would; the address-of marks Def
+  // odr-used, which is what schedules the deferred definition.
+  ExprResult Ref = S.BuildDeclRefExpr(Def, Def->getType().getNonReferenceType(),
```
vgvassilev, line 446: This still hoards much less memory though. Can we add a fixme note with two lines annotating that we want to probably move that under a scratch area where these ast nodes get deallocated and add a ifdef llvm_version < 24 to remind ourselves to remove it once the other fix lands hopefully upstream?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -402,6 +403,76 @@ static void ForceCodeGen(Decl* D, compat::Interpreter& I) {
...
+    return false;
+  TU->addDecl(Dummy);
+
+#ifdef CPPINTEROP_USE_CLING
+  cling::Interpreter::PushTransactionRAII RAII(&I);
+  I.getCI()->getASTConsumer().HandleTopLevelDecl(DeclGroupRef(Dummy));
+  return true;
+#else // CLANG_REPL
+  I.getCI()->getASTConsumer().HandleTopLevelDecl(DeclGroupRef(Dummy));
+  auto GeneratedPTU = I.Parse("");
```
vgvassilev, line 481: That’s worth another fixme pointing out that bug which forces parsing an empty string to reset the state in the compiler. I thought we had some raii to use to annotate these places.

lib/CppInterOp/CppInterOp.cpp
conrade-ctc: Done in d317206 — two FIXMEs added: one for moving the synthesized nodes under a deallocatable scratch area, one for removing the whole anchor mechanism once clang grows a direct emission API, with a `#if CLANG_VERSION_MAJOR >= 24` + `#warning` tripwire so the clang-24 bump forces the revisit.

lib/CppInterOp/CppInterOp.cpp
conrade-ctc: The RAII you're thinking of is `compat::SynthesizingCodeRAII` — under cling it's `PushTransactionRAII` (commit = emission), but its clang-repl variant is a stub with a TODO destructor (Compatibility.h). d317206 now uses it on both paths as the annotation, and the empty-parse flush carries a FIXME naming it as the thing the RAII's clang-repl destructor should eventually own.

unittests/CppInterOp/VariableReflectionTest.cpp
```diff
@@ -376,6 +382,147 @@ TYPED_TEST(CPPINTEROP_TEST_MODE,
...
+    decl += "; };";
+    Cpp::Declare(decl.c_str());
+    Cpp::DeclRef k = Cpp::GetNamed("Sweep" + n);
+    ASSERT_TRUE(k);
+    Cpp::DeclRef v = Cpp::GetNamed("v" + n, k);
+    ASSERT_TRUE(v);
+    EXPECT_TRUE(Cpp::GetVariableOffset(v));
+    std::string probe = "template <> struct Sweep";
+    probe += n;
+    probe += "<int>;";
```
vgvassilev: Is there any value in accumulating in this way? Why can't we do `R"(...)`?

conrade-ctc: Sorry, this seems to be a claude default, and I didn't catch it, shouldn't be building the string this way. I'll update this and push, and refine CLAUDE.md!

conrade-ctc: No value in it — that shape was appeasing clang-tidy's inefficient-concatenation check, at readability's expense. Reworked in the current head: the sweep bodies are raw strings (`R"(struct Sweep@ { inline static int v@ = @; };)"`) with a tiny `@`→index substitution, no accumulation.

vgvassilev: I still don't get it -- can't we have a normal preprocessor macro with a stringify operation?

conrade-ctc: Done in 60c19fe — the sweep sources are now assembled entirely at compile time: a `SWEEP_CASE(n)` macro stringifies the index (`#n`) into adjacent string literals, expanded into a `constexpr std::array` for 0–7. No runtime string building left in the test.

Same push also fixes what the osx cling lanes caught on the previous head: under cling, the odr-use marking in `BuildDeclRefExpr` can trigger an immediate static-member instantiation, and `DeclCollector` aborts if no transaction is open — the `SynthesizingCodeRAII` now opens before any Sema work in `EmitVariableViaOdrUse` (same pattern as the `InstantiateVariableDefinition` call site from #1067).

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -402,6 +403,94 @@ static void ForceCodeGen(Decl* D, compat::Interpreter& I) {
...
+  if (AddrOf.isInvalid())
+    return false;
+
+  // Anchor the odr-use in an external-linkage dummy the JIT must emit; the
+  // initializer's implicit conversion to 'const void*' is Sema-checked, so
+  // exotic types fail over to the fallback instead of asserting.
+  // FIXME: The synthesized nodes are parked in the TU for the rest of the
+  // session; move them under a scratch area whose AST nodes get deallocated
+  // after emission.
+  static unsigned Counter = 0;
```
vgvassilev: That will hurt the multiple interpreters setup. We need to move that variable in the InterpreterInfo object.

conrade-ctc: Hmmm, we must not have a test that exposes that. I'll try to add one, and fix. Thanks for all the iteration on this PR!

conrade-ctc: Done in d3bd5d3 — `OdrUseCounter` now lives in `InterpreterInfo` (accessed via `getInterpInfo(&I)`, and carried through the move ops since `DeleteInterpreter`'s mid-deque erase relocates later entries). Added a regression test first, which fails with the static: `VariableReflection_GetVariableOffset_OdrUseAnchorPerInterpreter` pins that a fresh interpreter's first anchor is always its own `v0`, independent of sibling interpreters' (or the process's) query history.

Same push covers the anchor's error paths for the codecov patch gate: the Parse/Execute failure arms now share one consume-and-fail block, exercised by a new `UnmaterializableDefinition` test (a definition whose dynamic initializer needs an undefined symbol — parses fine, materialization fails, the query fails cleanly).


(Amended 58b9a90 → 97716d3: the new failure-path test provokes an intentional JIT error whose "JIT session error :" stderr line MSBuild promotes to a build error on the win-msvc lane — the test now skips on `_WIN32`; coverage comes from the linux lanes.)

(Amended 97716d3 → d3bd5d3: the same test now also skips on `__EMSCRIPTEN__` — its intentional failed load poisons wasm dynamic linking, #1071 — and the Windows skip message no longer itself contains the `error :` token MSBuild scrapes.)

conrade-ctc:
@vgvassilev gentle ping — all review rounds are addressed (the counter now lives in per-interpreter state, with the regression test you asked for) and CI is green apart from the known root-lane failure that also fails on main. Ready for another look when you have a chance. If you have spare cycles, #1069 and #1070 have also been reworked per your feedback and are awaiting re-review since the 16th. Thanks!

vgvassilev (approved):
LGTM!

conrade-ctc:
@vgvassilev CI is green now. The red lanes were the out-of-process parameterization of the PR's new negative test: the intentional materialization failure makes the remote `dlupdate` fail, and the session blocks instead of reporting the error, so the binary hung until the job cap. The test now skips out-of-process — the same guard class as its existing wasm and Windows skips — and stays fully covered in-process. Are we good to go with an all green CI?

PR #1069, opened 2026-07-16
Format function arg defaults with the constants-as-written printer

conrade-ctc:
`GetFunctionArgDefault` re-printed default arguments at maximum float precision, so `3.14` rendered as `3.1400000000000001`. Print them with `printPretty`, the interpreter's `ASTContext`, and `ConstantsAsWritten = true` instead: literal leaves keep their source spelling (`4.0` stays `4.0`).

Two clang-side gaps remain: the maximum-precision expansion where `ConstantsAsWritten` cannot apply (fix under review in llvm/llvm-project#218471), and cooked UDL floats (`90.0_deg` prints as `90._deg`). A canary test expects the old output before clang 24 and the fixed output from clang 24 on.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -68,6 +68,7 @@
 #include "clang/Basic/Version.h"
 #include "clang/Frontend/CompilerInstance.h"
 #include "clang/Interpreter/Interpreter.h"
+#include "clang/Lex/Lexer.h"
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/Overload.h"
 #include "clang/Sema/Ownership.h"
@@ -5485,6 +5486,32 @@ std::string GetFunctionArgDefault(ConstFuncRef func, size_t param_index) {
     PI = (FD->getTemplatedDecl())->getNonObjectParameter(param_index);
 
   if (PI->hasDefaultArg()) {
+    // Render the default as written: read its source range, the way clang's
+    // own code completion formats default arguments (GetDefaultValueString
+    // in SemaCodeComplete). Pretty-printing the AST instead reproduces
+    // floating literals at representation precision ("3.1400000000000001"
+    // for a written 3.14), and normalizing that through std::stod terminated
+    // the exception-free build on symbolic defaults such as
+    // `double ratio = kDefaultRatio`.
+    Sema& S = getSema();
+    CharSourceRange SrcRange =
+        CharSourceRange::getTokenRange(PI->getDefaultArgRange());
+    bool Invalid = SrcRange.isInvalid();
+    if (!Invalid) {
+      llvm::StringRef SrcText = Lexer::getSourceText(
+          SrcRange, S.getSourceManager(), S.getLangOpts(), &Invalid);
+      if (!Invalid) {
+        // The lexed range sometimes includes the leading '=' (see the FIXME
+        // in SemaCodeComplete's GetDefaultValueString); keep only the value.
+        SrcText = SrcText.ltrim();
+        SrcText.consume_front("=");
+        SrcText = SrcText.ltrim();
+        if (!SrcText.empty())
+          return INTEROP_RETURN(SrcText.str());
+      }
+    }
+
+    // Fallback for defaults with no usable source range: print the AST.
     std::string Result;
     llvm::raw_string_ostream OS(Result);
     const Expr* DefaultArgExpr = nullptr;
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -5502,10 +5503,21 @@ std::string GetFunctionArgDefault(ConstFuncRef func, size_t param_index) {
...
-      Result = oss.str();
+      // The printed default need not be a numeric literal (an identifier
+      // such as `kDefaultRatio`, or a call such as `quiet_NaN()`); std::stod
+      // would throw std::invalid_argument, and this library is built
+      // without exception support, so the throw terminates the process.
+      // Parse non-throwingly and keep the printed spelling when the default
+      // is not a plain literal.
+      errno = 0;
+      char* ParseEnd = nullptr;
+      double DefaultArgValue = std::strtod(Result.c_str(), &ParseEnd);
```
vgvassilev: Maybe we can take a step back here -- is that the way clang does this?

conrade-ctc: good point... claude pointed out to me that there is a more standard, generic fix... two possibilities... let me try that them, and repost... might close one of the two PRs actually, since they are related. Give me a sec...

conrade-ctc: Good call — reworked. It now does what `SemaCodeComplete`'s `GetDefaultValueString` does: read the parameter's `getDefaultArgRange()` source text via `Lexer::getSourceText` (including stripping the sometimes-present leading `=`, per the FIXME there), falling back to the AST pretty-print only when there's no usable range. That makes the output as-written by construction, so the float-precision FIXME and the `std::stod` normalization are gone entirely. Literal defaults now render exactly as written (`4.0` where the old code produced `4.`) — existing test expectations updated to match.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -5978,6 +5979,32 @@ std::string GetFunctionArgDefault(ConstFuncRef func, size_t param_index) {
     PI = (FD->getTemplatedDecl())->getNonObjectParameter(param_index);
 
   if (PI->hasDefaultArg()) {
+    // Render the default as written: read its source range, the way clang's
+    // own code completion formats default arguments (GetDefaultValueString
+    // in SemaCodeComplete). Pretty-printing the AST instead reproduces
+    // floating literals at representation precision ("3.1400000000000001"
```
vgvassilev: I believe that's a bug in clang -- it should print the value as written. Can we follow up there, and land that as a temporary workaround?

conrade-ctc: makes sense, i'll iterate on that and reproducing the bug upstream. i'll post a PR for that separately, and we can treat this as a work-around. I'll update this PR as well with the proper guards so that we can clean up later.

conrade-ctc: also, in the process of iterating on the upstream, a lighter version of this diff that uses a different pathway was found, so i'm verifying it and probably will update this PR to swap in that change instead.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -5986,20 +5991,11 @@ std::string GetFunctionArgDefault(ConstFuncRef func, size_t param_index) {
...
-      if (!Result.empty() && Result.back() == '.')
-        return INTEROP_RETURN(Result);
-      auto DefaultArgValue = std::stod(Result);
-      std::ostringstream oss;
-      oss << DefaultArgValue;
-      Result = oss.str();
-    }
+    ASTContext& Ctx = getASTContext();
+    PrintingPolicy Policy(Ctx.getLangOpts());
+    Policy.ConstantsAsWritten = true;
```
vgvassilev, line 6025: Oh, that’s nice. And probably kills our upstream story probably for good.

lib/CppInterOp/CppInterOp.cpp
conrade-ctc: Reworked: the patch now prints defaults with `printPretty`, the interpreter's `ASTContext`, and `ConstantsAsWritten = true`, instead of the hand-rolled source-text read. All existing tests pass unchanged. The clang-side precision bug is being fixed in llvm/llvm-project#218471; until then, defaults that `ConstantsAsWritten` cannot cover (cooked UDL floats, invalid-range literals) still print at maximum precision. A version-gated canary test expects the old output before clang 24 and the fixed output from clang 24 on, so a cherry-pick or a slipped landing shows up as a named failure.

lib/CppInterOp/CppInterOp.cpp
conrade-ctc: @vgvassilev, do you think the UDL hole that is listed above is worth a PR to clang as well? it's a separate gap that might be valuable to clean up.

vgvassilev: Oh, ok. Did not realize that’s a disk access. That’s pretty slow for the pch case and requires headers to be around..

conrade-ctc: hmmm, by UDL i mean user-defined-literal, not the disk one :) Just want to make sure we're on the same page... there is apparently a minor hole in the UDL float impl that mixes `90.0_deg` and `90._deg` for example... claude identified that as a follow up that is related and needs a separate treatment upstream.

vgvassilev (approved):
Lgtm!

PR #1070, opened 2026-07-16
Reset sticky diagnostic state after a rejected InstantiateTemplate

conrade-ctc:
# Description

A rejected `Cpp::InstantiateTemplate` (e.g. a type argument supplied for a non-type template parameter) emits its error diagnostic outside any incremental parse, and `DiagnosticsEngine`'s `ErrorOccurred` flag is sticky. The next `IncrementalParser::Parse` sees the stale flag after a perfectly good parse, reports "Parsing failed.", and soft-resets the engine — so exactly one subsequent `Cpp::Declare`/`Process` fails spuriously. Downstream (cppyy), one rejected template subscript makes the next `cppyy.include`/`cppyy.cppdef` report failure while actually loading the code, which can silently corrupt fail-soft feature probes.

Per review, the rejection is now contained by a `DiagnosticErrorTrap`-based RAII (`StickyDiagnosticResetRAII`): the trap detects errors introduced during the instantiation, and on scope exit the guard restores a clean state with the same `Diags.Reset(/*soft=*/true)` + client `clear()` cleanup that `IncrementalParser` performs on a real parse failure (and that `InstantiateFunctionDefinition` performs inline). The rejection's diagnostic text is still emitted to the client as before.

## Type of change

- [x] Bug fix

## Testing

New `ScopeReflection_RejectedInstantiationDoesNotPoisonNextParse`: a rejected instantiation followed by two `Cpp::Declare` calls, both of which must succeed. Fails before the fix (the first Declare reports failure), passes after.

## Checklist

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -5206,6 +5206,35 @@ std::string ObjToString(const char* TyRef, void* obj) {
   return INTEROP_RETURN(getInterp().toString(TyRef, obj));
 }
 
+namespace {
+/// Traps diagnostics emitted by Sema entry points invoked outside any
+/// incremental parse and, when errors were introduced, restores a clean
+/// diagnostic state on scope exit — the same cleanup IncrementalParser
+/// performs on a real parse failure. Without it, DiagnosticsEngine's sticky
+/// ErrorOccurred flag makes the next IncrementalParser::Parse report the
+/// stale flag as its own failure (and then soft-reset it), spuriously
+/// failing exactly one parse.
+class StickyDiagnosticResetRAII {
+  clang::DiagnosticsEngine* m_Diags;
+  clang::DiagnosticErrorTrap m_Trap;
+
+public:
+  StickyDiagnosticResetRAII(clang::DiagnosticsEngine& Diags)
+      : m_Diags(&Diags), m_Trap(Diags) {}
+  StickyDiagnosticResetRAII(const StickyDiagnosticResetRAII&) = delete;
+  StickyDiagnosticResetRAII&
+  operator=(const StickyDiagnosticResetRAII&) = delete;
+  StickyDiagnosticResetRAII(StickyDiagnosticResetRAII&&) = delete;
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -5303,7 +5303,20 @@ DeclRef InstantiateTemplate(compat::Interpreter& I, DeclRef tmpl,
...
-  return InstantiateTemplate(TmplD, TemplateArgs, S, instantiate_body);
+  // A rejected instantiation emits its error diagnostic outside any
+  // incremental parse, and DiagnosticsEngine's ErrorOccurred flag is sticky:
+  // the next IncrementalParser::Parse would report the stale flag as its own
+  // failure and then soft-reset it, poisoning exactly one parse. Restore a
+  // clean diagnostic state if this call introduced the error (same cleanup
+  // the incremental parser performs on a real parse failure).
+  clang::DiagnosticsEngine& Diags = S.getDiagnostics();
+  bool HadErrors = Diags.hasErrorOccurred();
+  Decl* Result = InstantiateTemplate(TmplD, TemplateArgs, S, instantiate_body);
```
vgvassilev: Can can insert a diagnostic trap object -- we have that elsewhere in the codebase.

conrade-ctc: indeed, thanks for pointing that out, update forth-coming!

conrade-ctc: Done — the manual snapshot is now a `DiagnosticErrorTrap`-based RAII (`StickyDiagnosticResetRAII`): the trap detects errors introduced during the instantiation, and on scope exit the guard restores a clean state with the same `Diags.Reset(/*soft=*/true)` + client `clear()` cleanup `IncrementalParser` performs on a real parse failure. I left `InstantiateFunctionDefinition`'s inline copy of this cleanup untouched — it could migrate to the same guard, but #849 is already touching that spot and I didn't want the two PRs to collide.

vgvassilev: Can you rebase on top of the new master?

conrade-ctc: Rebased on top of main.

PR #1072, opened 2026-07-23
IsAllocator redesigned and intercommunicating with GetAllocType

keremsahn:
IsAllocator now returns an AllocType value instead of bool, and it recognizes user added attrbiutes with specific keyword:
cppAlloc`AllocTypeEnum` 
For example
```cpp
[[clang::annotate("cppAllocMalloc")]] void* mergeFunc();
```
cppAllocMalloc attribute is added to mergeFunc with specific signature, and IsAllocator recognizes cppAllocMalloc  attribute as AllocType::Malloc. 

Also traverser now asks IsAllocator whether the callee is labeled with an attribute before recursively analyzing it
@Vipul-Cariappa @vgvassilev

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1798,7 +1798,7 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
       varMap[VD] = handleExpr(RHS);
       return true;
     }
-    if (BO->isCompoundAssignmentOp()) {
+    if (VD->getType()->isPointerType() && BO->isCompoundAssignmentOp()) {
       if (undoLog)
         undoLog->try_emplace(VD, varMap[VD]);
       varMap[VD] = AllocType::Unknown;
@@ -1870,6 +1870,20 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
     return AllocType::New;
   }
 
+  static bool isNullOpt(const clang::CXXConstructExpr* CCE) {
+    // Expected: std/cpp::optional constructor
+    const clang::CXXConstructorDecl* CCD = CCE->getConstructor();
+    // optional(nullopt_t) noexcept; -> one arg
+    if (CCD->getNumParams() != 1)
+      return false;
+    clang::QualType ParamTy =
+        CCD->getParamDecl(0)->getType().getNonReferenceType();
+    // FIXME: Copy constructor of optional is not handled
+    if (const auto* CRD = ParamTy->getAsCXXRecordDecl())
+      return CRD->getName() == "nullopt_t";
+    return false;
+  }
+
   std::optional<AllocType> handleExpr(const clang::Expr* expr) {
     const clang::Expr* finExpr = expr->IgnoreParenCasts();
     // Case: return new __type__
@@ -1888,10 +1902,20 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
     }
 
     // Case: malloc or another func call
-    if (const auto* CE = dyn_cast<CallExpr>(finExpr))
+    if (const auto* CE = dyn_cast<CallExpr>(finExpr)) {
+      // Case: std::optional<int*> ptr = new int; return *ptr;
+      if (const auto* COCE = dyn_cast<clang::CXXOperatorCallExpr>(CE)) {
+        if (COCE->getOperator() == OverloadedOperatorKind::OO_Star &&
+            !COCE->isInfixBinaryOp())
+          return handleExpr(COCE->getArg(0));
+      }
       return handleCall(CE);
+    }
 
     // Case: NULL, nullptr or (int*)(__integerLiteral__)
+    if (llvm::isa<clang::CXXNullPtrLiteralExpr>(finExpr) ||
+        llvm::isa<clang::GNUNullExpr>(finExpr))
+      return AllocType::Null;
     const clang::Expr* parExpr = expr->IgnoreParens();
     while (const auto* CE = dyn_cast<CastExpr>(parExpr)) {
       if (CE->getCastKind() == CK_NullToPointer)
@@ -1899,6 +1923,23 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
       parExpr = CE->getSubExpr();
     }
 
+    // Case: constructor cast
+    if (const auto* CCE = dyn_cast<clang::CXXConstructExpr>(finExpr)) {
+      // std::nullopt or cpp::nullopt
+      if (isNullOpt(CCE))
+        return AllocType::Null;
+      const clang::Expr* stripped = expr->IgnoreParens();
+      // ExprWithCleanups -> ImplicitCastExpr -> CXXConstructExpr pattern
+      if (const auto* EWC = dyn_cast<clang::ExprWithCleanups>(stripped))
+        stripped = EWC->getSubExpr();
+      // We do not want to analyze explicit conversions
+      const auto* ICE = dyn_cast<clang::ImplicitCastExpr>(stripped);
+      bool isImplicitConv =
+          ICE && ICE->getCastKind() == CK_ConstructorConversion;
+      // ImplicitCastExpr -> CXXConstructExpr pattern
+      if (isImplicitConv)
+        return handleExpr(CCE->getArg(0));
+    }
     return AllocType::None;
   }
 };
@@ -1908,8 +1949,8 @@ static std::optional<AllocType>
 AnalyzeAllocType(const clang::FunctionDecl* Fn,
                  std::unordered_map<const clang::FunctionDecl*,
                                     std::optional<AllocType>>& visitedFuncs) {
-  const clang::QualType QT = Fn->getReturnType();
-  if (!QT->isPointerType())
+  QualType QT = Fn->getReturnType();
+  if (!QT->isPointerType() && !QT->isRecordType())
     return AllocType::None;
   const Stmt* fnBody = Fn->getBody();
   if (!fnBody)
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1650,6 +1650,9 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
   // Result var keeps the combination all possible values of previous return
   // statements
   std::optional<AllocType> result;
+  // A map every branch writes the previous value of overwriten variables
+  std::unordered_map<const clang::VarDecl*, std::optional<AllocType>>* undoLog =
+      nullptr;
 
   AllocationTraverser(std::unordered_map<const clang::FunctionDecl*,
                                          std::optional<AllocType>>& cache)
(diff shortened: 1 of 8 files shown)


keremsahn:
> @Vipul-Cariappa @vgvassilev

I found an important issue while developing recursion support in deallocation. In this PR attribute injection is done in static function not in the API itself, but in recursive cases function in cache can be evaluated wrong even though the main function is always evaluated true, and the false evaluation is injected to the recursively analyzed functions. Therefore attribute injection mechanism should be moved to API function, which is not called in recursive cases

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1700,7 +1721,7 @@ bool IsFunctionProtoType(ConstTypeRef TyRef) {
 }
 
 static std::optional<AllocType>
-AnalyzeAllocType(const clang::FunctionDecl* Fn,
+AnalyzeAllocType(clang::FunctionDecl* Fn,
```
Vipul-Cariappa: Why is this change required?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1648,31 +1648,52 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
...
+      if (attrName == "cppAllocNone")
+        return INTEROP_RETURN(AllocType::None);
+      if (attrName == "cppAllocNew")
+        return INTEROP_RETURN(AllocType::New);
+      if (attrName == "cppAllocNewArr")
+        return INTEROP_RETURN(AllocType::NewArr);
+      if (attrName == "cppAllocMalloc")
+        return INTEROP_RETURN(AllocType::Malloc);
+      if (attrName == "cppAllocUnknown")
+        return INTEROP_RETURN(AllocType::Unknown);
```
Vipul-Cariappa: Not required. Why should we support "unknown" attribute?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1648,31 +1648,52 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
...
+        return INTEROP_RETURN(AllocType::Malloc);
     }
 
     if (FD->hasAttr<CFReturnsRetainedAttr>() ||
         FD->hasAttr<NSReturnsRetainedAttr>() ||
         FD->hasAttr<OSReturnsRetainedAttr>())
-      return INTEROP_RETURN(true);
+      return INTEROP_RETURN(AllocType::Malloc);
+    for (const auto* attr : FD->specific_attrs<clang::AnnotateAttr>()) {
+      llvm::StringRef attrName = attr->getAnnotation();
```
Vipul-Cariappa: Can you check if this same code also works with APINotes? If not, what additional changes are required?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1648,31 +1648,52 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
   return INTEROP_RETURN(nullptr);
 }
 
-bool IsAllocator(ConstFuncRef Fn) {
+std::optional<AllocType> IsAllocator(ConstFuncRef Fn) {
```
Vipul-Cariappa: Why introduce `std::optional`? Can we represent no attribute available with `Unknown`?
You can condition on `Unknown` in `AnalyzeAllocType`.

Vipul-Cariappa (commented):
Initial quick review.
This looks fine to me. But I have a few questions. I have not looked at the tests yet, nor at the record support part of the changes.
Is it possible for this PR to be independent of #1083? This PR was opened before that.

lib/CppInterOp/CppInterOp.cpp
keremsahn: You are right, I put it there to be consistent with enum

lib/CppInterOp/CppInterOp.cpp
keremsahn: No, I need another loop like
```cpp
for (const auto* attr : FD->specific_attrs<clang::SwiftAttrAttr>())
```

lib/CppInterOp/CppInterOp.cpp
keremsahn: You are also right about that, I put optional to distinguish no-attr functions with None-Unknown attred functions, for recursive base case in analyzer, but it makes sense to remove cppAllocUnknown and we can use this value for no-attr situation

lib/CppInterOp/CppInterOp.cpp
keremsahn: We also do not need Null, I guess? It is only useful in recursion, to not effect other returns

lib/CppInterOp/CppInterOp.cpp
keremsahn: First version of this PR was injecting attributes after analyzing, and I found serious problems with that and removed the feature, but forgot to change const and const_cast. I reversed them git diff is clear now.

Also it is not dependent to any other PR now, it would be better to merge this PR first

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1648,31 +1648,56 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
...
+        return INTEROP_RETURN(AllocType::OperatorNew);
+      if (attrName == "cppAllocOperatorNewArr")
+        return INTEROP_RETURN(AllocType::OperatorNewArr);
+    }
   }
 
-  return INTEROP_RETURN(false);
+  // Nullopt is returned because when analyzer calls this API it should know
+  // whether an attribute injected before or FD has a meaningful attribute
+  return INTEROP_RETURN(AllocType::Unknown);
```
Vipul-Cariappa: Is this comment relevant?

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -856,24 +856,55 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_IsAllocator) {
...
-  EXPECT_FALSE(Cpp::IsAllocator(Decls[3]));
-  // Builtin check
-  code = R"(
-  //There is nothing lstdlib.h is included at args
+
+    void __attribute__((ownership_takes(malloc, 1))) Deallocator(void* p);
+    void* __attribute__((cf_returns_retained)) CFAllocFunc();
+
+    void __attribute__((annotate("cppAllocNone"))) NoneFunc();
+    void __attribute__((annotate("cppAllocNew"))) NewFunc();
```
Vipul-Cariappa: A void function, that returns `New`!!

Vipul-Cariappa (commented):
Can we have one test case for a templated function and one for methods?

One more test where the annotation is not present in the original function. For example:
```c++
int* foo() { return new int; }
__attribute__((annotate("cppAllocNew"))) int* foo();
```

unittests/CppInterOp/FunctionReflectionTest.cpp
keremsahn: 😅 Since i did not implement or call any of them and the functions are just for attribute holding, I did not care but I can change it if you think it is required

lib/CppInterOp/CppInterOp.cpp
keremsahn: No, I forgot to delete it after removing attribute injection

keremsahn:
> Can we have one test case for a templated function and one for methods?
> 
> One more test where the annotation is not present in the original function. For example:
> 
> ```c++
> int* foo() { return new int; }
> __attribute__((annotate("cppAllocNew"))) int* foo();
> ```

I added one test for a method, but testing redeclaration is not responsibility of this function because this function is just there to recognize whether the given decl has the target attributes, but redeclaration is more about clang's merging mechanism and which decl you take and give to test, so lookup. These tests are already written in CppJIT and they pass because CppJIT's lookup mechanism takes the last decl, in despite of CppInterOp's GetNamed that takes the canonical decl which will not include the attribute if you redeclare the function after implementing it.

And for the templates, I am aware they do not enter in both analyzer and IsAllocator because they are FunctionTemplateDecl, but CppJIT gives the instantiated version so both works fine except caching when you call function with different T parameters. We can discuss it but I do not add it for now because I know it will not pass

unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -1749,14 +1837,27 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, FunctionReflection_GetAllocType) {
...
-  Interp->declare(code);
+#endif
 
+  Interp->declare(code);
 #define TESTAC(N, EXP)                                                         \
-  EXPECT_EQ(                                                                   \
-      Cpp::GetAllocType(Cpp::ConstFuncRef { Cpp::GetNamed("func" #N).data }),  \
-      Cpp::AllocType::EXP)
+  EXPECT_EQ(Cpp::GetAllocType(Cpp::FuncRef { Cpp::GetNamed("func" #N).data }), \
+            Cpp::AllocType::EXP)
```
Vipul-Cariappa: Can this change be reverted? I believe the function has not changed to non-const.

Vipul-Cariappa (commented):
Could you please rebase?

This PR almost looks good to me. The templated functions not being supported is fine for now. But you should open an issue about it, and we should support it. Maybe in a different PR.

Regarding redeclarations: You may want `FunctionDecl->getCanonicalDecl()` or `FunctionDecl->getMostRecentDecl()`. CppJIT and CppInterOp only use canonical decls (i.e. the first declaration). If you cannot fix this in this PR, open an issue. We can merge it in the current state, then work on redeclaration.
The redeclaration problem is also applicable to #1099.

keremsahn:
> Could you please rebase?
> 
> This PR almost looks good to me. The templated functions not being supported is fine for now. But you should open an issue about it, and we should support it. Maybe in a different PR.
> 
> Regarding redeclarations: You may want `FunctionDecl->getCanonicalDecl()` or `FunctionDecl->getMostRecentDecl()`. CppJIT and CppInterOp only use canonical decls (i.e. the first declaration). If you cannot fix this in this PR, open an issue. We can merge it in the current state, then work on redeclaration. The redeclaration problem is also applicable to #1099.

Templated functions are supported:
```cpp
  const auto* D = UnwrapUsingShadowToFunction(unwrap<clang::Decl>(Fn));
  if (const auto* FTD = dyn_cast<FunctionTemplateDecl>(D))
    D = FTD->getTemplatedDecl();
```
recall: #1104, you commented to state where it is used, and one example is here.

Supporting templates is not really much of thing. I can just do a refactor for the analyzer adding a few lines and it will also be fine. 

And for redeclaration, I doubt CppJIT uses CanonicalDecls, it takes last decl each time as I remember, and that is why first including and then redeclaring was working. If the way CppJIT does lookup changed, I also need to change this PR.

https://github.com/compiler-research/cppjit/pull/35/changes
Check this PR, there is a test where redeclaration is after including, so if it was the case that CppJIT looks to the CanonicalDecls, this test would fail.

Another approach can be traversing decls, but  I thought this would be excessive

unittests/CppInterOp/FunctionReflectionTest.cpp
keremsahn: Yes, the reason I changed it first version of this PR had attribute injection so I removed consts, and forgot to change things when injection plan was abandoned. I will revert here

Vipul-Cariappa (commented):
LGTM!
Could you please open an issue for redeclarations.

Vipul-Cariappa (approved):

PR #1073, opened 2026-07-24
Bind weak globals the process already defines on the in-process JIT

conrade-ctc:
## Problem

Singletons *defined* in headers -- function-local statics in inline functions (Meyers) and C++17 inline static data members -- compile to weak/linkonce_odr global variables. When jitted code includes such a header while a loaded library already defines the symbol, the ORC JIT materializes its own copy: JITDylib definitions win over the process-symbol generator, which is only consulted for symbols the JITDylib lacks. The process and the interpreter then each hold a distinct instance of what the program means to be one singleton (diverging state, double destruction at teardown), where a dynamic linker would have unified them.

## Fix

`compat::bindProcessWeakGlobals`, run on each module before Execute (in-process, ELF): demote every mutable weak global-variable *definition* whose symbol `dlsym(RTLD_DEFAULT, ...)` already resolves to an external declaration, so the JIT binds the process copy.

- A dynamically initialized static's guard (`_ZGV<...>`) moves with its data, and only when the process exports both -- splitting them would re-run or skip initialization on the shared copy.
- Only mutable variables: duplicated constants are ODR-harmless (and folding keeps jitted code fast); functions may legitimately be re-jitted; thread_locals are excluded (the JIT's emulated-TLS storage regime is separate).
- Demoted globals are dropped from `llvm.used`/`llvm.compiler.used`.
- ELF + in-process only; Windows lacks exported weak symbols for this to bind against, and Mach-O needs its own validation (leading-underscore mangling, no comdats).

## Status: workaround

This is a stopgap for a deficiency in clang's in-process JIT (clang-repl / ORC in `clang::Interpreter`): the JIT keeps its own copy of weak definitions the surrounding process already exports. The proper fix belongs in `clang::Interpreter`, where it covers every embedder (in- and out-of-process) -- tracked upstream by llvm/llvm-project#211786. The pass and its two call sites are gated behind a single `CPPINTEROP_WORKAROUND_BIND_PROCESS_WEAK_GLOBALS` toggle, an `#error` fires at LLVM 25, when ORC is expected to unique globals, so the workaround gets removed then.

## Testing

`DynamicLibraryManagerTest.cpp`: TestSharedLib gains both singleton shapes plus AOT accessors. `DynamicLibraryManager_SharedWeakGlobals` loads the library, JITs the same definitions, and asserts address identity and cross-boundary mutation visibility for both flavors; red before the fix, green after. `SharedDynamicInitSingleton`, `SharedUsedWeakGlobal`, `JitOnlyWeakGlobalKept`, and `WeakConstantNotDemoted` cover the guard pairing, the `llvm.used` rewrite, and the two keep paths. Three `BindProcessWeakGlobals` tests check the bind decisions on a hand-built module. `check-cppinterop` passes; changed lines are clang-format and clang-tidy clean.

### DIFF
--- a/lib/CppInterOp/CppInterOpInterpreter.h
+++ b/lib/CppInterOp/CppInterOpInterpreter.h
@@ -368,7 +368,23 @@ public:
   }
 
   llvm::Error ParseAndExecute(llvm::StringRef Code, clang::Value* V = nullptr) {
-    return inner->ParseAndExecute(Code, V);
+    // Value-returning execution keeps clang's LastValue handling (private to
+    // clang::Interpreter), so delegate. The no-value path -- used by wrapper
+    // compilation -- is split so the module can be sanitized before Execute.
+    if (V)
+      return inner->ParseAndExecute(Code, V);
+    auto PTU = inner->Parse(Code);
+    if (!PTU)
+      return PTU.takeError();
+    if (PTU->TheModule) {
+#ifndef _WIN32
+      if (!outOfProcess)
+        compat::bindProcessWeakGlobals(*PTU->TheModule);
+#endif
+      if (llvm::Error Err = inner->Execute(*PTU))
+        return Err;
+    }
+    return llvm::Error::success();
   }
 
   llvm::Error Undo(unsigned N = 1) { return compat::Undo(*inner, N); }
@@ -462,6 +478,11 @@ public:
 
     if (PTU)
       *PTU = &*PTUOrErr;
+
+#ifndef _WIN32
+    if (PTUOrErr->TheModule && !outOfProcess)
+      compat::bindProcessWeakGlobals(*PTUOrErr->TheModule);
+#endif
 
     if (auto Err = Execute(*PTUOrErr)) {
       llvm::logAllUnhandledErrors(std::move(Err), llvm::errs(),
--- a/lib/CppInterOp/CppInterOpInterpreter.h
+++ b/lib/CppInterOp/CppInterOpInterpreter.h
@@ -377,7 +377,9 @@ public:
     if (!PTU)
       return PTU.takeError();
     if (PTU->TheModule) {
-#ifndef _WIN32
+      // WORKAROUND: see bindProcessWeakGlobals in Compatibility.h -- remove
+      // with the pass once the clang JIT fix lands.
+#if !defined(_WIN32) && CPPINTEROP_WORKAROUND_BIND_PROCESS_WEAK_GLOBALS
       if (!outOfProcess)
         compat::bindProcessWeakGlobals(*PTU->TheModule);
 #endif
@@ -479,7 +481,9 @@ public:
     if (PTU)
       *PTU = &*PTUOrErr;
 
-#ifndef _WIN32
+      // WORKAROUND: see bindProcessWeakGlobals in Compatibility.h -- remove
+      // with the pass once the clang JIT fix lands.
+#if !defined(_WIN32) && CPPINTEROP_WORKAROUND_BIND_PROCESS_WEAK_GLOBALS
     if (PTUOrErr->TheModule && !outOfProcess)
       compat::bindProcessWeakGlobals(*PTUOrErr->TheModule);
 #endif
--- a/lib/CppInterOp/Compatibility.h
+++ b/lib/CppInterOp/Compatibility.h
@@ -30,8 +30,16 @@
 #include "clang/Sema/Sema.h"
 
 #include "llvm/ADT/IntrusiveRefCntPtr.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/DerivedTypes.h"
+#include "llvm/IR/GlobalValue.h"
+#include "llvm/IR/GlobalVariable.h"
+#include "llvm/IR/Module.h"
+#include "llvm/IR/Value.h"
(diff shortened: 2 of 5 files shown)


vgvassilev:
Isn’t that a pure jit problem? I suspect we need to move to llvm so that we can have a discussion with the experts what to do here.

conrade-ctc:
Yes, I figured that there was an upstream fix needed for this, we just added this patch on our forks because otherwise we need to enforce a particular way of doing singleton, which we don't want to impose. I will also point Claude at clang to investigate there.

vgvassilev:
I suspect that would require a deeper design discussions with the jit folks.

As usual, the current set of workaround-level changes should be fine provided we have a upstream tracking issue/pr; the changes well scoped within 1-2 routines and we have an ifdef to remind us when to get rid of them.

conrade-ctc:
> I suspect that would require a deeper design discussions with the jit folks.
> 
> As usual, the current set of workaround-level changes should be fine provided we have a upstream tracking issue/pr; the changes well scoped within 1-2 routines and we have an ifdef to remind us when to get rid of them.

I have some tests that repro this for clang, and also got claude to provide a possible fix... if it's not insane looking, I can open it as a PR for you to put your eyes on and decide if it's an approach worth iterating on if that makes sense. I'll clean up this PR to get it to be bit be a bit more contained and make sure it's more clearly a work-around that we're looking at fixing upstream.

conrade-ctc:
> > I suspect that would require a deeper design discussions with the jit folks.
> > As usual, the current set of workaround-level changes should be fine provided we have a upstream tracking issue/pr; the changes well scoped within 1-2 routines and we have an ifdef to remind us when to get rid of them.
> 
> I have some tests that repro this for clang, and also got claude to provide a possible fix... if it's not insane looking, I can open it as a PR for you to put your eyes on and decide if it's an approach worth iterating on if that makes sense. I'll clean up this PR to get it to be bit be a bit more contained and make sure it's more clearly a work-around that we're looking at fixing upstream.

Ah, claude's solution for this upstream is exactly the same... that is, basically does the second pass to `bindProcessWeakGlobals` when adding a module (and it's current ELF only...). If that doesn't seem the right place to start to you, i can just file an issue with the repo details. If might be worth iterating on, let me know and I can open the PR with that candidate (it's already ready to post, test is red then green with the fix, larger clang suite is all green as well).

conrade-ctc:
@vgvassilev, does this one seem reasonble to merge soon? I'm going through the current list of open PRs, and wondering which might be closer to merging in before I send up more PRs :) I've got another handful queued up on my side that I haven't sent, and I don't want to clog your queue with too much!

lhames:
This seems reasonable as a temporary workaround.

The real fix for this is in the works, but probably a few months away yet: GOT entry binding needs to move from the JIT linker to the ORC runtime. At runtime we'll look up addresses emulating the same order as the dynamic loader, so we'll bind to external code first*.

\* Unless you're federating with the system dynamic loader, in which case you'll bind against JIT'd code iff that's the correct resolution order.

vgvassilev:
Thank you, Lang! @conrade-ctc can we write a tight set of tests that hit the uncovered lines?

conrade-ctc:
The new tests call `bindProcessWeakGlobals` directly on a hand-built module and close the two real gaps: `679` (a module guard the process does not export) and `712` (a used list holding no demoted entry), plus the used-list rebuild as a contrast case.

The other four flagged lines are counter placement, not untested code — every statement in the patch executes under the tests. gcc emits no counter on a lambda's closing brace (`658`) or on the leading line of a multi-line statement (`675`, `716-717`); their counters sit on the adjacent lines, which are covered. The macOS job cannot fill them in either: the function is ELF-only and returns at the triple check. They will stay flagged unless the matrix gains a clang-instrumented ELF coverage job.

conrade-ctc:
@vgvassilev, tests added, CI is green/clean again. Thanks.

lib/CppInterOp/CppInterOpInterpreter.h
```diff
@@ -368,7 +368,25 @@ class Interpreter {
   }
 
   llvm::Error ParseAndExecute(llvm::StringRef Code, clang::Value* V = nullptr) {
-    return inner->ParseAndExecute(Code, V);
+    // Value-returning execution keeps clang's LastValue handling (private to
+    // clang::Interpreter), so delegate. The no-value path -- used by wrapper
+    // compilation -- is split so the module can be sanitized before Execute.
+    if (V)
+      return inner->ParseAndExecute(Code, V);
```
lhames: Should this get the same `compat::bindProcessWeakGlobals` treatment as the PTU module below? (At a glance it's not clear to me why the modules would be treated differently, but there may be a good reason)

lhames:
@conrade-ctc Can you add a TODO to remove this once ORC supports global uniquing (ETA LLVM 25)?

Otherwise this looks good to me.

vgvassilev:
> @conrade-ctc Can you add a TODO to remove this once ORC supports global uniquing (ETA LLVM 25)?
> 
> Otherwise this looks good to me.

Thank you, Lang! Yes, our usual `#ifdef LLVM_VERSION_MAJOR` and an `#error` should be a good way to implement a todo.

conrade-ctc:
Done: a TODO with the LLVM 25 ETA, and an `#error` on `LLVM_VERSION_MAJOR >= 25` right above the toggle, so the build stops us from carrying it past that point.

vgvassilev (approved):
Lgtm!

PR #1074, opened 2026-07-28
Accept named constant entities as non-type template arguments

conrade-ctc:
Instantiating a template through cppyy's string subscript with the *name* of a constexpr variable or enum constant fails on the CppInterOp-based stack:

```python
cppyy.cppdef("template <int N> class C {}; constexpr int IntArg = 42;")
cppyy.gbl.C["IntArg"]   # error: template argument for non-type template
                        # parameter must be an expression
```

cling-based cppyy accepted this (the argument was pasted textually, so name lookup made it an expression). On this stack the name reaches `InstantiateTemplate` as a `TemplateArgInfo` whose `m_Type` is the *entity's declared type* and whose `m_IntegralValue` is a non-numeric string — which the current code feeds straight to `llvm::APSInt` (an assert/UB on non-digit input).

This PR:
- Factors the `TemplateArgInfo` → `TemplateArgument` conversion shared by `InstantiateTemplate` and `BestOverloadFunctionMatch` into one helper.
- Keeps numeric `m_IntegralValue` on the existing `APSInt` path.
- Treats a non-numeric `m_IntegralValue` as the (possibly qualified) name of a constant entity: it is resolved (a `GetNamed` walk that, unlike `GetScopeFromCompleteName`, also finds variables and enum constants) and wrapped in a `DeclRefExpr`, so Sema converts it exactly like a written argument — constant evaluation for integral parameters, array-to-pointer decay for `const char*` parameters, prvalue handling for enum constants.
- Treats a non-numeric `m_IntegralValue` naming a class/alias template as a template-template argument (`TemplateName`).
- Fails cleanly (null result) for unknown names or non-constant entities, with Sema's proper diagnostics.

Companion PR: compiler-research/cppyy-backend#219 makes `AppendTypesSlow` produce these name-valued `TemplateArgInfo`s (today it resolves such names to their *type* via `Cpp::GetType`, which is how the bad argument arises). Related discussion on the string-argument surface: https://github.com/compiler-research/cppyy-backend/pull/137#discussion_r2079357491

Tests: new `ScopeReflection_InstantiateTemplateNamedNTTPArg` covers a constexpr variable, a namespace-qualified name, an enum constant, char-array decay for a `const char*` parameter, and unknown-name failure; `ScopeReflection_InstantiateTemplateTemplateArg` covers plain/qualified/alias template-template arguments and null-info failure. `check-cppinterop` green.

Out of scope (intentionally): arbitrary constant expressions in strings (`"IntArg + 1"`, `"&obj"`, `"'x'"`) — only (possibly qualified) names of constant entities, templates, and integer literals are resolved.

Adds `Cpp::SupportsNamedTemplateArguments()` so callers can probe for this feature.


conrade-ctc:
A follow-up report on the same stack showed template-template arguments were still broken: a string naming a class template raised "Cannot find Templated Arg" (a template has no QualType, so the old type-only path had nothing to carry). The new commit extends the named-entity path: a non-numeric value naming a class/alias template now becomes a Template-kind `TemplateArgument` (`TemplateName`), converted by Sema like a written argument.

Probing the argument matrix around this also surfaced two hardening fixes, included in the commit:
- `IsEnumType` dereferenced a null `QualType` (reachable from cppyy when a `TemplateProxy` object is used as a template argument — a segfault in practice); it now returns false.
- A `TemplateArgInfo` with null `m_Type` and no value now fails cleanly instead of building a `TemplateArgument` from a null type.

New unit tests: `ScopeReflection_InstantiateTemplateTemplateArg` (plain, namespace-qualified, and alias templates; unknown-name and null-info failures) plus negative-literal coverage on the NTTP test and a null check in `EnumReflection_IsEnumType`.

vgvassilev (approved):
LGTM!

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2449,6 +2450,57 @@ bool GetClassTemplatedMethods(const std::string& name, ConstDeclRef parent,
...
+
+  llvm::StringRef Value(Info.m_IntegralValue);
+  llvm::StringRef Digits = Value;
+  (void)(Digits.consume_front("-") || Digits.consume_front("+"));
+  if (!ArgTy.isNull() && !Digits.empty() &&
+      Digits.find_first_not_of("0123456789") == llvm::StringRef::npos) {
+    auto Res = llvm::APSInt(Value);
+    Res = Res.extOrTrunc(S.getASTContext().getIntWidth(ArgTy));
+    return TemplateArgument(S.getASTContext(), Res, ArgTy);
+  }
```
vgvassilev, line 2613: On a second thought -- can we check if clang doesn't already have such conversion logic -- iirc there was something like that because we need to convert literals in the language.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2449,6 +2450,57 @@ bool GetClassTemplatedMethods(const std::string& name, ConstDeclRef parent,
...
+
+  Decl* Named = GetNamedFromCompleteName(Value.str());
```
vgvassilev, line 2616: I believe we do not need to do parsing here. We should have the access to the declaration through `ArgTy`.

lib/CppInterOp/CppInterOp.cpp
conrade-ctc: It does, and we already rely on it: every argument built here goes through `CheckTemplateArgumentList`, which applies the converted-constant-expression rules (`CheckTemplateArgument` -> `BuildConvertedConstantExpression`). The helper only builds the reference. It now uses `Sema::BuildDeclarationNameExpr` for that, so clang also picks the value kind and type - the enum-constant special case is gone. The numeric branch keeps the pre-existing `extOrTrunc`: the API accepts out-of-range literals that a strict conversion would reject, and Sema still re-checks the truncated value.

lib/CppInterOp/CppInterOp.cpp
conrade-ctc: `ArgTy` is the template parameter's type (`int` in the tests), so it cannot identify the entity - many declarations share one type, and a `QualType` never reaches a `ValueDecl`. For template-template arguments it is null by design. The name is the only handle the API carries, so I think the lookup should stay. The helper mirrors `GetScopeFromCompleteName` but also finds non-scope decls.

PR #1077, opened 2026-07-29
Resolve glibc libc_nonshared.a symbols for the in-process JIT

conrade-ctc:
The in-process JIT cannot resolve `at_quick_exit`. glibc keeps it (with `atexit`, `pthread_atfork`, and `__stack_chk_fail_local`) in `libc_nonshared.a`, so every executable and DSO links its own private copy. dlsym never sees these, and the ORC process-symbol generator fails the lookup:

```
JIT session error: Symbols not found: [ at_quick_exit ]
```

Repro: `Cpp::Process("extern \"C\" int at_quick_exit(void (*)(void)); int r = at_quick_exit([]{});")` — any header that calls `std::at_quick_exit` inline hits this.

The fix (in-process, glibc-only, installed in `Interpreter::create`):
- `__stack_chk_fail_local` is plain code, so a definition generator resolves it to this library's own copy.
- `at_quick_exit` and `pthread_atfork` register callbacks, and a JIT'd callback must never end up inside host glibc: glibc would call it after the JIT is gone. A small IR module in the main JITDylib defines both (the same pattern LLJIT uses for `atexit`). The shims store the callbacks in a host-side registry, tagged by owning interpreter.
- Quick-exit callbacks run exactly once: on a real `quick_exit`, or at interpreter teardown, whichever comes first.
- Atfork callbacks fire through fixed host-side hooks while their interpreter is alive and are dropped at teardown.
- `atexit` needs no entry: LLJIT's platform support already defines a JIT-aware `atexit` and runs its handlers at `deinitialize()`.

Tests cover symbol resolution, the run-exactly-once rule, live fork dispatch, and two death tests for `quick_exit`/`fork` after teardown (previously SIGSEGV in freed JIT memory).


lhames:
If I'm reading it right this is unsafe: It's pointing JIT'd code to register with the host `atexit`, but by the time that callback happens the JIT'd code may already have been torn down -- leading to a use-after-free.

You can already inject C code: I'd say that you should just build a small glibc_nonshared implementation that you load into your JIT as a first step. As long as that code doesn't depend on `atexit`, etc. itself, you can then arrange to run the JIT-registered atexit callbacks before you tear the JIT down.

conrade-ctc:
You're right — reproduced: JIT'd code registers `at_quick_exit(&jit_fn)`, the interpreter is destroyed, then `quick_exit(0)` faults with `$pc = jit_fn` in an unmapped page. `pthread_atfork` has the same hazard (fork after teardown faults in the freed prepare handler). `atexit` was already safe: the generic platform's per-JITDylib `atexit` shadows the fallback and runs its handlers at `deinitialize()`.

I am iterating on a fix along the lines you suggest and will update the PR.

Great catch, thanks!

conrade-ctc:
Updated per the review, @lhames for viz/re-review

The absolute-symbol table now only carries `__stack_chk_fail_local`, which is plain code and safe to share. `at_quick_exit` and `pthread_atfork` no longer hand JIT'd function pointers to glibc. A small IR module in the main JITDylib defines both, the same way the LLJIT platform defines `atexit`. The shims store the callbacks in a host-side registry, tagged with the interpreter that owns them.

Quick-exit callbacks run exactly once: on a real `quick_exit` if one happens, otherwise when the interpreter is destroyed — while the JIT can still run them. Atfork callbacks fire through fixed host-side hooks while their interpreter is alive, and are dropped when it goes away. `atexit` needs nothing: the per-JITDylib platform version already handles it.

Tests: two death tests cover the post-teardown paths (`quick_exit` or `fork` after `DeleteInterpreter` used to fault in freed JIT memory; both exit cleanly now), plus tests that callbacks run exactly once before teardown and that live fork dispatch works.

lib/CppInterOp/Compatibility.h
```diff
@@ -31,16 +31,44 @@
...
+#include <cstdlib>
 #include <cstring>
 #include <memory>
 #include <string>
 
+#ifndef _WIN32
+#include <pthread.h>
+#endif
+
+#ifdef __GLIBC__
```
vgvassilev: Maybe we can move that into a separate header CompatibilityGLIBC.h?

conrade-ctc: Moved: the glibc block now lives in `lib/CppInterOp/CompatibilityGLIBC.h`, so `Compatibility.h` is untouched by this PR. `CppInterOpInterpreter.h` and the test include the new header directly.

vgvassilev (approved):
Lgtm!

Issue #1078, opened 2026-07-30
GetAllCppNames: no names for a translation unit, and an intermittent crash in decls_begin()


conrade-ctc:
## Summary

`Cpp::GetAllCppNames(Cpp::GetGlobalScope(), names)` does not enumerate a translation unit. On a TU holding roughly 15,000 declarations it either returns an empty set or crashes the process inside `clang::DeclContext::decls_begin()`. Same call, same interpreter state: one run returned empty, the next faulted — so this looks like undefined behaviour rather than a missing null check.

Separately, and from code reading only, the `TagDecl` branch of the same function advances its iterator before the loop's end check.

## Environment

- CppInterOp `main` (321657d). Both functions quoted below are identical at that commit to the build I hit this on.
- clang-repl backend, in-process, LLVM 21.1, Linux x86-64.
- Driven through cppyy (visible in the stack frames), after the interpreter has included a large header closure.

## 1. Translation unit yields nothing, or crashes

```python
# after the interpreter has parsed a large header closure
names = cppyy.gbl.std.set['std::string']()
cppyy.gbl.Cpp.GetAllCppNames(cppyy.gbl.Cpp.GetGlobalScope(), names)
print(names.size())      # 0, for a TU with ~15k declarations
```

Re-running the identical script, the same call faults instead:

```
#4  clang::DeclContext::getParentASTContext() const                  DeclBase.cpp
#5  clang::DeclContext::LoadLexicalDeclsFromExternalStorage() const
#6  clang::DeclContext::decls_begin() const
#7  Cpp::GetAllCppNames(Cpp::ConstDeclRef, std::set<std::string>&)   lib/CppInterOp/CppInterOp.cpp
#10 Cpp::JitCall::Invoke(void*, Cpp::JitCall::ArgList, void*) const  include/CppInterOp/CppInterOpTypes.h
#11 WrapperCall(Cpp::FuncRef, unsigned long, void*, void*, void*)    clingwrapper/src/clingwrapper.cxx:1099
#12 CPyCppyy::VoidExecutor::Execute(...)                             src/Executors.cxx:435
```

The `DeclRef` is non-null, and `GetGlobalScope()` itself returned successfully.

One candidate cause: `GetGlobalScope` does not return the translation unit declaration — it returns a redeclaration-chain lookup from it (`lib/CppInterOp/CppInterOp.cpp:995`):

```cpp
DeclRef GetGlobalScope() {
  return INTEROP_RETURN(
[...]

PR #1079, opened 2026-07-31
Do not skip past decls_end() for an empty tag in GetAllCppNames

conrade-ctc:
The increment that skips a tag's injected class name ran before the loop's end check, so an enum with no enumerators advanced a past-the-end iterator and crashed the process.

```cpp
decl = DC->decls_begin();
decl++;                       // before the end check below
...
for (/* decl set above */; decl != DC->decls_end(); decl++)
```

The existing test's tag cases are all `CXXRecordDecl`s, which always contain the injected class name, so they never reach an empty context.

Part of compiler-research/CppInterOp#1078.


vgvassilev (approved):
LGTM!

PR #1080, opened 2026-08-02
[ci] Drop the obsolete Box.h staging shim in the ROOT job

aaronj0:
The ROOT job sed-injects CppInterOp/Box.h into ROOT's cppinterop_src_files header-staging list, as ROOT did not yet stage the (then new) Box.h header.

Since that was addressed in the last upgrade and it is now automatically staged, drop the sed that was previously staging manually. This should fix the current failing CI failure as Box.h is no longer staged twice: 

```
[83](https://github.com/compiler-research/CppInterOp/actions/runs/30551262339/job/90900286147?pr=1076#step:12:184)
-- Enabled support for:  builtin_cling runtime_cxxmodules shared testsupport thisroot_scripts
-- Configuring done (5.0s)
CMake Error:
  Running

   '/usr/local/bin/ninja' '-C' '/home/runner/work/CppInterOp/CppInterOp/root-build' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:1625: etc/cppinterop/CppInterOp/Box.h is defined as an output multiple times


  ninja: error: etc/cppinterop/CppInterOp/Box.h is defined as an
  output multiple times
```

### DIFF
--- a/.github/workflows/root.yml
+++ b/.github/workflows/root.yml
@@ -159,19 +159,6 @@ jobs:
         sed -i '/CPPINTEROP_ENABLE_TESTING.*Enable gtest in CppInterOp/s/ FORCE)/)/' \
           root/interpreter/CMakeLists.txt
 
-        # ROOT's interpreter/CMakeLists.txt hardcodes the list of headers
-        # it stages into ${CMAKE_BINARY_DIR}/etc/cppinterop/CppInterOp/.
-        # That list (search "set(cppinterop_src_files") names only the
-        # original three public headers and three generated .inc files.
-        # CppInterOp.h now also #includes <CppInterOp/Box.h>; without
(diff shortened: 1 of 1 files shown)


aaronj0:
@vgvassilev We should get this in before #1067 and once the token is updated in cling and ROOT syncs, the cling jobs should pass again.

vgvassilev (approved):
Lgtm!

vgvassilev:
> @vgvassilev We should get this in before #1067 and once the token is updated in cling and ROOT syncs, the cling jobs should pass again.

Cling is synchronized so these failures are probably something else.

aaronj0:
> Cling is synchronized so these failures are probably something else.

The jobs are currently fetching cling-llvm20 which does not exist.. Pushing a fix for that now.

PR #1083, opened 2026-08-09
Record support for GetAllocType

keremsahn:
Built upon #1065 

Restrict the pointer-arithmetic Unknown-downgrade in VisitBinaryOperator to
  pointer-typed variables only, so unrelated scalar compound-assignments no
  longer poison tracked results.
```cpp
if (VD->getType()->isPointerType() && BO->isCompoundAssignmentOp()) {
      if (undoLog)
        undoLog->try_emplace(VD, varMap[VD]);
      varMap[VD] = AllocType::Unknown;
```

  Allow AnalyzeAllocType to also traverse record-typed (class/struct) return
  values, not just raw pointers, so wrapper types like std::optional<T*> get
  analyzed instead of short-circuiting to None.

  In handleExpr:
  - Unwrap unary operator* calls (e.g.  on an optional) to
    recurse into the wrapped value instead of falling back to a call analysis.
```cpp
if (const auto* CE = dyn_cast<CallExpr>(finExpr)) {
      // Case: std::optional<int*> ptr = new int; return *ptr;
      if (const auto* COCE = dyn_cast<clang::CXXOperatorCallExpr>(CE)) {
        if (COCE->getOperator() == OverloadedOperatorKind::OO_Star &&
            !COCE->isInfixBinaryOp())
          return handleExpr(COCE->getArg(0));
      }
```

  - Detect nullptr/NULL literals via isNullPointerConstant API, which I were not aware of 
  ```cpp
      // Case: NULL, nullptr or (int*)(__integerLiteral__)
    if (finExpr->isNullPointerConstant(getASTContext(),
                                       Expr::NPC_ValueDependentIsNotNull) !=
        Expr::NPCK_NotNull)
      return AllocType::Null;
 ```
 
 
 
  - Recognize implicit converting-constructor calls (ExprWithCleanups ->
    ImplicitCastExpr(CK_ConstructorConversion) -> CXXConstructExpr) and
    recurse into the constructor's argument, so New/Malloc/etc. tags
    propagate through implicit wrapping (e.g. ).
    ```cpp
    // Case: constructor cast
    if (const auto* CCE = dyn_cast<clang::CXXConstructExpr>(finExpr)) {
      // std::nullopt or cpp::nullopt
      ```
  - Add isNullOpt() to recognize std::nullopt/cpp::nullopt construction as
    AllocType::Null.
   
   ```cpp
   static bool isNullOpt(const clang::CXXConstructExpr* CCE)
   ```

  Known limitation (documented via FIXME + test): optional's copy/move
  constructor is not a converting constructor, so copying an optional loses
  its tracked AllocType and reports None.

### DIFF
--- a/include/CppInterOp/CppInterOpTypes.h
+++ b/include/CppInterOp/CppInterOpTypes.h
@@ -411,11 +411,11 @@ enum class AllocType : unsigned char {
   New,
   NewArr,
   Malloc,
-  Unknown,
   CustomAlloc,
-  Null,
   OperatorNew,
-  OperatorNewArr
+  OperatorNewArr,
+  Null,
+  Unknown
 };
 
 enum class DeallocType : unsigned char {
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1730,8 +1730,8 @@ struct AllocationTraverser : RecursiveASTVisitor<AllocationTraverser> {
     return RecursiveASTVisitor::TraverseDecl(D);
   }
 
-  std::optional<AllocType> join(std::optional<AllocType> a,
-                                std::optional<AllocType> b) {
+  static std::optional<AllocType> join(std::optional<AllocType> a,
+                                       std::optional<AllocType> b) {
     if (a == AllocType::Null)
       return b;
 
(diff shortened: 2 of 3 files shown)


keremsahn:
@Vipul-Cariappa

PR #1085, opened 2026-08-20
Use the defining declaration's type for unbounded arrays

aaronj0:
`GetVariableType` returns the declared type of the passed decl. In the case opf `extern std::string arr[];`, we get back an incomplete array since it has no dimension (`len()` fails) and no element stride (every index read the first element). Prefer the defining declaration's complete type for arrays of unknown bound.

Fixes the test04_array_of_strings case of cppyy-test-stltypes.


aaronj0:
Draft, requires further investigation on why this fails on Cling

edit: fixed for cling, the test case emitted an extra decl, so I lookup by name now instead of using the same Decl index for clang-repl and cling

vgvassilev (approved):
LGTM!

PR #1086, opened 2026-08-20
Cycle through ambiguous lookups in `ExistsFunctionTemplate`

aaronj0:
Addresses the FIXME in `ExistsFunctionTemplate`, with an included test:

When the name lookup was ambiguous (an overload set), the existing code path blindly returned true, so any function with two or more non-template overloads was presented as a function template. In cppyy this wrapped plain overloaded functions in a TemplateProxy, whose error aggregation does not preserve the C++-exception priority of CPPOverload (e.g. a `std::runtime_error` thrown by an overload surfaced as a generic "Template method resolution failed" TypeError).

The fix was already documented in the FIXME comment: cycle through the found decls (following using declarations) and check whether any is actually a templated function.

Fixes the test49_overloads_with_runtime_errors case of cppyy-test-regression

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -2358,8 +2358,24 @@ bool ExistsFunctionTemplate(const std::string& name, ConstDeclRef parent) {
     return INTEROP_RETURN(IsTemplatedFunction(ND) ||
                           IsTemplateInstantiationOrSpecialization(ND));
 
-  // FIXME: Cycle through the Decls and check if there is a templated function
-  return INTEROP_RETURN(true);
+  // The name is ambiguous, i.e. an overload set: cycle through the found
+  // decls and check if any of them is a templated function. Blindly
+  // returning true here would present any function with two or more
+  // non-template overloads as a template.
+  auto& S = getSema();
+  auto& Ctx = getASTContext();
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -2221,8 +2221,24 @@ bool ExistsFunctionTemplate(const std::string& name, ConstDeclRef parent) {
...
+  auto& S = getSema();
+  auto& Ctx = getASTContext();
+  clang::LookupResult R(S, &Ctx.Idents.get(name), SourceLocation(),
+                        Sema::LookupOrdinaryName,
+                        RedeclarationKind::ForVisibleRedeclaration);
+  CppInternal::utils::Lookup::Named(&S, R, Within);
+  for (const NamedDecl* Found : R) {
+    const Decl* D = Found;
+    if (const auto* USD = llvm::dyn_cast<UsingShadowDecl>(Found))
+      D = USD->getTargetDecl();
```
vgvassilev, line 2374: Can we add a test case here?

aaronj0: Thanks for catching that, extended test to cover this

vgvassilev (approved):
LGTM!

PR #1087, opened 2026-08-20
Fix crash in `IsAbstract` for classes without a definition

aaronj0:
`CXXRecordDecl::isAbstract()` reads the definition data, which a forward-declared class does not have, so asking whether an incomplete class is abstract crashed. Answer false instead (not known to be abstract), reading through the definition when there is one.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -744,7 +744,8 @@ bool IsAbstract(ConstDeclRef DRef) {
   INTEROP_TRACE(DRef);
   const auto* D = unwrap<clang::Decl>(DRef);
   if (const auto* CXXRD = llvm::dyn_cast_or_null<clang::CXXRecordDecl>(D))
-    return INTEROP_RETURN(CXXRD->isAbstract());
+    return INTEROP_RETURN(CXXRD->hasDefinition() &&
+                          CXXRD->getDefinition()->isAbstract());
 
   return INTEROP_RETURN(false);
 }
--- a/unittests/CppInterOp/ScopeReflectionTest.cpp
+++ b/unittests/CppInterOp/ScopeReflectionTest.cpp
@@ -463,12 +463,17 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, ScopeReflection_IsAbstract) {
     int sum(int a, int b) {
       return a+b;
     }
+
+    class FwdDeclared;
   )";
 
   GetAllTopLevelDecls(code, Decls);
   EXPECT_FALSE(Cpp::IsAbstract(Decls[0]));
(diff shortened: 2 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -744,7 +744,8 @@ bool IsAbstract(ConstDeclRef DRef) {
   INTEROP_TRACE(DRef);
   const auto* D = unwrap<clang::Decl>(DRef);
   if (const auto* CXXRD = llvm::dyn_cast_or_null<clang::CXXRecordDecl>(D))
-    return INTEROP_RETURN(CXXRD->isAbstract());
+    return INTEROP_RETURN(CXXRD->hasDefinition() &&
```
aaronj0, line 747: I'm not sure if trying `GetOrForceDefinition` here before returning is a good idea.

Vipul-Cariappa (approved):
LGTM!

unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -463,12 +463,17 @@ TYPED_TEST(CPPINTEROP_TEST_MODE, ScopeReflection_IsAbstract) {
...
+    class FwdDeclared;
   )";
 
   GetAllTopLevelDecls(code, Decls);
   EXPECT_FALSE(Cpp::IsAbstract(Decls[0]));
   EXPECT_TRUE(Cpp::IsAbstract(Decls[1]));
   EXPECT_FALSE(Cpp::IsAbstract(Decls[2]));
+  // A class without a definition is not known to be abstract, and asking
+  // must not crash.
+  EXPECT_FALSE(Cpp::IsAbstract(Decls[3]));
```
vgvassilev, line 476: Yes, however, this change makes the answer ambiguous. That is a forward declared but abstract class will get false here.

aaronj0: So I assume the right fix is to force the definition and then check?

aaronj0: i.e Based on my knowledge, we cannot know if a class is abstract without its definition

lib/CppInterOp/CppInterOp.cpp
vgvassilev: It is better because it will return the correct result at least. However, we will need to drop the constness. This is something I believe I have a solution for in my local changes for error handling. In such cases we will return operation invalid if the definition is not complete and it is up to the client to decide if they want to deserialize or not. For now, calling for completion is the right thing to do. We also should add a test infrastructure exercising that.

aaronj0: Thanks, will open a follow up.

PR #1089, opened 2026-08-20
Guard `GetSizeOfType` against a null type

aaronj0:
QualType::getFromOpaquePtr(nullptr) produces a null QualType, and the subsequent getAs<TagType>() dereferences it. Return size 0 instead, so callers can detect that the size is unavailable

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -887,6 +887,8 @@ size_t GetEnumConstantValue(ConstDeclRef DRef) {
 
 size_t GetSizeOfType(ConstTypeRef TyRef) {
   INTEROP_TRACE(TyRef);
+  if (!TyRef)
+    return INTEROP_RETURN(0);
   QualType QT = QualType::getFromOpaquePtr(TyRef.data);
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -887,6 +887,8 @@ size_t GetEnumConstantValue(ConstDeclRef DRef) {
 
 size_t GetSizeOfType(ConstTypeRef TyRef) {
   INTEROP_TRACE(TyRef);
+  if (!TyRef)
+    return INTEROP_RETURN(0);
```
vgvassilev, line 891: This is another case of invalid operation. I think the check needs to be done on the client side. This change is acceptable as it only breaks non-standard c/c++: https://godbolt.org/z/99qfxzxG5

aaronj0: Makes sense, I agree we should not sacrifice correctedness since 0 does not signify error.  It is better to fix with proper runtime error-handling and not change the API behaviour temporarily

aaronj0:
Closing as the discussion points to an ambiguity that is better addressed with runtime error handling. A meantime fix is to address this on the caller side with a nullcheck.

PR #1090, opened 2026-08-20
Force definition completion in `IsAbstract` before returning false

aaronj0:
Follow-up to #1087 per review: abstractness is a property of the class definition, so answering false for a class whose definition is not loaded yet is ambiguous. Complete the definition through `GetOrForceDefinition` before answering; this drops the constness of the argument. A forward declaration with no definition anywhere still answers false.


vgvassilev (approved):
LGTM!

PR #1093, opened 2026-08-24
Search base classes in GetNamed for class scopes

aaronj0:
`Lookup::Named` performs a redeclaration-style lookup (`RedeclarationKind::ForVisibleRedeclaration`), for whihc `Sema::LookupQualifiedName` stops at the queried context since a redeclaration must live in the same scope. A lookup of an existing member name per [class.qual] also searches the base classes. `GetNamed` therefore missed any member inherited from a base, e.g. the `type` member typedef of `std::tuple_element`, implemented by older libstdc++ (gcc8/alma8) with recursive inheritance with `type` declared in a base of looked-up specialization. Newer libstdc++ declares it directly in the specialization, hiding the problem. `cling::LookupHelper` performs full C++ member lookup (qualifier is built by parsing the C++ name and building a `Sema::BuildCXXNestedNameSpecifier`) and was not affected.

Retry the lookup with RedeclarationKind::NotForRedeclaration when the scope is a class and the direct lookup found nothing, accepting only an unambiguous result. Fixes test20_tuple_element of cppyy-test-cpp11features on alma8.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1257,6 +1257,34 @@ DeclRef GetNamed(const std::string& name, ConstDeclRef parent /*= nullptr*/) {
   if (ND && ND != (clang::NamedDecl*)-1)
     return INTEROP_RETURN(ND->getCanonicalDecl());
 
+  // Redeclaration-style lookup stops at the class itself; [class.qual]
+  // member lookup also searches the bases. Retry accordingly.
+  // FIXME: GetNamed should only return decls that are named and reachable
+  // within the provided decl context — the same contract question as the
+  // LookupUnqualified FIXME above: expose distinct lookup routes so callers
+  // pick the semantics they want, instead of GetNamed approximating both.
+  if (!ND && Within) {
+    if (auto* RD = llvm::dyn_cast<clang::CXXRecordDecl>(Within)) {
+      auto* Def = llvm::dyn_cast_or_null<clang::CXXRecordDecl>(
+          unwrap<clang::Decl>(GetOrForceDefinition(DeclRef(RD))));
+      if (!Def)
+        return INTEROP_RETURN(nullptr);
+      auto& S = getSema();
+      clang::DeclarationName DName = &S.Context.Idents.get(name);
+      clang::LookupResult R(S, DName, clang::SourceLocation(),
+                            clang::Sema::LookupOrdinaryName,
+                            RedeclarationKind::NotForRedeclaration);
+      R.suppressDiagnostics();
+      S.LookupQualifiedName(R, Def);
+      if (R.empty())
+        return INTEROP_RETURN(nullptr);
+      R.resolveKind();
+      if (!R.isSingleResult())
+        return INTEROP_RETURN(nullptr);
+      return INTEROP_RETURN(R.getFoundDecl()->getCanonicalDecl());
+    }
+  }
+
   // Slow path: only when qualified lookup missed AND `Within` is a
   // namespace whose enclosing chain carries at least one using-directive
   // (the only reason qualified-vs-unqualified disagree at namespace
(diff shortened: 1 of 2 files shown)


Vipul-Cariappa (approved):
LGTM! But too much unwanted comments. The code is self describing, and I assume the commit message is verbose enough.

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1257,6 +1257,33 @@ DeclRef GetNamed(const std::string& name, ConstDeclRef parent /*= nullptr*/) {
...
+          unwrap<clang::Decl>(GetOrForceDefinition(DeclRef(RD))));
+      if (Def) {
+        auto& S = getSema();
+        clang::DeclarationName DName = &S.Context.Idents.get(name);
+        clang::LookupResult R(S, DName, clang::SourceLocation(),
+                              clang::Sema::LookupOrdinaryName,
+                              RedeclarationKind::NotForRedeclaration);
+        R.suppressDiagnostics();
+        S.LookupQualifiedName(R, Def);
+        if (!R.empty()) {
```
Vipul-Cariappa: Can this nesting of `if`s be avoided using early returns?

aaronj0: yes, makes sense. I've updated it with early returns

aaronj0:
> LGTM! But too much unwanted comments. The code is self describing, and I assume the commit message is verbose enough.

Thanks, I've updated with less (verbose) comments

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1257,6 +1257,30 @@ DeclRef GetNamed(const std::string& name, ConstDeclRef parent /*= nullptr*/) {
...
+          unwrap<clang::Decl>(GetOrForceDefinition(DeclRef(RD))));
+      if (!Def)
+        return INTEROP_RETURN(nullptr);
+      auto& S = getSema();
+      clang::DeclarationName DName = &S.Context.Idents.get(name);
+      clang::LookupResult R(S, DName, clang::SourceLocation(),
+                            clang::Sema::LookupOrdinaryName,
+                            RedeclarationKind::NotForRedeclaration);
+      R.suppressDiagnostics();
+      S.LookupQualifiedName(R, Def);
```
vgvassilev, line 1278: I am not sure I get these changes. The lookups are split by kinds -- we can't find everything at the same time as this is more dangerous than one might think.

aaronj0: With `ForVisibleRedeclaration` this means inherited members are structurally unreachable. Based on the code here: https://clang.llvm.org/doxygen/SemaLookup_8cpp_source.html#l02434 we only exit early for the following reason: "Don't descend into implied contexts for redeclarations" stating [namespace.qual]p6

Is there a way to keep `ForVisibleRedeclaration` for other `Lookup::Named` callers that need to resolve names like [class.qual]? If we pass a CXXRecordDecl I assume we should trigger the `lookupInBases` done there. What do you think is the right solution here? With Cling, `LookupHelper` resolves `Base::member` through Sema's member lookup with full [class.qual] semantics

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1257,6 +1257,30 @@ DeclRef GetNamed(const std::string& name, ConstDeclRef parent /*= nullptr*/) {
   if (ND && ND != (clang::NamedDecl*)-1)
     return INTEROP_RETURN(ND->getCanonicalDecl());
 
+  // Redeclaration-style lookup stops at the class itself; [class.qual]
+  // member lookup also searches the bases. Retry accordingly.
+  if (!ND && Within) {
+    if (auto* RD = llvm::dyn_cast<clang::CXXRecordDecl>(Within)) {
+      auto* Def = llvm::dyn_cast_or_null<clang::CXXRecordDecl>(
```
aaronj0, line 1268: This is a bug

vgvassilev (approved):
Lgtm!

PR #1096, opened 2026-08-27
Return enum constant values as their int64_t bit pattern

aaronj0:
Supersedes #970, from the ROOT migration commit https://github.com/root-project/root/pull/21261/changes/4ccb37c834c69c8e2d0a6a75f162c2e049ebb389 with fixes for the Windows and WASM failures where 32-bit unsigned long overflows stoul. The fallback in that PR parsed 1ULL<<63 with std::stoul which throew out_of_range (unsigned long is 32-bit on those platforms)

GetEnumConstantValue returns `size_t` and calls `APSInt::getExtValue` unguarded, which asserts above `INT64_MAX`. Round-tripping large values through std::stoul throws where unsigned long is 32 bit (LLP64 Windows, wasm32), and a size_t return truncates there regardless. ROOT's Windows CI hit this as an uncaught out_of_range in MetaClingTests `test_GH_20925`.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -878,12 +878,16 @@ TypeRef GetEnumConstantType(ConstDeclRef DRef) {
   return INTEROP_RETURN(nullptr);
 }
 
-size_t GetEnumConstantValue(ConstDeclRef DRef) {
+int64_t GetEnumConstantValue(ConstDeclRef DRef) {
   INTEROP_TRACE(DRef);
   const auto* D = unwrap<clang::Decl>(DRef);
   if (const auto* ECD = llvm::dyn_cast_or_null<clang::EnumConstantDecl>(D)) {
     const llvm::APSInt& Val = ECD->getInitVal();
-    return INTEROP_RETURN(Val.getExtValue());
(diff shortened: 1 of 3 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -878,12 +878,16 @@ TypeRef GetEnumConstantType(ConstDeclRef DRef) {
   return INTEROP_RETURN(nullptr);
 }
 
-size_t GetEnumConstantValue(ConstDeclRef DRef) {
+int64_t GetEnumConstantValue(ConstDeclRef DRef) {
```
vgvassilev, line 881: Is that a platform-independent typedef?

vgvassilev (approved):
LGTM % comment.

lib/CppInterOp/CppInterOp.cpp
aaronj0: Yes, int64_t is available on all platforms from `<cstdint>`. We already have API's using typedefs from cstdint: GetBaseClassOffset returns `int64_t`, and InsertOrReplaceJitSymbol accepts `uint64_t` param, which matches `APSInt::getExtValue()`

PR #1097, opened 2026-08-27
Support lean embedder installs with components and unversioned output

aaronj0:
Label the header install rules `COMPONENT cppinterop-headers` and add an `install-cppinterop-headers` target, so embedders can drive a minimal install (library plus headers) through the same LLVM component machinery that already provides `install-clangCppInterOp(-stripped)`. Add `CPPINTEROP_SHARED_LIBRARY_VERSIONING` (default ON) to optionally ship the shared library as a single unversioned file, matching the Darwin layout: archives that cannot carry symlinks, such as in the case of Python wheels, otherwise materialize the SOVERSION namelink as a second full copy of the library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ option(CPPINTEROP_BUILD_TABLEGEN_ONLY "Only build cppinterop-tblgen (for cross-c
 # add_llvm_library(... BUILDTREE_ONLY)
 option(CPPINTEROP_INSTALL "Generate install rules for CppInterOp" ON)
 
-#=Set OFF to ship the shared library as a single unversioned file, matching
+# OFF to ship the shared library as a single unversioned file, matching
 # the Darwin layout: archives that cannot carry symlinks (Python wheels)
 # would otherwise materialize the SOVERSION namelink as a second full copy.
 option(CPPINTEROP_SHARED_LIBRARY_VERSIONING
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,12 @@ option(CPPINTEROP_BUILD_TABLEGEN_ONLY "Only build cppinterop-tblgen (for cross-c
 #=Set OFF to suppress install() rules and route clangCppInterOp through
 # add_llvm_library(... BUILDTREE_ONLY)
 option(CPPINTEROP_INSTALL "Generate install rules for CppInterOp" ON)
+
+#=Set OFF to ship the shared library as a single unversioned file, matching
+# the Darwin layout: archives that cannot carry symlinks (Python wheels)
+# would otherwise materialize the SOVERSION namelink as a second full copy.
+option(CPPINTEROP_SHARED_LIBRARY_VERSIONING
+  "Attach the LLVM SOVERSION and namelink to shared clangCppInterOp" ON)
 if(EMSCRIPTEN)
   set(CPPINTEROP_EXTRA_WASM_FLAGS "-fwasm-exceptions" CACHE STRING "Extra flags for wasm")
 endif()
@@ -504,6 +510,7 @@ if(CPPINTEROP_INSTALL)
 
   install(DIRECTORY include/
     DESTINATION include
+    COMPONENT cppinterop-headers
     FILES_MATCHING
     PATTERN "*.def"
     PATTERN "*.h"
@@ -513,12 +520,14 @@ if(CPPINTEROP_INSTALL)
 
   install(DIRECTORY tools/
     DESTINATION include/CppInterOp/tools
+    COMPONENT cppinterop-headers
     FILES_MATCHING
     PATTERN "*.h"
   )
 
   install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/
     DESTINATION include
+    COMPONENT cppinterop-headers
     FILES_MATCHING
     PATTERN "CMakeFiles" EXCLUDE
     PATTERN "*.inc"
@@ -617,6 +626,17 @@ if(NOT EMSCRIPTEN AND NOT CPPINTEROP_TABLEGEN_EXE)
   add_subdirectory(utils/TableGen)
 endif()
 add_subdirectory(lib)
+
+if(CPPINTEROP_INSTALL)
+  # Installing the headers needs to depend on generating the tablegen'd
+  # public headers.
+  add_custom_target(cppinterop-headers DEPENDS CppInterOpTableGen)
+  if(NOT LLVM_ENABLE_IDE)
+    add_llvm_install_targets(install-cppinterop-headers
+                             DEPENDS cppinterop-headers
+                             COMPONENT cppinterop-headers)
+  endif()
+endif()
 if (CPPINTEROP_ENABLE_TESTING)
   add_subdirectory(unittests)
 endif(CPPINTEROP_ENABLE_TESTING)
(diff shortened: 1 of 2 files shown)


aaronj0:
@vgvassilev An overview from AI to elaborate further the motivation behind this:

CppInterOp today offers only an all-or-nothing install — cmake --install delivers the
  library with its SOVERSION symlink chain, the headers, the CMake package config, and (when available) the
  out-of-process JIT runtime. An embedder that repackages CppInterOp inside an archive needs a different subset:
  just the runtime library and the headers, and as one library file, because zip-based formats like Python wheels
  cannot represent symlinks — the libclangCppInterOp.so → .so.21.1 namelink gets materialized as a second full
  copy of a ~220 MB library. Rather than having every embedder post-process the install (delete, dereference,
  strip by hand), the patch lets them ask CppInterOp's build system for exactly the lean subset, through the same
  component machinery LLVM and Clang already use (install-clang-headers, install-clangCppInterOp-stripped).

  Previous layout — what a cppjit wheel actually shipped (verified on the live install; the EP's full install plus
  the glob-copy's symlink materialization):

```
  cppjit/interop/
  ├── lib/
  │   ├── libclangCppInterOp.so        # 220 MB — materialized namelink, full copy
  │   ├── libclangCppInterOp.so.21.1   # 220 MB — the real file, duplicate content
  │   ├── cmake/CppInterOp/            # CppInterOpConfig.cmake etc. — dead weight at runtime
  │   └── cppinterop-rt/               # liborc_rt.a + llvm-jitlink-executor (when LLVM provides them)
  └── include/
      ├── CppInterOp/                  # CppInterOp.h, CppInterOpTypes.h, *.inc, ...
      └── clang/Interpreter/
```

  Lean layout — driving the install by component (install-clangCppInterOp-stripped install-cppinterop-headers)
  with CPPINTEROP_SHARED_LIBRARY_VERSIONING=OFF:
```
  cppjit/interop/
  ├── lib/
  │   └── libclangCppInterOp.so        # exactly one file, unversioned, stripped
  └── include/
      ├── CppInterOp/
      └── clang/Interpreter/
```

  Same headers, one library instead of two copies plus turds. Measured effect on this host: the stripped single
  library is 172 MB on disk (apt LLVM 21 build) and the resulting cppjit wheel is 57 MB zipped — comfortably under
  PyPI's 100 MiB cap, which the previous layout exceeded. A plain cmake --install without the component targets
  remains exactly the previous behavior, so no the behaviour does not change for any existing CppInterOp consumers.

lib/CppInterOp/CMakeLists.txt
```diff
@@ -225,6 +225,13 @@ add_llvm_library(clangCppInterOp
   ${link_libs}
 )
 
+# llvm_add_library() stamps LLVM's VERSION/SOVERSION on shared libraries on
+# ELF platforms; unsetting them ships one unversioned file, as on Darwin.
+if(NOT CPPINTEROP_SHARED_LIBRARY_VERSIONING)
+  set_property(TARGET clangCppInterOp PROPERTY VERSION)
+  set_property(TARGET clangCppInterOp PROPERTY SOVERSION)
```
mcbarton, line 232:
```suggestion
  set_target_properties(clangCppInterOp PROPERTIES NO_SONAME 1)
```
This is how we used to remove the versioned library for the Emscripten build using emsdk 3.1.73 (taken from here https://github.com/compiler-research/CppInterOp/pull/816 where this change was reverted due to no longer being needed since moving to emsdk 4.0.9)

aaronj0: Yes this is functionally viable and has the same effect and installs exactly one unversioned libclangCppInterOp.so. But the current approach is better for a few reasons:

1. It is not advised by CMake's own documentation: https://cmake.org/cmake/help/latest/prop_tgt/NO_SONAME.html recommends it "only for leaf private libraries or plugins", because dependents of a soname-less library can end up with the full link-time path in their dynamic section. clangCppInterOp is linked against directly by most consumers (only the wheels dlopen it), so we shouldn't drop the soname on them.

2. NO_SONAME leaves out DT_SONAME entirely, while unsetting the properties keeps DT_SONAME equal to the unversioned filename:

```
# NO_SONAME 1
$ readelf -d libfoo.so | grep -i soname
(no DT_SONAME)

# VERSION/SOVERSION unset
$ readelf -d libbar.so | grep -i soname
0x000000000000000e (SONAME)  Library soname: [libbar.so]
```

3. add_llvm_library only stamps `VERSION/SOVERSION` on non-Apple UNIX, so unsetting them is a no-op on macOS and today's dylib stays exactly the same. NO_SONAME applies to the macOS build too, where I assume it drops the install_name.

CMakeLists.txt
```diff
@@ -617,6 +626,17 @@ if(NOT EMSCRIPTEN AND NOT CPPINTEROP_TABLEGEN_EXE)
   add_subdirectory(utils/TableGen)
 endif()
 add_subdirectory(lib)
+
+if(CPPINTEROP_INSTALL)
+  # Installing the headers needs to depend on generating the tablegen'd
+  # public headers.
+  add_custom_target(cppinterop-headers DEPENDS CppInterOpTableGen)
+  if(NOT LLVM_ENABLE_IDE)
```
vgvassilev, line 634: Why is this condition needed?

CMakeLists.txt
```diff
@@ -17,6 +17,12 @@ option(CPPINTEROP_BUILD_TABLEGEN_ONLY "Only build cppinterop-tblgen (for cross-c
 #=Set OFF to suppress install() rules and route clangCppInterOp through
 # add_llvm_library(... BUILDTREE_ONLY)
 option(CPPINTEROP_INSTALL "Generate install rules for CppInterOp" ON)
+
+#=Set OFF to ship the shared library as a single unversioned file, matching
```
vgvassilev:
```suggestion
#OFF to ship the shared library as a single unversioned file, matching
```

CMakeLists.txt
aaronj0: This mirrors the guard AddLLVM itself puts around calls to `add_llvm_install_targets(install-<name>)`, and the guard in clang around `install-clang-headers`.

`LLVM_ENABLE_IDE` is set for the IDE generators (Xcode/VS), where LLVM skips creating the per-component install-* utility targets since each one shows up as a separate target in the IDE. Since we call `add_llvm_install_targets` directly here, the same guard applies. We could drop it if we don't care about IDE generators for the standalone build, but it might cause issues on a LLVM_ENABLE_IDE build. Haven't tested that out.

CMakeLists.txt
aaronj0: Done

CMakeLists.txt
vgvassilev: In this case, most of these things call llvm routines, why they are not handled automatically by the underlying llvm cmake infrastructure?

aaronj0: Looks like it is not handled automatically: `add_llvm_install_targets` creates the targets unconditionally.

Based on the commit logs, the reason seems to be that the guards are older than the function. They were previously per-call-site `CMAKE_CONFIGURATION_TYPES` checks that was extracted to `add_llvm_install_targets` to add the -stripped versions (a7ac2cb6fe74) . This mechanical refactor left the guards where they were, then the condition was renamed to LLVM_ENABLE_IDE (db49209c90cd) at the same call sites. Moving the check inside the function would be a fair upstream cleanup but until that happens, I assume out-of-tree callers should obey that contract?

vgvassilev: We should also proceed with moving this change if it makes sense upstream.

vgvassilev (approved):
Lgtm!

PR #1098, opened 2026-08-28
Support LLVM 23

Vipul-Cariappa:
This PR enables CppInterOp to be used with LLVM 23.

Review each commit separately; that might be easy. Reading the commit message for the diagnostics change would make it easy to review.

Note: CppJIT currently fails with LLVM 23; it requires https://github.com/llvm/llvm-project/pull/218149. Otherwise, CppJIT itself can be patched temporarily until the LLVM PR lands.

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -33,10 +33,10 @@ jobs:
         # YAML-quote `wasm` / `skip-in-prs` so workflow `if:` checks
         # comparing them work uniformly under act and github-hosted.
         include:
-          - { name: ubu24-arm-emscr-llvm22,   os: ubuntu-24.04-arm, arch: arm64,  clang-runtime: &clang_rt '22', wasm: 'true' }
-          - { name: osx26-arm-emscr-llvm22,   os: macos-26,         arch: arm64,  clang-runtime: *clang_rt, wasm: 'true' }
-          - { name: ubu24-x86-emscr-llvm22,   os: ubuntu-24.04,     arch: x86_64, clang-runtime: *clang_rt, wasm: 'true' }
-          - { name: win2025-x86-emscr-llvm22, os: windows-2025,     arch: x86_64, clang-runtime: *clang_rt, wasm: 'true', skip-in-prs: 'true' }
+          - { name: ubu24-arm-emscr-llvm23,   os: ubuntu-24.04-arm, arch: arm64,  clang-runtime: &clang_rt '23', wasm: 'true' }
+          - { name: osx26-arm-emscr-llvm23,   os: macos-26,         arch: arm64,  clang-runtime: *clang_rt, wasm: 'true' }
+          - { name: ubu24-x86-emscr-llvm23,   os: ubuntu-24.04,     arch: x86_64, clang-runtime: *clang_rt, wasm: 'true' }
+          - { name: win2025-x86-emscr-llvm23, os: windows-2025,     arch: x86_64, clang-runtime: *clang_rt, wasm: 'true', skip-in-prs: 'true' }
 
     steps:
     - uses: actions/checkout@v6
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,12 +48,14 @@ jobs:
         # patches. The setup step pairs it with clang-runtime.
         include:
           # Ubuntu Arm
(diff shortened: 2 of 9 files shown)


CMakeLists.txt
```diff
@@ -96,14 +96,14 @@ include(GNUInstallDirs)
   ## Define supported version of clang and llvm
 
   set(CLANG_MIN_SUPPORTED 20.0)
```
mcbarton, line 104: I think we now have a policy of only supporting the 3 most recent llvm versions. An llvm update PR is normally accompanied with a PR which drops the oldest llvm version.

.github/workflows/main.yml
```diff
@@ -50,6 +50,7 @@ jobs:
           # Ubuntu Arm
           - { name: ubu24-arm-gcc12-llvm22-vg, os: ubuntu-24.04-arm, compiler: gcc-12, clang-runtime: '22', Valgrind: On }
```
mcbarton: You should probably update the Valgrind jobs to llvm 23. I believe this may have been done in a separate PR though for the llvm 22 upgrade.

.github/workflows/main.yml
```diff
@@ -50,6 +50,7 @@ jobs:
...
           # Ubuntu X86
+          - { name: ubu24-x86-gcc12-llvm23, os: ubuntu-24.04, compiler: gcc-12, clang-runtime: '23'}
           - { name: ubu24-x86-gcc12-llvm22, os: ubuntu-24.04, compiler: gcc-12, clang-runtime: '22' }
           - { name: ubu24-x86-clang22-llvm22-asan-ubsan, os: ubuntu-24.04, compiler: clang-22, clang-runtime: '22', sanitizer: "Address;Undefined", flavor: asan }
           - { name: ubu24-x86-clang22-llvm22-msan, os: ubuntu-24.04, compiler: clang-22, clang-runtime: '22', sanitizer: "Memory", flavor: msan }
```
mcbarton: These will also want updating to llvm 23

mcbarton (commented):
This PR doesn't update the Emscripten jobs, so we don't know if there has been any regressions there. There was between llvm 21 and 22, and we ended up disabling tests and just accepted the regressions.

mcbarton:
I just realised the only job which has been updated to llvm 23 in the ci is the Ubuntu x86 job, so there is no check that nothing broke for llvm 23 on MacOS or Windows.

anutosh491:
I think llvm 23 supports last value printing through clang-repl !

Maybe we should take care of last value printing being supported through cppinterop here ? (so that xeus-cpp can make use of it ?)

What say @Vipul-Cariappa ?

mcbarton:
@Vipul-Cariappa you didn't update the deployment ci to llvm 23. When you do that update the xeus-cpp branch to `cppinterop-v2.0` like Vassil did to the Emscripten ci. That way CppInterOp will have a llvm 23 deployment when this PR goes in.

.github/workflows/emscripten.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         # YAML-quote `wasm` / `skip-in-prs` so workflow `if:` checks
         # comparing them work uniformly under act and github-hosted.
         include:
-          - { name: ubu24-arm-emscr-llvm22,   os: ubuntu-24.04-arm, arch: arm64,  clang-runtime: &clang_rt '22', wasm: 'true' }
+          - { name: ubu24-arm-emscr-llvm22,   os: ubuntu-24.04-arm, arch: arm64,  clang-runtime: &clang_rt '23', wasm: 'true' }
```
vgvassilev:
```suggestion
          - { name: ubu24-arm-emscr-llvm23,   os: ubuntu-24.04-arm, arch: arm64,  clang-runtime: &clang_rt '23', wasm: 'true' }
```

CMakeLists.txt
```diff
@@ -102,14 +102,14 @@ include(GNUInstallDirs)
...
```
vgvassilev:
```suggestion
  set(CLANG_MIN_SUPPORTED 21.0)
```

These need bumping too.

vgvassilev (commented):
If we are to support 3 last versions that would mean dropping 20. Also please make sure you clean up if there are any irrelevant LLVM_VERSION_MAJOR checks.

PR #1099, opened 2026-08-30
Ownership Representation

keremsahn:
Besides Alloc-DeallocType, a new added for representing functions ownership returns-takes behaviour based on ownership_... attributes, also added API for getting index of deallocated parameters for functions take ownership, and if it is specified another API is available to get which parameter decides the size of allocation for functions return ownership. Also some replacements that are not important.

@Vipul-Cariappa

### DIFF
--- a/include/CppInterOp/CppInterOpTypes.h
+++ b/include/CppInterOp/CppInterOpTypes.h
@@ -342,6 +342,11 @@ enum QualKind : unsigned char {
   All = Const | Volatile | Restrict
 };
 
+inline QualKind operator|(QualKind a, QualKind b) {
+  return static_cast<QualKind>(static_cast<unsigned char>(a) |
+                               static_cast<unsigned char>(b));
+}
+
 /// Enum modelling programming languages.
 enum class InterpreterLanguage : unsigned char {
   Unknown,
@@ -427,9 +432,18 @@ enum class DeallocType : unsigned char {
   Opaque   // Could not analyze
 };
 
-inline QualKind operator|(QualKind a, QualKind b) {
-  return static_cast<QualKind>(static_cast<unsigned char>(a) |
-                               static_cast<unsigned char>(b));
+enum class OwnershipBehaviour : unsigned char {
+  Unknown = 0, // If function does not have any ownership attribute, it is
+               // assumed to be Unknown, not None
+  OwnershipReturns = 1 << 0,
+  OwnershipHolds = 1 << 1,
+  OwnershipTakes = 1 << 2
+};
+
+inline OwnershipBehaviour operator|(OwnershipBehaviour A,
+                                    OwnershipBehaviour B) {
+  return static_cast<OwnershipBehaviour>(static_cast<unsigned char>(A) |
+                                         static_cast<unsigned char>(B));
 }
 
 enum class ValueKind : std::uint8_t {
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -1767,6 +1767,82 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
   return INTEROP_RETURN(nullptr);
 }
 
+bool IsFunctionProtoType(ConstTypeRef TyRef) {
+  INTEROP_TRACE(TyRef);
+  QualType QT = QualType::getFromOpaquePtr(TyRef.data);
+  const auto* T = QT.getTypePtr();
+  return INTEROP_RETURN(llvm::isa_and_nonnull<clang::FunctionProtoType>(T));
+}
+
+OwnershipBehaviour GetOwnershipBehaviour(ConstFuncRef Fn) {
+  INTEROP_TRACE(Fn);
+  if (!Fn)
+    return INTEROP_RETURN(OwnershipBehaviour::Unknown);
+  const auto* D = UnwrapUsingShadowToFunction(unwrap<clang::Decl>(Fn));
+  if (const auto* FTD = dyn_cast<FunctionTemplateDecl>(D))
+    D = FTD->getTemplatedDecl();
+  const auto* FD = dyn_cast<FunctionDecl>(D);
+  if (!FD)
+    return INTEROP_RETURN(OwnershipBehaviour::Unknown);
+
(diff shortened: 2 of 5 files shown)


lib/CppInterOp/CppInterOp.td
```diff
@@ -783,6 +783,54 @@ def IsFunctionProtoType : CppInterOpAPI {
...
+
+def GetDeallocationIndexes : CppInterOpAPI {
+  let Doc = [{Returns the 0-based indices of the parameters whose ownership the
+function takes or holds, as annotated by ownership_takes/ownership_holds.}];
+
+  let ReturnType = "std::vector<unsigned>";
+  let Args = [
+    Arg<"ConstFuncRef", "func">
+  ];
+}
```
Vipul-Cariappa, line 806: We should return an `int` here.
If we still want to represent multiple arguments, we can represent them as a bit mask, where the nth LSB is 1 if the function takes ownership of the nth argument.

lib/CppInterOp/CppInterOp.td
```diff
@@ -783,6 +783,54 @@ def IsFunctionProtoType : CppInterOpAPI {
...
+
+def IsOwnershipHolds : CppInterOpAPI {
+  let Doc = "Checks whether the function is annotated with ownership_holds.";
+
+  let ReturnType = "bool";
+  let Args = [
+    Arg<"ConstFuncRef", "func">
+  ];
+}
```
Vipul-Cariappa: Combine into one function:
```c++
OwnershipBehaviour GetOwnershipBehaviour(ConstFuncRef);
```

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1762,6 +1762,96 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
...
+  const auto* T = QT.getTypePtr();
+  return INTEROP_RETURN(llvm::isa_and_nonnull<clang::FunctionProtoType>(T));
+}
+
+static OwnershipBehaviour GetOwnershipBitmask(ConstFuncRef Fn) {
+  if (!Fn)
+    return OwnershipBehaviour::Unknown;
+  const auto* D = unwrap<clang::Decl>(Fn);
+  if (const auto* FD = dyn_cast<FunctionDecl>(D)) {
+    OwnershipBehaviour result = OwnershipBehaviour::Unknown; // 0b000
```
```suggestion
    OwnershipBehaviour result = OwnershipBehaviour::Unknown;
```
?

lib/CppInterOp/CppInterOp.cpp
```diff
@@ -1762,6 +1762,96 @@ TypeRef GetFunctionReturnType(ConstFuncRef func) {
   return INTEROP_RETURN(nullptr);
 }
 
+bool IsFunctionProtoType(ConstTypeRef TyRef) {
+  INTEROP_TRACE(TyRef);
+  QualType QT = QualType::getFromOpaquePtr(TyRef.data);
+  const auto* T = QT.getTypePtr();
+  return INTEROP_RETURN(llvm::isa_and_nonnull<clang::FunctionProtoType>(T));
+}
```
Vipul-Cariappa, line 1775: Why do we move this function?

include/CppInterOp/CppInterOpTypes.h
```diff
@@ -342,6 +342,11 @@ enum QualKind : unsigned char {
   All = Const | Volatile | Restrict
 };
 
+inline QualKind operator|(QualKind a, QualKind b) {
+  return static_cast<QualKind>(static_cast<unsigned char>(a) |
+                               static_cast<unsigned char>(b));
+}
+
```
Vipul-Cariappa, line 349: Why is this required? Not used anywhere.

Vipul-Cariappa (changes requested):

lib/CppInterOp/CppInterOp.td
keremsahn: That was how i started to implement, but that does not work. A function can both return ownership, take ownership of one argument and holds another one, so one attribute does not disable others. If I had implement something like GetOwnershipBehaviour I had to choose either first or last attribute in AST which makes no sense

keremsahn: But if you meant returning a bitmask that will probably work

lib/CppInterOp/CppInterOp.cpp
keremsahn: Just to be consistent, I implemented Isallocator in the wrong place at first, they should be adjacent with analyzers.

include/CppInterOp/CppInterOpTypes.h
keremsahn: It is used 
```cpp
switch (kind) {
      case OwnershipAttr::Returns:
        result = result | OwnershipBehaviour::OwnershipReturns;
        break;
      case OwnershipAttr::Takes:
        result = result | OwnershipBehaviour::OwnershipTakes;
        break;
```

lib/CppInterOp/CppInterOp.cpp
keremsahn: Just wanted to show it is 0 integral, acts as start point

lib/CppInterOp/CppInterOp.td
keremsahn: Yes makes sense

lib/CppInterOp/CppInterOp.cpp
Vipul-Cariappa: That is fine. We should revert this. Always better to have a low diff on feature implementation or bug fix PRs. We can have a refactor PR if required. But this is not a big deal.

include/CppInterOp/CppInterOpTypes.h
Vipul-Cariappa: Ok, but you are defining the `operator|` on `QualKind` here and not `OwnershipBehaviour`?

keremsahn: Yeah sorry, I thought you wrote comment for the `OwnershipBehaviour operator|`

This is a replacement, formerly `QualKind operator|` was at the bottom enum QualType, but my placement of other enums separated them so I replaced it to look good and have nice formatting.

As you suggested I can do replacements in another PR

PR #1101, opened 2026-09-01
Report the dlopen failure reason from LoadLibrary

conrade-ctc:
Two commits, reviewed and landed together per the discussion below (#1107 is folded in here and closed).

**1. Report the dlopen failure reason on stderr.** DynamicLibraryManager::loadLibrary captured the dlerror() text but logged it only under LLVM_DEBUG. cppjit's load_library reads stderr to build its Python error, so it saw an empty reason. The text now goes to std::cerr, because cppjit captures only std::cerr's rdbuf, not llvm::errs(). Paths.cpp already gets this text from ::dlerror().

**2. Add an optional reason out-parameter to LoadLibrary.** `Cpp::LoadLibrary(lib_stem, lookup, std::string* error = nullptr)`. When `error` is set it receives the dlerror() text, or `<stem>: library not found` when the search-path lookup fails, and the reason is not also printed to stderr. The two-argument form keeps the behaviour of commit 1.

- `std::string*` cannot cross the C ABI, so the .td entry is `NoCWrapper` and `cppinterop_LoadLibrary(const char*, bool)` is kept by hand in CXCppInterOp.cpp, with a test.
- `DynamicLibraryManager_LoadFailureReason` covers a dlopen failure, a failed lookup, and that success clears a stale reason.
- Coverage: the wasm branch of `Interpreter::loadLibrary` is wrapped in `LCOV_EXCL_START/STOP`, since no coverage lane runs it.

Once this lands, compiler-research/cppjit#63 bumps its pin and reads the reason from here instead of the ctypes probe.

### DIFF
--- a/lib/CppInterOp/DynamicLibraryManager.cpp
+++ b/lib/CppInterOp/DynamicLibraryManager.cpp
@@ -24,6 +24,7 @@
 #endif
 
 #include <fstream>
+#include <iostream>
 #include <sys/stat.h>
 #include <system_error>
 
@@ -369,6 +370,9 @@ DynamicLibraryManager::loadLibrary(StringRef libStem, bool permanent,
     // We emit callback to LibraryLoadingFailed when we get error with error
     // message.
     // TODO: Implement callbacks
+
+    if (!errMsg.empty())
+      std::cerr << errMsg << '\n';
 
     LLVM_DEBUG(dbgs() << "DynamicLibraryManager::loadLibrary(): " << errMsg);
 
--- a/lib/CppInterOp/DynamicLibraryManager.cpp
+++ b/lib/CppInterOp/DynamicLibraryManager.cpp
@@ -344,7 +344,7 @@ std::string DynamicLibraryManager::lookupLibrary(
 
 DynamicLibraryManager::LoadLibResult
 DynamicLibraryManager::loadLibrary(StringRef libStem, bool permanent,
-                                   bool resolved) {
+                                   bool resolved, std::string* errMsg) {
 #define DEBUG_TYPE "Dyld::loadLibrary:"
   LLVM_DEBUG(dbgs() << "Dyld::loadLibrary: " << libStem.str() << ", "
                     << (permanent ? "permanent" : "not-permanent") << ", "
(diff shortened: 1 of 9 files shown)


lib/CppInterOp/DynamicLibraryManager.cpp
```diff
@@ -370,6 +371,10 @@ DynamicLibraryManager::loadLibrary(StringRef libStem, bool permanent,
     // message.
     // TODO: Implement callbacks
 
+    // std::cerr, not llvm::errs(): cppyy captures only std::cerr's rdbuf.
```
aaronj0:
```suggestion

```

aaronj0 (approved):
LGTM! I think updating https://github.com/compiler-research/cppjit/pull/63 to make use of this directly is a more robust (and platform independent) solution

vgvassilev (approved):
LGTM!

conrade-ctc:
@aaronj0 @vgvassilev a sequencing note, since these three depend on each other: this PR first (approved by both of you), then #1107, which stacks on it and adds the `std::string*` reason out-parameter, and only then compiler-research/cppjit#63, whose pin bump needs the #1107 commit and whose CI stays red until then. If you prefer to review #1107 together with this one, that works too. Thanks!

aaronj0:
> @aaronj0 @vgvassilev a sequencing note, since these three depend on each other: this PR first (approved by both of you), then #1107, which stacks on it and adds the `std::string*` reason out-parameter, and only then [compiler-research/cppjit#63](https://github.com/compiler-research/cppjit/pull/63), whose pin bump needs the #1107 commit and whose CI stays red until then. If you prefer to review #1107 together with this one, that works too. Thanks!

Sorry for the extra churn but can you cherry-pick the commit from #1107 here since they can be reviewed and and landed as a atomic change.

conrade-ctc:
Done. The branch is rebased onto main and now carries both commits: the stderr report and the `std::string*` out-parameter from #1107 (same commit, 1764cbc, already green over there). PR title and description updated to cover both; I'm closing #1107 in favour of this one. The head moved, so the approvals need a re-tick.

@aaronj0 for viz

PR #1103, opened 2026-09-01
Share native TLS with prebuilt libraries on the in-process JIT

conrade-ctc:
The in-process ORC JIT compiles with emulated TLS. A jitted reference to a prebuilt library's `thread_local` then needs a `__emutls_v.<sym>` companion that nothing defines. `std::call_once` is the usual break: jitted code writes libstdc++'s `__once_call`, but `__once_proxy` reads the native copy. Defining the companions does not help, because the two sides then use different storage.

This PR adds an interim pass, before the module reaches the JIT. It turns each `llvm.threadlocal.address` use of an external `thread_local` declaration into a `dlsym`-based helper call. Jitted and native code then share the per-thread storage. The scope is narrow: ELF, in-process, and only symbols the dynamic linker already resolves.

Is native TLS support for the in-process JIT the intended long-term fix? If the ORC owners plan it, this pass retires.

Known gap: to run the pass between `Parse` and `Execute`, the no-value `ParseAndExecute` path skips clang's `LastValue` dump-and-clear. A value-producing input on that path no longer prints and can leave a stale `LastValue` for a later `evaluate`. An `IRTransformLayer` on the LLJIT would avoid the split.


conrade-ctc:
@vgvassilev one design point I would like your view on. To rewrite the module before it reaches the JIT, the pass splits the no-value `ParseAndExecute` into `Parse` + `Execute`. That drops clang's `LastValue` dump-and-clear on that path (see the description). An alternative is to leave `ParseAndExecute` intact and install the rewrite as an LLJIT `IRTransformLayer` transform. If that seems better to you, I am happy to rework it that way.

PR #1104, opened 2026-09-01
Helper for FunctionTemplateDecl -> FunctionDecl

keremsahn:
```cpp
D = UnwrapUsingShadowToFunction(D);
if (const auto* FTD = dyn_cast_or_null<FunctionTemplateDecl>(D))
    D = FTD->getTemplatedDecl();
return dyn_cast_or_null<FunctionDecl>(D);
```
I used this pattern at least four times and going to use again, so a helper removes repeating code

@Vipul-Cariappa


Vipul-Cariappa (commented):
This is fine, I guess. But could you also add its usages in the existing code?

PR #1105, opened 2026-09-02
Fix JitCall codegen for move-only by-value arguments

aaronj0:
Based on TClingCallFunc. make_narg_ctor never moved by-value arguments, and make_narg_call used a heuristic with triviality bits which is not comparable across standard libraries. Check the actual copy constructor at both sites, as TClingCallFunc does, completing the parameter type through GetOrForceDefinition first; a type with no reachable definition is assumed copyable and the wrapper compile reports otherwise.

Supersedes #971 (also addresses the review comments) as this PR is a superset of that one:
MSVC's std::unique_ptr has a non-trivial deleted copy constructor, so check assumes "don't move", passes lvalues, and fails compilation on Windows (RNTupleWriter::Append(std::unique_ptr<RNTupleModel>)). #971 alone would not have fixed it.

Fixes failures in the ROOT migration.

### DIFF
--- a/lib/CppInterOp/CppInterOp.cpp
+++ b/lib/CppInterOp/CppInterOp.cpp
@@ -3992,6 +3992,30 @@ void collect_type_info(const FunctionDecl* FD, QualType& QT,
   get_type_as_string(QT, type_name, C, Policy);
 }
 
+bool IsCopyConstructorDeleted(QualType QT) {
+  CXXRecordDecl* RD = QT->getAsCXXRecordDecl();
+  if (!RD) {
+    // For types that are not C++ records (such as PODs), we assume that they
+    // are copyable, ie their copy constructor is not deleted.
+    return false;
+  }
+
+  RD = unwrap<CXXRecordDecl>(GetOrForceDefinition(DeclRef(RD)));
+  if (!RD)
+    return false;
+
+  if (RD->hasSimpleCopyConstructor())
+    return false;
+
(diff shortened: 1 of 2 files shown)


lib/CppInterOp/CppInterOp.cpp
```diff
@@ -3992,6 +3992,33 @@ void collect_type_info(const FunctionDecl* FD, QualType& QT,
...
+bool IsCopyConstructorDeleted(QualType QT) {
+  CXXRecordDecl* RD = QT->getAsCXXRecordDecl();
+  if (!RD) {
+    // For types that are not C++ records (such as PODs), we assume that they
+    // are copyable, ie their copy constructor is not deleted.
+    return false;
+  }
+
+  RD = llvm::dyn_cast_or_null<CXXRecordDecl>(
+      unwrap<Decl>(GetOrForceDefinition(DeclRef(RD))));
```
```suggestion
  RD = unwrap<CXXRecordDecl>(GetOrForceDefinition(DeclRef(RD)));
```

vgvassilev (commented):
LGTM % comment. Consider dropping the assert, too.

lib/CppInterOp/CppInterOp.cpp
aaronj0: Done

aaronj0:
> LGTM % comment. Consider dropping the assert, too.

Thanks! Suggestion applied and assert has been dropped.

vgvassilev (approved):
Lgtm!

PR #1107, opened 2026-09-03
Add an optional reason out-parameter to LoadLibrary

conrade-ctc:
Follow-up to #1101 from the discussion on compiler-research/cppjit#63; this branch is stacked on #1101.

`Cpp::LoadLibrary(lib_stem, lookup, std::string* error = nullptr)`. When `error` is set it receives the dlerror() text, or `<stem>: library not found` when the search-path lookup fails, and the reason is not also printed to stderr. The two-argument form keeps the #1101 behaviour.

- `std::string*` cannot cross the C ABI, so the .td entry is `NoCWrapper` and `cppinterop_LoadLibrary(const char*, bool)` is kept by hand in CXCppInterOp.cpp, with a test.
- `DynamicLibraryManager_LoadFailureReason` now covers a dlopen failure, a failed lookup, and that success clears a stale reason.

Once this lands, cppjit#63 bumps its pin and reads the reason from here instead of the ctypes probe.

Coverage: the wasm branch of `Interpreter::loadLibrary` is wrapped in `LCOV_EXCL_START/STOP`, since no coverage lane runs it.

### DIFF
--- a/include/CppInterOp/CXCppInterOp.h
+++ b/include/CppInterOp/CXCppInterOp.h
@@ -63,6 +63,10 @@ CPPINTEROP_API intptr_t cppinterop_Evaluate(const char* code, bool* HadError);
 CPPINTEROP_API CppInterOpArray
 cppinterop_GetClassTemplatedMethods(const char* name, CppConstDeclRef parent);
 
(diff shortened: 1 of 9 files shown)


conrade-ctc:
Folded into #1101 at @aaronj0's request so both commits review and land as one change. Closing.

PR #1111, opened 2026-09-21
Skip linkage specs when walking parent scopes DeclContext

aaronj0:
A linkage specification (extern "C" / extern "C++") or a C++20 export block is a transparent declaration context, not a scope, but GetParentScope returned it as a declaration's parent. On Windows the  canonical declaration of namespace std sits inside an extern "C++" block in the MSVC CRT headers, so the reported parent of std was an unnamed decl instead of the translation unit. Skip to the enclosing scope. The skip is ordered after the UnwrapUsingShadowToFunction unwrap introduced in #995.

The test extends ScopeReflection_GetParentScope: a class inside a linkage-spec'd namespace reports that namespace as parent, and namespaces declared inside extern "C++" and extern "C" blocks report the same 
parent as a plainly declared namespace.


vgvassilev (approved):
LGTM!

PR #1112, opened 2026-09-22
Build in-process JIT with the large code model on x86_64 ELF

aaronj0:
Solves the `relocation target 0x7f22eaed6000 (DW.ref.__gxx_personality_v0) is out of range of Delta32 fixup` issues hit in conda environments and when loading large libraries.


aaronj0:
cc @lhames @vgvassilev what would be the best way to determine when the large codemodel is needed?

I'm unsure if this is also required on aarch64 platforms since I don't have a machine to reproduce it on. I can try on the cppjit CI once conda builds are in.

